PR #1, opened 2019-12-20
Add protocol specifications

Tibo-lg:
This PR adds the description of the protocol for two parties to set up a DLC.

### DIFF
--- a/Protocol.md
+++ b/Protocol.md
@@ -1,0 +1,187 @@
+# DLC Protocol
+
+This document describes a protocol for two parties to set up a Discreet Log Contract (DLC).
+
+## Preliminaries
+
+The following protocol assumes that both parties have already agreed on the contract terms (the possible outputs of the contract).
+Negotiation of the contract, or decisions on the underlying asset are thus not part of the exchanged information.
+
+## Description
+
+The two parties exchange a set of three messages, and terminates when the non-initiating party broadcasts the Fund Transaction to the Bitcoin network.
+The following figure gives a brief overview of the protocol:
+
+    +-------+                    +-------+
+    |       |                    |       |
+    |       |---- (1) offer  --->|       |
+    |       |                    |       |
+    |       |<--- (2) accept ----|       |
+    |   A   |                    |   B   |
+    |       |---- (3) sign   --->|       |
+    |       |                    |       |
+    |       |                    |  (4) broadcast fund-tx
+    |       |                    |       |
+    +-------+                    +-------+
+
+### Offer
+
+The initiating party starts the protocol by sending an `offer` message to the other party.
+This message includes the following information:
+1. Contract information (an array of triples defined below)
+1. Oracle information (optional)
+1. A's fund amount (satoshi)
+1. A's [extended public key](https://github.com/discreetlogcontracts/dlcspecs/blob/master/KeyDerivation.md)
+1. A's inputs
+1. A's change address
+1. Estimatesmartfee (satoshi/vbyte)
+1. CET CSV delay
+1. CET locktime (corresponding to the contract maturity date)
+1. Refund locktime
+
+#### Contract information
+Contract information consists of a total value (sum of both outputs) and a map to be used to create CETs.
+The map consists of:
+`message` - `a_output` (satoshi)
+where `message` is the data on which the oracle will be providing a (set of) signatures, and `a_output` represents the amount that A will receive for this outcome (B's outcome can be computed by substracting `a_output` to the total value).
+The interpretation of the message semantics is left to the application.
+
+#### Oracle information
+The oracle public key and R point(s) to use to build the CETs as well as meta information to identify the oracle to be used in the contract.
+Example:
+`public_key` - `R_point`<sub>1</sub> - `R_point`<sub>2</sub> - `meta data`
+The meta data can either be human readable for a user to be able to interpret it, or machine readable so that an application can interpret it and present it to the user.
+If both parties already have this information, transmission is unnecessary.
+
+#### A's fund amount (satoshi)
+How much A inputs into the contract.
+
+#### A's extended public key
+The extended public key to be used to [derive](https://github.com/discreetlogcontracts/dlcspecs/blob/master/KeyDerivation.md) public keys for the Fund transaction, refund transaction and CETs.
+
+#### A's inputs
+The set of UTXOs to be used as input to the fund transaction.
+
+#### A's change address
+The address to use to send the change if the sum of A's inputs is greater than the fund amount plus the fees.
+
+#### Estimatesmartfee (satoshi/vbyte)
+The fee rate to be used when computing fees for the different transactions.
+
+#### CET CSV Delay
+The number to use as input to the CSV in the [CETs](https://github.com/discreetlogcontracts/dlcspecs/blob/master/Transactions.md#outputs-1).
+
+#### CET lock time
+The number to use as `nLockTime` for the CET transactions.
+This should correspond to the maturity date of the contracts, invalidating early CET broadcasting.
+
+#### Refund locktime
+The locktime to be used for the refund transaction.
+It should be set at a later date than the maturity date of the contract.
+
+---
+
+### Accept
+After receiving the `offer` message from A, and after [validating](#offer-validation) the provided information, B creates and sends back an `accept` message containing:
+1. B's fund amount (satoshi)
+1. B's [extended public key](https://github.com/discreetlogcontracts/dlcspecs/blob/master/KeyDerivation.md)
+1. B's inputs
+1. B's change output
+1. CET signatures
+1. Refund signature
+
+#### B's fund amount (satoshi)
+How much B inputs into the contract.
+
+#### B's extended public key
+The extended public key to be used to [derive](https://github.com/discreetlogcontracts/dlcspecs/blob/master/KeyDerivation.md) public keys for the Fund transaction, refund transaction and CETs.
+
+#### B's inputs
+The set of UTXOs to be used as input to the fund transaction.
+
+#### B's change address
+The address to use to send the change if the sum of A's inputs is greater than the fund amount plus the fees.
+
+#### CET signatures
+A set of signatures from B, one for each CET input.
+
+#### Refund signature
+A signature from B for the refund transaction input.
+
+---
+
+### Sign
+
+After receiving and [validating](#sign-validation) the content of the `offer` message, A replies with a `sign` message containing:
+1. Fund transaction signatures
+1. CET signatures
+1. Refund signature
+
+#### Fund transaction signatures
+A set of signatures from A, one (in case of P2WPKH) or several (in case of multisig P2WSH) for each of A's input to the fund transaction.
+
+#### CET signatures
+A set of signatures from A, one for each CET input.
+
+#### Refund signature
+A signature from A for the refund transaction input.
+
+### Broadcast
+Upon receiving the `sign` message from A, B [validates](#sign-validation) its content, signs the fund transaction and broadcasts it to the Bitcoin network.
+
+## Validation
+
+This section describes how each party needs to validate the messages received.
+We distinguish between user validation, performed by the user (through a user interface), and machine validation, that can be done without user interaction.
+
+### Offer Validation
+
+#### User validation
+* Contract Information: confirm that the contract information matches what was agreed upon.
+* Oracle information: confirm that the oracle specified in the message matches what was agreed upon.
+* A fund amount: confirm that the input amount matches what was agreed upon.
+* Estimatesmartfee: confirm that it corresponds to what was agreed upon.
+* CET CSV delay: confirm that it corresponds to what was agreed upon.
+* Refund locktime: confirm that it corresponds to what was agreed upon.
+
+#### Machine validation
+
+* A fund amount: given this and B fund amount, need to validate that the amount incorporates the required fees for the CET or refund transaction as well as the closing transaction, and that the total input amount is greater than the payout amounts plus the fees to be paid.
+* A public key: validate that the public key is well formed.
+* Alice inputs: validate that the referenced outputs are in fact unspent and are either P2WSH or P2WPKH.
+
+### Accept validation
+
+#### User validation
+* B fund amount: confirm that the fund amount matches what was agreed upon.
+
+#### Machine validation
+* Bob public key: validate that the public key is well formed.
+* Bob inputs: validate that the referenced outputs are in fact unspent and are either P2WSH or P2WPKH.
+* CET signatures: validate that the signatures are valid for each of the CETs.
+* Refund signature: validate that the signature is valid.
+
+### Sign validation
+
+#### Machine validation
+* Fund transaction signatures: validate that the signatures are valid for each of the provided inputs.
+* CET signatures: validate that the signatures are valid for each of the CETs input.
+* Refund signature: validate that the signature is valid for the refund transaction input.
+
+## Contract cancellation
+
+At any point in the protocol up until the sending of the `sign` message, both parties can safely decide to cancel the contract.
+An empty `cancel` message can be used to inform the other party of the desire to cancel the contract.
+If one of the party doesn't send a message after a given timeout, the other party should assume the contract to be canceled.
+Timeouts should be proportional to the number of possible outputs of the contract, as the signing and verification of the CET signatures is expected to be the bottleneck.
+
+After party A sends the `sign` transaction, they cannot unilaterally cancel the contract anymore.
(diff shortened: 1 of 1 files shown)


Protocol.md
```diff
@@ -0,0 +1,177 @@
...
+This message includes the following information:
+1. Contract information (an array of triples defined below)
+1. Oracle information (optional)
+1. A's fund amount (satoshi)
+1. A's public key
+1. A's inputs
+1. A's change outputs (optional)
+1. Estimatesmartfee (satoshi/kbyte)
+1. CET CSV delay (optional)
+1. Refund locktime
```
nkohen: I think these last two should be the same as a single (absolute/CLTV) locktime

Protocol.md
```diff
@@ -0,0 +1,177 @@
...
+
+The initiating party starts the protocol by sending an `offer` message to the other party.
+This message includes the following information:
+1. Contract information (an array of triples defined below)
+1. Oracle information (optional)
+1. A's fund amount (satoshi)
+1. A's public key
+1. A's inputs
+1. A's change outputs (optional)
+1. Estimatesmartfee (satoshi/kbyte)
```
nkohen: I think it is cleanest to do `satoshi/vbyte` since this is an integer and we need `vbyte` as everything is segwit

Protocol.md
```diff
@@ -0,0 +1,177 @@
...
+### Offer
+
+The initiating party starts the protocol by sending an `offer` message to the other party.
+This message includes the following information:
+1. Contract information (an array of triples defined below)
+1. Oracle information (optional)
+1. A's fund amount (satoshi)
+1. A's public key
+1. A's inputs
+1. A's change outputs (optional)
```
nkohen: Should only be one output (at least for now that is how it is setup in Transaction.md) and I don't think there should be an option not to specify this as it is necessary for constructing the funding tx

Protocol.md
```diff
@@ -0,0 +1,177 @@
...
+    +-------+                    +-------+
+
+### Offer
+
+The initiating party starts the protocol by sending an `offer` message to the other party.
+This message includes the following information:
+1. Contract information (an array of triples defined below)
+1. Oracle information (optional)
+1. A's fund amount (satoshi)
+1. A's public key
```
nkohen: nit: extended public key (link to key generation)

Protocol.md
```diff
@@ -0,0 +1,177 @@
...
+    |       |                    |       |
+    +-------+                    +-------+
+
+### Offer
+
+The initiating party starts the protocol by sending an `offer` message to the other party.
+This message includes the following information:
+1. Contract information (an array of triples defined below)
+1. Oracle information (optional)
+1. A's fund amount (satoshi)
```
nkohen, line 33: Isn't this agreed to as part of the contract?

Protocol.md
```diff
@@ -0,0 +1,177 @@
...
+1. A's inputs
+1. A's change outputs (optional)
+1. Estimatesmartfee (satoshi/kbyte)
+1. CET CSV delay (optional)
+1. Refund locktime
+
+#### Contract information
+Contract information consists of a list of triples to be used to create CETs.
+The triple consists of:
+`message` - `a_output` (satoshi) - `b_output` (satoshi)
```
nkohen: over the wire I think that it would be best just to show one person's view as `b_output = total - a_output` always

nkohen: this way it also lets us use normal nice maps

Protocol.md
```diff
@@ -0,0 +1,177 @@
...
+where `message` is the data on which the oracle will be providing a (set of) signatures, and `a_output` and `b_output` represent the amount that A and B will receive for this outcome respectively.
+The interpretation of the message semantics is left to the application.
```
nkohen, line 47: I think this is fine for now but we should consider standardizing what is signed to being an index or a hash I think (in the future)

Protocol.md
```diff
@@ -0,0 +1,177 @@
...
+
+#### Oracle information
+The oracle R point to use to build the CETs as well as meta information to identify the oracle to be used in the contract:
```
nkohen: I think that we should make mention of the oracle pub key and also what happens here when a single event is composed of multiple R values?

Protocol.md
```diff
@@ -0,0 +1,177 @@
...
+`R_point` - `meta data`
+The meta data can either be human readable for a user to be able to interpret it, or machine readable so that an application can interpret it and present it to the user.
+If both parties already have this information, transmission is unnecessary.
+
+#### A's fund amount (satoshi)
+How much A inputs into the contract.
+
+#### A's public key
+The public key to be used in the Fund transaction, refund transaction and CETs.
```
nkohen: "extended public key" and link to key generation

Protocol.md
```diff
@@ -0,0 +1,177 @@
...
+
+#### A's change output (optional)
+If the sum of A's inputs is greater than the fund amount plus the fee for the fund transaction, A can specify a change output to be added to the fund transaction.
+
+#### Estimatesmartfee (satoshi/kbyte)
+The fee rate to be used when computing fees for the different transactions.
+
+#### CET CSV Delay (optional)
+The number to use as input to the CSV in the CET transactions.
+This is optional, and a default of 144 (roughly one day) can be use which should give enough time for a party to broadcast its closing transaction and get it included in the blockchain in most circumstances.
```
nkohen: I still think that it makes more sense just to use the same locktime as refund here since these funds should be locked up until the oracle broadcasts a signature anyway so it isn't really hurting the counter-party if you broadcast early and can't follow through

Protocol.md
```diff
@@ -0,0 +1,177 @@
...
+
+#### Refund locktime
+The locktime to be used for the refund transaction.
+It should be set at a later date than the maturity date of the contract.
+
```
nkohen, line 81: might be nice to put a vertical line here

Protocol.md
```diff
@@ -0,0 +1,177 @@
...
+### Accept
+After receiving the `offer` message from A, and after [validating](#offer-validation) the provided information, B creates and sends back an `accept` message containing:
+1. B's fund amount (satoshi)
+1. B's public key
+1. B's inputs
+1. B's change output (optional)
```
nkohen: same as above

Protocol.md
```diff
@@ -0,0 +1,177 @@
...
+* Refund signature: validate that the signature is valid.
+
+### Sign validation
+
+#### Machine validation
+* Fund transaction signatures: validate that the signatures are valid for each of the provided inputs.
+* CET signatures: validate that the signatures are valid for each of the CETs input.
+* Refund signature: validate that the signature is valid for the refund transaction input.
+
+## Contract cancellation
```
nkohen, line 171: We should probably specify a `Cancel` message here, especially in the case of wanting to cancel after sign. This way things are clear, especially since the person cancelling should always be able to outbid the fee of the funding tx which cannot be raised.

Protocol.md
Tibo-lg: Thanks the `s` was a typo. The reason it is optional is that A (or B) might not need a change output. Do you think we should specify 0?

Protocol.md
Tibo-lg: I'm not sure I see the point of a cancel message after signing (or sending the signature). As written, if a party wants to cancel after that they should try to double spend the funding transaction, but it doesn't really involve the other party AFAICT. And for cancellation prior to that, I think as written an empty message is enough (of course we should later on define some message format and IDs to recognize it, but I feel that is a next step).

Protocol.md
Tibo-lg: Yes. I guess we could skip it, but we could then also consider that a lot of other information could be part of the offline agreement. IMHO it is better to communicate all the information that is needed to reconstruct the transactions and have the receiving party validate it, rather than having them also input the amount.

Protocol.md
Tibo-lg: We discussed about this but it feels to us that the purpose of the two is quite different. If we use a single timestamp, it should be quite after the contract maturity time (to confirm that no oracle signature is available) which means that the time at which the "defending" party can recover all the funds in the case of a bad CET broadcast would be quite late, which seems somehow undesirable.

Protocol.md
Tibo-lg: Yes I agree it needs to be further defined. But maybe it has more to do with some oracle specifications?

Protocol.md
Tibo-lg: Updated

Protocol.md
Tibo-lg: Updated

Protocol.md
Tibo-lg: Updated

Protocol.md
Tibo-lg: I updated to include public key and multiple R values, let me know

Protocol.md
Tibo-lg: Updated

Protocol.md
Tibo-lg: Added here and before the `Sign` section

Protocol.md
nkohen: Essentially I'm saying that whatever message we specify for use in normal cancellation should also be recommended for use when double-spending to cancel. Essentially it is the difference between polite and impolite disconnection in most networks applications, it's always nice to have a way of saying goodbye so the other party can know definitively what happened rather than just hanging up without saying anything. (Certainly there's nothing forcing users to follow this and send the message when doing a double-spend cancel but if out-of-the-box implementations do it then it will provide a better UX overall I think)

Protocol.md
nkohen: I guess it is true that if the funding inputs are just right then there should not be a change output, but I think that this should be handled in transaction construction (i.e. removing outputs below the dust limit) rather than here as I predict that having this be optional will cause people to lose change more likely than to publish unusable utxos. I guess I think we should just have it be uniform and non-optional but that it should get dropped if it isn't needed (which is a deterministic thing since the fee rate is agreed to here)

Protocol.md
nkohen: I guess my concern in using shorter timeouts for the CET CSV delay than in the refund case is that if I broadcast the CET corresponding to what has happened, but do so early (say that the event has completed earlier than the agreed to maturity date). This could most likely happen in bad implementations (and if someone does a DLC "by hand"), especially ones that use multiple oracles (where some have already published sigs while others haven't).

In this case I think that the person who published early should still be able to claim their outputs once the signatures do eventually get published, rather than punishing them for acting too early.

I feel that this does not actually do any harm because there is no way to exploit this scheme to hold funds hostage for any longer than `refund locktime - maturity time`.

Open to any counter-arguments though!

Protocol.md
nkohen: I totally agree that it should cross over the wire at some point, we should just make sure that it doesn't do so twice :) I'm fine with leaving this here for now

Protocol.md
nkohen: looks good, though we should remember to eventually specify some kind of standard order for the R values

Protocol.md
Tibo-lg: Based on your arguments, I wonder if we shouldn't just drop this then. Since we have the fee rate and the extended pub key with derivation scheme, we should be able to determine whether to add a change output and how to construct it if needed?

Protocol.md
Tibo-lg: > I guess my concern in using shorter timeouts for the CET CSV delay than in the refund case is that if I broadcast the CET corresponding to what has happened, but do so early

I feel that would be better solved by adding a timelock to the CETs corresponding to maturity time.

> I feel that this does not actually do any harm because there is no way to exploit this scheme to hold funds hostage for any longer than refund locktime - maturity time.

I think it depends on how long is this period. As it is an edge case, I would think that the refund lock time should be quite long to give time to confirm that the oracle signature is not available.

Ichiro0219: >In this case I think that the person who published early should still be able to claim their outputs once the signatures do eventually get published, rather than punishing them for acting too early.

Let's say Alice and Bob bet if the BTCUSD rate toutch "Moon(Cap)" or "Floor" which is called digital option at 1st January and maturity date is 31st January.
Oracle publish signature if rate touch "To the moon(Cap)" or "Floor" every day(from 1st to 31st January).

In this case, Alice and Bob make CETs using R-point that is commited to each day (from 1st to 31st January)and evevt("Moon" or "Floor").
Ex) Alice hold CET (Moon,1st) ,CET (Moon,2nd)・・CET (Moon,31st January)
        Bob hold  CET (Floor,1st) ,CET (Floor,2nd) ・・CET (Floor,31st January)

By doing like this Alice and Bob can claim output when event happened.

nkohen: As per our conversation irl, we want the Refund locktime to be a CLTV well-after maturity and we want to add an `nLockTime` to every CET with expected maturity. Once this `nLockTime` is in place then penalty can be enforced using a sufficiently long `CSV`

Protocol.md
nkohen: I think that we decided that for now we will make the change SPK specified to avoid forcing derivation to be off of this DLC's key chain because funding change should not necessarily be owned by the custodian of this DLC.

Protocol.md
nkohen: I've been convinced, there's no real reason to have a cancel message as there is nothing that could have been done in response to there not being one (impolite close)

nkohen (commented):
Once the change SPKs and CET CSV are no longer marked optional I am personally good with merging this as our WIP interaction spec so that more people can see it

Protocol.md
Tibo-lg: I just wanted to clarify something here. Should it be a change output (including amount and address) or just an address and then the change is computed automatically? I feel the second option is a bit easier but I just wanted to confirm.

nkohen: I agree that it should be the second option: just an address

This being because can compute change deterministically from this

Protocol.md
Tibo-lg: I've removed the optional from the `CSV` and added a `CET lock time`.

Protocol.md
Tibo-lg: Removed the optional and changed to be just an address.

Protocol.md
```diff
@@ -0,0 +1,187 @@
...
+This message includes the following information:
+1. Contract information (an array of triples defined below)
+1. Oracle information (optional)
+1. A's fund amount (satoshi)
+1. A's [extended public key](https://github.com/discreetlogcontracts/dlcspecs/blob/master/KeyDerivation.md)
+1. A's inputs
+1. A's change address
+1. Estimatesmartfee (satoshi/vbyte)
+1. CET CSV delay
+1. CET locktime (corresponding to the contract maturity date)
```
nkohen, line 39: Same as the comment above, I think that in the long term we should not send over the wire things that have already been agreed to (funding amount, contract maturity date) and should instead specify what all assumed knowns are (as is done in Transactions.md) so that implementations can ensure to implement their own communication to get that done (or perhaps it will get specified in another spec file). But since we aren't that far along in development yet I think it is a fine idea to leave this here for now and get rid of it later since it could be helpful for early development to over-specify

Protocol.md
```diff
@@ -0,0 +1,187 @@
...
+The set of UTXOs to be used as input to the fund transaction.
+
+#### A's change address
+The address to use to send the change if the sum of A's inputs is greater than the fund amount plus the fees.
+
+#### Estimatesmartfee (satoshi/vbyte)
+The fee rate to be used when computing fees for the different transactions.
+
+#### CET CSV Delay
+The number to use as input to the CSV in the CET transactions.
```
nkohen: nit: link to the `ToLocalOutput` of `Transacitons.md` and also the T in CET is transaction so probably just CETs or spelling it all out will do

Protocol.md
```diff
@@ -0,0 +1,187 @@
...
+
+#### CET lock time
+The number to use as `nLockTime` for the CET transactions.
+This should correspond to the maturity date of the contracts, and aims to prevent (accidental) early broadcast of a CET.
```
nkohen: nit: "This should correspond to the maturity date of the contract, invalidating early CET broadcasting."

nkohen (approved):
Just a couple nits, otherwise ACK

Protocol.md
Tibo-lg: Added a link to the output section of CET in `Transaction.md` (that's the closest thing I could link to I think)

Protocol.md
Tibo-lg: Updated

Protocol.md
Tibo-lg: Yeah I guess we need to decide what we prefer here. I would personally prefer to have all the necessary information required to build the transactions be transmitted, because if we rely on implementation specific communication I feel we kind of miss the point of having compatibility across implementations.

nkohen: Yeah, I think I'm open to changing this in the future but that it should be here for now, just thought I'd leave a note :+1:

PR #2, opened 2019-12-27
Funding Fees

nkohen:
Add up-front fees to funding tx and make funding fees proportional to funding inputs for each party


Tibo-lg (approved):
LGTM

PR #6, opened 2020-01-28
Add MutualClosingTransaction

Ichiro0219:


Transactions.md
```diff
@@ -169,3 +169,45 @@ Where
...
+  * Local Mutual Closing Public Key: `ECPublicKey`
+  * Local Payout: `CurrencyUnit`
+  * Remote Mutual Closing Public Key: `ECPublicKey`
+  * Remote Payout: `CurrencyUnit`
+  * nLockTime: `UInt32`
+  * DLC Funding Output: `ScriptPubKey`
+  * Fee Rate: `FeeUnit`
+  
+Where
+  - After the contract maturity time, Mutual Closing Transaction is made for fee reduction
```
nkohen: should probably also mention improved privacy, since not using Mutual Closing Tx means your txs have a very distinct template

Transactions.md
```diff
@@ -169,3 +169,45 @@ Where
 
 Where
   - `P2WPKH(Local Penalty Public Key)`'s value is `P2WSH(Remote's ToLocalOutput).value - fee`
+
+## Mutual Closing Transaction
```
nkohen: I would prefer this case go above the two non-cooperative closing cases as it is the most common case.

Transactions.md
```diff
@@ -169,3 +169,45 @@ Where
...
+### <a name="MutualClosingKnownValues">Known Values</a>
+  * Local Mutual Closing Public Key: `ECPublicKey`
+  * Local Payout: `CurrencyUnit`
+  * Remote Mutual Closing Public Key: `ECPublicKey`
```
nkohen: Doesn't have to happen on this PR but I think that KeyDerivation.md needs an update

nkohen: I went ahead and did it: https://github.com/discreetlogcontracts/dlcspecs/pull/7

Transactions.md
Ichiro0219: Added about development in privacy.

Transactions.md
Ichiro0219: I moved above Closing Transaction (Unilateral).

Transactions.md
Ichiro0219: OK, Thanks.

nkohen (approved):
ACK

Issue #8, opened 2020-01-30
Key Derivation Concerns


benthecarman:
I have some concerns with the current way the DLC spec outlines how to do key derivations. From my understanding, we are sending extended public keys because if we have 100 possible outcomes, we don't want to pass 101 pubkeys (100 outcomes + timeout).

However, using extended public keys will likely break compatibility with most, if not all, wallets that are not aware of the DLC spec. This is because we are using the [account path](https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki#account), as the `event_id`. This will likely break [traditional account discovery](https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki#account-discovery). 

For example, if I have a DLC with `event_id = 21` a wallet using traditional account discovery will not find transactions related to these DLCs unless my wallet already has used accounts 0 through 20.  A possible solution could be to have both parties give the `event_id` they will be using so the wallet can find the relevant keys, however there can still be some problems with this as the funding key exists at `m / purpose' / coin_type' / event_id' / 0` and most wallets will be looking at the `address_index` level, not the `change` level.

Another possible issue is that the current schema makes it impossible to have it so your winnings are being sent to another party, as you must be able to sign with the parent keys.

Curious on your guys' thoughts and open to discussion about this.

EDIT 2/11/20:

Discussed this further with @nkohen @Christewart

Another issue that arises is a wallet's gap limit. If I execute a DLC that has 100 outcomes and my 99th key is used for that outcome, most wallets will not find the transaction because the 98 keys before will not have any txs so the gap limit would need to be at least 99 to see them.

In my mind there are only 2 real solutions to these problems.

1) Send over all keys: This solves all the above issues besides the address gap limit problem, wallets could then just pick the next _N_ keys and make sure they are tracking all of those. However, if their gap limit is less than _N_ they would need to know that a dlc was executed that used all _N_ keys.

2) Change spec to only use 8 keys: This solution should solve all the problems.
[...]

PR #10, opened 2020-02-14
Rework and simplify key derivation

nkohen:
Moved from using custom KeyDerivation to just using 2 keys and 2 addresses that are reused across CETs. Specifically Funding Key, ToLocal Key, ToRemote Address, and Final Address.

This PR is in response to #8


Ichiro0219:
One things we need to think about is avoiding Rogue key attack.
If Alice offer her Tolocal public key as "P(Alice)-P(Oracle(massage))", She can broadcast CET and Closing Tx without Oracle's signature(massage) because 
P(Alice) - P(Oracle(massage))+P(Oracle(massage))=P(Alice).

One idea is using the same public key for CETs' Tolocal public key and FundingTx's public key.
If Tolocal public key for CETs and public key for FundingTx is the same one,Alice can't do Rogue key attack because she can't sign FundTx.

benthecarman:
> One idea is using the same public key for CETs' Tolocal public key and FundingTx's public key.  If Tolocal public key for CETs and public key for FundingTx is the same one,Alice can't do Rogue key attack because she can't sign FundTx

If we want to avoid key reuse we could have the other party sign a message with their toLocal key (or even with funding + to Local), however, this will only enforce it if the user is validating the signatures.

Tibo-lg:
I guess the issue with key reuse is that it easily leaks that a DLC is going on. At the same time, I think it's not so difficult to guess it anyway based on the scripts used?

> If we want to avoid key reuse we could have the other party sign a message with their toLocal key (or even with funding + to Local), however, this will only enforce it if the user is validating the signatures.

I guess this works as well but it makes the protocol more complicated and error prone.

So not sure what is the best solution to be honest, it seems like both have their downside.

Ichiro0219:
>I guess the issue with key reuse is that it easily leaks that a DLC is going on. At the same time, I think it's not so difficult to guess it anyway based on the scripts used?

If Pubkey is reused between Tolocal and FundTx of CET, the Pubkey recorded in the blockchain will be different.
But if the contents of the contract can be understood by a third party (that is, if m (massage) is predictable), it is recognized that the DLC is going on bacause third party can calculate pubkey (Alice/Bob) from CET Pubkey if "m" is already known.

FundTx PubKey = pubkey (Alice/Bob)
CET Pubkey = pubkey (Alice/Bob) + R + H (R | m) X
R: Point Oracle published
X: Pubkey of Oracle
m: Massage

It is difficult to describe in this commnet, but dose it make sence?

nkohen:
I like the idea of adding the Funding PubKey to the Sig Point to avoid the Rouge Key attack but as mentioned above, the Funding PubKey becomes known to the network so the oracle can discover when they are used.

So I propose that you don't just add the FundingPubKey

nkohen:
Would our problems be solved if we don't change the keys but require proof that the private keys are known (using a digital signature) in the following way:

1. In the Offer message, a schnorr signature using the ToLocal key of the entire message must be provided
2. In the Accept message, a schnorr signature of the rest of the accept message using the ToLocal key must be provided

Unless there's something tricky I don't get, this should stop someone from using a key they know minus they oracle signature as they will not be able to produce the above validation of their ToLocal key

nkohen:
Another alternative, is to replace the `ToLocal` key to be
```
R + sha(R || m)*O + FundingPubKey + Hash(FundingPubKey || ToRemoteAddress)*G
```
rather than @Ichiro0219 's proposal:
```
R + sha(R || m)*O + FundingPubKey
```
The reason the top one is better is that if my ToLocalKey ends up on chain then the ToRemoteAddress should never end up on the chain so this will hide my DLC from everyone else (other than my counter-party).

nkohen:
I am more confident that this latter protocol cannot be gamed as reversing hashes is harder than getting points to work out in a clever way

nkohen:
It also has the benefit of reducing the protocol by one key :)

nkohen:
> Another alternative, is to replace the `ToLocal` key to be
> 
> ```
> R + sha(R || m)*O + FundingPubKey + Hash(FundingPubKey || ToRemoteAddress)*G
> ```
> 
> rather than @Ichiro0219 's proposal:
> 
> ```
> R + sha(R || m)*O + FundingPubKey
> ```
> 
> The reason the top one is better is that if my ToLocalKey ends up on chain then the ToRemoteAddress should never end up on the chain so this will hide my DLC from everyone else (other than my counter-party).

Actually, it looks like this doesn't reduce by a key as in the case where my counter-party does a force close, both my ToLocal (above) and my ToRemote will end up on chain and this will leak that this was a DLC. The fix for this is to throw a third key into the hash, so we still need that extra key exchanged.

benthecarman:
I think it's better to just sign the offer and accept messages, if we do key tweaking we are back to where wallets must be DLC conscious to find their funds

Tibo-lg:
> Actually, it looks like this doesn't reduce by a key as in the case where my counter-party does a force close, both my ToLocal (above) and my ToRemote will end up on chain and this will leak that this was a DLC. The fix for this is to throw a third key into the hash, so we still need that extra key exchanged.

Does it have to be a key? Can't it be just a value that both parties know (I'm thinking something like a contract ID)?

> I think it's better to just sign the offer and accept messages, if we do key tweaking we are back to where wallets must be DLC conscious to find their funds

Maybe I'm misunderstanding something, but there is a tweak by default in DLC with the oracle signature, so I'm unsure what would be different with @nkohen proposal.

benthecarman:
> Maybe I'm misunderstanding something, but there is a tweak by default in DLC with the oracle signature, so I'm unsure what would be different with @nkohen proposal

~~Yes, but at the end of the dlc the user's funds are not sent to this key, it is only an intermediate step.  For a wallet to support dlcs it'd be easiest for them if their funds are sent to a key that is on their wallets key chain~~

Was thinking of `toRemote` not `toLocal` ignore this

nkohen:
Implemented the following in Local's ToLocal output in the end:
```
R + sha(R || m)*O + FundingPubKey + SHA256(ToLocalPubKey)*G
```
I did this because `ToLocalPubKey` is still needed in the case where your counter-party broadcasts their CET, but putting it in the hash makes it harder to rogue key attack. And then `FundingPubKey` is used to give only one party control of this output.

Protocol.md
```diff
@@ -56,15 +56,26 @@ If both parties already have this information, transmission is unnecessary.
 #### A's fund amount (satoshi)
 How much A inputs into the contract.
 
-#### A's extended public key
-The extended public key to be used to [derive](https://github.com/discreetlogcontracts/dlcspecs/blob/master/KeyDerivation.md) public keys for the Fund transaction, refund transaction and CETs.
+#### A's public keys
+The following public keys:
+
+1. Funding Public Key
```
Tibo-lg, line 62: I feel it wouldn't hurt to add a small descriptions to theses as you've done for the addresses (especially for the second one I feel it would help).

Protocol.md
```diff
@@ -184,4 +206,4 @@ In the even that the fund transaction gets included first, both party will be bo
 Takatoshi Nakagawa
 Ichiro Kuwahara
 Thibaut Le Guilly
-Nadav Kohen
```
Tibo-lg, line 187: Is that intentional? I like to have an empty line at the end personally :P

Protocol.md
```diff
@@ -56,15 +56,26 @@ If both parties already have this information, transmission is unnecessary.
...
+2. CET ToLocal Public Key
+
+#### A's addresses
+
+The following addresses:
+
+1. Funding Change Address
+   - The address to use to send the change if the sum of A's inputs is greater than the fund amount plus the fees.
+2. ToRemote Address
+   - The address to which remote funds will be sent in the unilateral case.
```
Tibo-lg: I would say `in the case of a unilateral close of the contract` to be a bit more precise.

Tibo-lg: Or maybe mention that it is to be used in the `CETs`.

Protocol.md
```diff
@@ -56,15 +56,26 @@ If both parties already have this information, transmission is unnecessary.
...
+3. Final Address
+   - The address to which funds will be sent in refund and mutual transactions.
```
Tibo-lg: What's the reason for needing a different address for the refund/mutual and unilateral?

Transactions.md
```diff
@@ -1,6 +1,8 @@
 # DLC Transactions
 
-It is recommended that all keys be generated/derived as specified [here](https://github.com/discreetlogcontracts/dlcspecs/blob/master/KeyDerivation.md#dlc-key-derivation).
+## A Note on Key Derivation
+
+There is no strict constraint on how the three keys (Funding, ToLocal and ToRemote) and one address (Final Address) used in a DLC are generated. We do note that absent external considerations, it does seem reasonable to use [BIP 44](https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki) with four sequential address indices. We think this will usually be the best option for implementing key derivation because it is compatible with [normal wallet account discovery](https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki#account-discovery).
```
Tibo-lg: I'm not sure this is really useful information. I would rather keep things simple and to the point.

Transactions.md
```diff
@@ -39,21 +41,24 @@ Where
...
 ### <a name="CETKnownValues">Known Values</a>
   * Oracle Signature Point: `ECPublicKey`
-  * Local CET Public Key: `ECPublicKey`
+  * Local Funding Public Key: `ECPublicKey`
+  * Local CET ToLocal Public Key: `ECPublicKey`
   * Local Payout: `CurrencyUnit`
-  * Remote CET Public Key: `ECPublicKey`
+  * Remote CET ToLocal Public Key: `ECPublicKey`
+  * Remote CET ToRemote Address: `BitcoinAddress`
   * Remote Paytout: `CurrencyUnit`
```
Tibo-lg: Maybe you can fix the typo here at the same time?

Transactions.md
nkohen: I think we should have the recommendation of using BIP 44 someplace so as to document the wallet account discovery issue in the spec, do you have a better place for it in mind?

Tibo-lg (approved):
LGTM

benthecarman (approved):

PR #14, opened 2020-02-21
Describe format of DLC-transaction format LN-like

ariard:
IMO _transaction format_ to get valid signatures should be really a different spec than _validation_ of protocol messages.

i.e
> The sum of each Funding Inputs' value is at least that of its Total Collateral

Should be in `protocol.md` and a requirement on `offer` receiver to check than announced outpoint cover at least announced `offer_collateral`.

Also, I didn't write down the closing transaction because their input isn't specificed enough to know if they refer to to_local/to_remote output of CET/Mutual closing or Refund. I think it would be awesome if one of you write the onchain part a la BOLT5 where you describe on seeing a remote CET transaction what the party should do.

Overall, really excited by this :)


nkohen:
We now do LN-like things :)

PR #23, opened 2020-02-27
Sequence Numbers and Locktimes

nkohen:
Specified squence numbers and locktimes to avoid fingerprinting and RBF in Transactions.md

PR #27, opened 2020-02-28
Update roadmap to reflect #10, #23

benthecarman:


README.md
```diff
@@ -7,12 +7,14 @@ Pull requests and comments welcome.
 
 ## Specification Roadmap
 
-### Works in Progress
+### Completed
 
 - Rework partipants shared public keys and addresses ([#10](https://github.com/discreetlogcontracts/dlcspecs/pull/10))
 
 - Funding, CET, and Closing Transaction `sequence` numbers and `nLocktime` ([#23](https://github.com/discreetlogcontracts/dlcspecs/pull/23))
```
nkohen: I think I would prefer these were just wiped, no need to clutter the readme with such low level things, if we want a section like this it should probably only contain high-level stuff

PR #28, opened 2020-02-28
Update name of keys used in CET

Tibo-lg:
Updated the name of the keys used in CETs as it was a bit confusing to me. Also updated the description in the protocol as the key is not only used to compute a key but also in the penalty case.


benthecarman (approved):
ACK, this makes a lot more sense

PR #34, opened 2020-03-18
typos in Protocol desc

AdamISZ:
... and I have a few questions about CET construction, will comment below.

### DIFF
--- a/Protocol.md
+++ b/Protocol.md
@@ -110,7 +110,7 @@ The following public keys:
 1. Funding Public Key
    - B's public key used in the multi-signature [funding output](Transactions.md#FundingOutputs)
 2. Sweep Public Key
-   - A's public key used in the creation of the CETs [output script](Transactions.md#CETOutputs).
+   - B's public key used in the creation of the CETs [output script](Transactions.md#CETOutputs).
 
 #### B's addresses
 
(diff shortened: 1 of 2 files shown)


AdamISZ:
First, do we not think anti-fee sniping values for nSequence/nLocktime make sense for funding?

Second, I am quite confused about the "sweep" keys. [Here](https://github.com/discreetlogcontracts/dlcspecs/blob/master/Transactions.md#known-values-1), we have a 'remote sweep public key' and a 'local sweep public key' defined, and in Protocol.md it says :

    Sweep Public Key
    - A's public key used in the creation of the CETs [output script](Transactions.md#CETOutputs)

... but in the definition of the output script for the CET (the output with sig point + local key OR remote + timeout), there is no use of any key called "sweep".

A third point I'm confused about is related: why is the construction for the CET script:

    Oracle Signature Point + Local Funding Public Key + SHA256(Local CET ToLocal Public Key)*G

? From the paper the construction was something like `S_i + A` where `A` is a local owned key and `S_i` is the i-th signature point calculated based on the oracle. Here you're tweaking a sum of "funding public key" and a hash of another local public key; I don't understand what these two keys are. Apologies if this is already explained and I missed it.

Tibo-lg:
I just merged #33 which fixed some issues related to the sweep keys, so hopefully it should be understandable now.

> A third point I'm confused about is related: why is the construction for the CET script:

There is a possibility of rogue key attack if a party creates its public key for the CET as `Pubkey - Oracle Sig Point`. To prevent that we reuse the key from the funding transaction, but to increase privacy we tweak it (otherwise if the CET ends up on chain it is easy to see that it was part of a DLC). In order to minimize the amount of data that needs to be exchanged, we reuse the "sweep pubey" for the tweak. The sweep pubkey is the key that is used in the penalty case (if a party publishes the wrong CET). You can check #10 where this was discussed. (But maybe we need to explain the rational in the documents as well...)

> First, do we not think anti-fee sniping values for nSequence/nLocktime make sense for funding?

We used to do that but we removed it, I don't recall why exactly so I'll let @nkohen or @benthecarman  answer this one.

Otherwise thanks for catching the typos!

Tibo-lg (approved):
LGTM

Tibo-lg:
Regarding the anti fee snipping, the reasons for removing it were:
* It is quite complicated to implement and we are focussing on first having a compatible implementations before adding too many details (for example it seems like in order to avoid fingerprinting you should not do it all the time)
* We are closely following the specifications for lightning dual funding, and we will probably just adopt what they come up with, including anything about anti fee snipping if it is included.

nkohen (approved):
LGTM, will wait to hear back from @AdamISZ before merging to make sure we've answered his questions

AdamISZ:
Thanks for the responses guys. Re: anti-fee sniping that's clear. Re: the key subtraction attack defence element of the key construction, that makes sense (although I'll need to think about it more, but that's on me). Re: the "sweep" keys, yes #33 helps. But when I get a chance to look at this stuff again I'll figure out if I actually understand it this time :)

Issue #15, opened 2020-02-21
Logging spec meetings


ariard:
If you're doing spec meeting often best would be to write down a short resume or at least what was talk about. Given our timezones I guess it would allow someone to track spec progress without being there everytime. Something committed under `meeting/` ?

PR #36, opened 2020-05-01
Added meeting directory and first meeting doc

nkohen:
as per the suggestion in #15 

I propose that any additions or modifications to the agenda be proposed here, and then after the meeting I will add notes from the meeting to the doc and anything I miss can be proposed here as well, and then after the meeting notes are in we can merge.


Tibo-lg (approved):

ariard:
For future meeting give in time in UTC ;)

benthecarman:
Should add section about DLC coordination/coordinators

Issue #46, opened 2020-08-05
Partition of Specification and Assignment


nkohen:
On this issue I would like to discuss how we want to partition [the spec outline](https://dynalist.io/d/TlMSefaI7Uq4hE4M9bykU2Mh) into files/modules and subsequently decide who is going to write what.

I think the following high-level module separation makes sense:

- Oracle Requirements (as in things oracles MUST do, not SHOULD, such as provide R values)
- Oracle Best Practices/Standards
- Contract Negotiation
- Transaction Specification
- Signing Protocol (with verification steps included)
- Contract Execution (state machine?)

Perhaps we want something even more granular than this? And/or modifications to this? Please leave your opinions here.

Follow-up question: How do people feel about breaking up large documents into smaller ones contained in folders, e.g. Transaction Specification is a folder containing separate documents for funding transaction, CETs, Refund transaction, rather than having all of these things in a single file.

LLFourn:
I think ECDSA adaptor signatures in their own file similar to BIP340.

ariard:
I agree with Lloyd let put adaptor signatures in its own file section but integrate verification steps as part of contract negotiation. Negotiation can be meltdown with off-chain execution as BOLT 3 is doing.

We could have an onchain ("what should you do when you see an invalid CET onchain ?") section, as this one is only concerning a participant and observed onchain transactions, not at all counterparty. It's really a different flow than exchanging off-chain peer messages to advance contract state.

A section on contract discovery/contract offers ?

> How do people feel about breaking up large documents into smaller ones contained in folders, 

I think for now it's better to let everything in one file, especially for transactions as some subsection are factorized over the whole, like fee computation.

Tibo-lg:
@nkohen I'm not sure I get what does contract negotiation entails?

> A section on contract discovery/contract offers ?

@ariard While I think this is really nice to have, I'm not sure that should be part of v0.1.

nkohen:
> @nkohen I'm not sure I get what does contract negotiation entails?

I think by contract negotiation we mean the handshake:
```
Offer ->
      <- Accept
Sign  ->
```

Tibo-lg:
Ok thanks and then what is the difference with the Signing protocol? Is that more about generating the signature themselves locally?

nkohen:
> Ok thanks and then what is the difference with the Signing protocol? Is that more about generating the signature themselves locally?

[...]

Issue #18, opened 2020-02-23
Ordering of Inputs and Outputs in transactions


benthecarman:
Currently, you can gain a fair amount of information from the ordering of the inputs and outputs in a DLC's transactions.

Right now, an observer is be able to tell that the beginning inputs of the funding tx belong the initiator, and the ending inputs belong to the recipient, they won't exactly know where it starts and stops but they will still be able to do a reasonable estimation, also they will know who's change address is whose.

To fix this, we'd want to shuffle the inputs and outputs in a deterministic way so the that 2 parties can build the funding tx independently, the best way to do that AFAIK would to have them both randomize the ordering using the same entropy.  We could make the `eventId` the entropy, but that is likely going to be calculable through public information. A possible solution could be to use, `SHA256(initiator toLocalCETKey | recipient toLocalCETKey)` as our entropy because it should be impossible for both to appear if both parties are not reusing keys. Other solutions are possible as well. 

If we want to avoid shuffling we could use [BIP 69](https://github.com/bitcoin/bips/blob/master/bip-0069.mediawiki) as it is much easier to implement. However, this could lead to creating a fingerprint for DLC transactions issues as there are likely not many txs that have `number of inputs >= 2`, `number of outputs = 3`, and BIP 69 complaint.

ariard:
In practice, BIP69 has been disregarded due to the legitimate concern of creating 2 set of transactions: BIP69-compliant vs non-BIP69 ones. I see less the number of inputs and outputs as it can be a plausible Coinjoin tx.

If you use watchtowers at some point, using pubkeys included in any witnessScript that you may hand over to them may be a privacy leak (if both parties use the same/decryption key is dependent on a single-party broadcast) ?

Why not exchange per-DLC ephemeral pubkeys between parties and ECDH a shared secret from then and use some PRF to shuffle inputs/outputs ? (by the way do you assume a authenticated and encrypted transport between DLC parties?)

nkohen:
I agree that, at least long term, BIP 69 is probably not the way to go with the DLC spec right now. I think we should probably merge both of your ideas and do an `ECDH` between the two party's `toLocalCETKey`s as this will give us better assurances than using the hash of public keys which are "safe" to share, this should also avoid the problems brought up by ariard about watch towers.

And to your question, we will likely be reusing authenticated and encrypted transport from lightning :)

Christewart:
[...]

Issue #12, opened 2020-02-14
Deterministic Fee Computation


nkohen:
In [transaction construction](https://github.com/discreetlogcontracts/dlcspecs/blob/master/Transactions.md), both parties must always agree on the fee of a given transaction.

This requires we define a deterministic function `computeFee` that takes in an transaction and the `feerate` at time of the DLC offer message, and returns the fee in satoshis.

I have opened this issue to allow there to be a place for the discussion of what this function should look like to happen.

Bitcoin-S currently uses the following static numbers to do CET and closing fee computations
```
  /** Experimental approx. vbytes for a CET */
  private val approxCETVBytes = 190

  /** Experimental approx. vbytes for a closing tx spending ToLocalOutput */
  private val approxToLocalClosingVBytes = 122
```
Note that I haven't checked that they are correct in quite a while (and changes have been made) and also we should have different numbers when outputs are removed due to the dust threshold.

Tibo-lg:
> we should have different numbers when outputs are removed due to the dust threshold.

Do you mean we might also discard outputs in CETs? I was personally thinking of only discarding change outputs in the fund transaction.

For CET and Closing, I think it makes sense to use constants, as I cannot really think of what could influence their size.

For the fund transaction input size estimation, we use this function: https://github.com/cryptogarageinc/cfd-core/blob/master/src/cfdcore_transaction.cpp#L72

At the moment we assume that all input are just P2WPKH, but if we update the protocol we could easily support P2WSH as well.

nkohen:
I'm quite confused as to why we wouldn't discard, for example, 0-amount outputs on CETs (in the case that one party wins all the funds). They are a pretty big privacy leak if we keep them. And if we are planning to remove them which I think we should this will affect the size of transactions and hence the fees.

Tibo-lg:
Good point, you are right that we should also discard them. Then the question is, do we take into account that there might be only CETs with single output in which case the fee added to the collaterals in the fund transaction will change, or do we always put the same fee regardless and just give the leftover to the winning party? I guess the former is a bit nicer, but also a bit more complex.

nkohen:
During our meeting we decided not to change fees when outputs are removed due to them being dust. This is easier to implement as well as generally being a good idea when we move to multi-output CETs (when we have multi-party DLCs)

nkohen:
Closed by #81

Issue #11, opened 2020-02-14
Deterministic Dust Limit Computation


nkohen:
In [transaction construction](https://github.com/discreetlogcontracts/dlcspecs/blob/master/Transactions.md), both parties must always agree on which outputs end up on each transaction in order to make signing possible.

This requires we define a deterministic function `belowDustLimit` that takes in an output and the `feerate` at time of the DLC offer message, and returns true or false. If it returns true, both parties should remove that output from the transaction in question.

I have opened this issue to allow there to be a place for the discussion of what this function should look like to happen.

Some useful links:
- https://bitcoin.stackexchange.com/a/41082
- https://github.com/bitcoin/bitcoin/commit/9022aa3

Tibo-lg:
I personally liked the motivation in Murch's thesis to discard output that are less than the cost of having the output (which I think it a bit different than the definition of dust).
It's in [here](http://murch.one/wp-content/uploads/2016/11/erhardt2016coinselection.pdf) section 3.5.5

Copy paste:
```
When encountering a tiny remainder after coin selection, it may be economically
beneficent to add it to the transaction fee instead of creating a change output. In
the case that the discarded amount were less than the cost of creating a change
output, the user actually would save costs by paying more fees. Therefore, the
transaction becomes more attractive for miners to select due to the higher fee.
```

nkohen:
I agree that we definitely want to get rid of outputs that aren't economically viable. The main question then is what fee rate to use (for now probably makes the most sense to just use the fee rate at contract construction for simplicity) and also we will want to make sure we don't have any outputs (even economically viable ones) which violate relay policy.

nkohen:
During our meeting, we agreed that we should specify removal of outputs that are not economically viable (cost more to spend than they are worth), as well as those beneath relay policy thresholds.

Tibo-lg:
Just an update on this. I decided to go with a slightly modified version of the [Bitcoin core implementation of GetDustThreshold](https://github.com/bitcoin/bitcoin/blob/c536dfbcb00fb15963bf5d507b7017c241718bf6/src/policy/policy.cpp#L18) to determine if an output should be ignored or not based on a given fee rate. I guess that could be a good default?

nkohen:
The plan is still to specify this dust computation for economic viability at some point but for now just to get the spec moving forward, I've just specified a `1000 satoshi` dust limit in #51

ghost:
[...]

PR #51, opened 2020-08-06
Funding Transaction Specification

nkohen:
Took my best shot at specifying the DLC Funding transaction as things stand right now.

The funding transaction weight has been tested here: https://blockstream.info/testnet/tx/e7d2446f0823beed3a93a2289e890c42c16f9a65230dd931866aee027462f5a5

TODO:
- [ ] Funding Transaction Test Vectors
- [x] References section
- [x] Authors section
- [x] Include mention of license at the bottom

### DIFF
--- a/Transaction/DLC_Funding_Transaction.md
+++ b/Transaction/DLC_Funding_Transaction.md
@@ -33,7 +33,7 @@ All funding inputs must be Segwit or nested P2SH(Segwit) in order to protect aga
 
 ## Change Outputs
 
-The funding transaction's change outputs should pay to the address specified in the relevant offer/accept message. A change output's value should equal the total funding amount of that party subtracted by their total collateral as well as their fees for both this transaction as well as their fees for the largest possible Contract Execution Transaction. If this value is below the dust limit, then no change output is added for that party.
+The funding transaction's change outputs should pay to the address specified in the relevant offer/accept message. A change output's value should equal the total funding amount of that party subtracted by their total collateral as well as their fees for both this transaction as well as their fees for the largest possible Contract Execution Transaction. If this value is below the dust limit of `1000 satoshis`, then no change output is added for that party.
 
 # Fees
 
@@ -103,7 +103,7 @@ Multiplying non-witness data by 4 results in a weight of:
 funding_transaction_weight = 4 * funding_transaction
 
 // 2 + 108 * num_inputs weight
-witness_wight = witness_header + witness * num_inputs
+witness_weight = witness_header + witness * num_inputs
 
 overall_weight = 462 + 272 * num_inputs weight
 ```
--- a/Transaction/DLC_Funding_Transaction.md
+++ b/Transaction/DLC_Funding_Transaction.md
@@ -1,0 +1,109 @@
+# DLC Funding Transaction
+
+This details the exact format of the on-chain funding transaction, which both sides need to agree on to ensure signatures are valid.
+
+# Table of Contents
+
+* [Transaction](#transaction)
+  * [Transaction Input and Output Ordering](#transaction-input-and-output-ordering)
+  * [Funding Inputs](#funding-inputs)
+  * [FundingOutput](#funding-output)
+  * [Change Outputs](#change-outputs)
+* [Fees](#fees)
+
+# Transaction
+
+## Transaction Input and Output Ordering
+
+The Offerer/Initiator's inputs come first and in the order that they were listed in the Offer message, the Accepter's inputs come after this in the order they are listed in the Accept message. The funding output comes first, followed by the Offerer's change output if applicable, and then the Accepter's change output if applicable.
+
+Note that this will likely change in the future.
+
+## Funding Inputs
+
+All funding inputs must be Segwit or nested P2SH(Segwit) in order to protect against malleability attacks.
+
+## Funding Output
+
+* The funding output script is a P2WSH to:
+
+`2 <offer_funding_pubkey> <accept_funding_pubkey> 2 OP_CHECKMULTISIG`
+
+* Where both `pub_key`s are in compressed format.
+
+## Change Outputs
+
+The funding transaction's change outputs should pay to the address specified in the relevant offer/accept message. A change output's value should equal the total funding amount of that party subtracted by their total collateral as well as their fees for both this transaction as well as their fees for the largest possible Contract Execution Transaction. If this value is below the dust limit, then no change output is added for that party.
+
+# Fees
+
+```
+p2wpkh: 22 bytes
+	- OP_0: 1 byte
+	- OP_DATA: 1 byte (public_key_HASH160 length)
+	- public_key_HASH160: 20 bytes
+
+p2wsh: 34 bytes
+	- OP_0: 1 byte
+	- OP_DATA: 1 byte (witness_script_SHA256 length)
+	- witness_script_SHA256: 32 bytes
+
+change_output: 31 bytes
+	- value: 8 bytes
+	- var_int: 1 byte (pk_script length)
+	- pk_script (p2wpkh): 22 bytes
+
+funding_output: 43 bytes
+	- value: 8 bytes
+	- var_int: 1 byte (pk_script length)
+	- pk_script (p2wsh): 34 bytes
+
+funding_input: 41 bytes
+	- previous_out_point: 36 bytes
+		- hash: 32 bytes
+		- index: 4 bytes
+	- var_int: 1 byte (script_sig length)
+	- script_sig: 0 bytes
+	- witness <----	"witness" is used instead of "script_sig" for
+ 			transaction validation; however, "witness" is stored
+ 			separately, and the cost for its size is smaller. So,
+ 		    the calculation of ordinary data is separated
+ 			from the witness data.
+	- sequence: 4 bytes
+
+witness_header: 2 bytes
+	- flag: 1 byte
+	- marker: 1 byte
+
+witness: 108 bytes
+	- number_of_witness_elements: 1 byte
+	- sig_length: 1 byte
+	- sig: 72 bytes
+	- pub_key_length: 1 byte
+	- pub_key: 33 bytes
+
+funding_transaction: 115 + 41 * num_inputs
+	- version: 4 bytes
+	- witness_header <---- part of the witness data
+	- count_tx_in: 1 byte
+	- tx_in: 41 bytes * num_inputs
+		funding_input
+	- count_tx_out: 1 byte
+	- tx_out: 43 + 31 + 31
+		funding_ouptut,
+		change_output,
+		change_output
+	- lock_time: 4 bytes
+```
+
+Multiplying non-witness data by 4 results in a weight of:
+
+```
+// 460 + 164 * num_inputs weight
+funding_transaction_weight = 4 * funding_transaction
+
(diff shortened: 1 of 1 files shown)


Transaction/DLC_Funding_Transaction.md
```diff
@@ -0,0 +1,109 @@
...
+
+* The funding output script is a P2WSH to:
+
+`2 <offer_funding_pubkey> <accept_funding_pubkey> 2 OP_CHECKMULTISIG`
+
+* Where both `pub_key`s are in compressed format.
+
+## Change Outputs
+
+The funding transaction's change outputs should pay to the address specified in the relevant offer/accept message. A change output's value should equal the total funding amount of that party subtracted by their total collateral as well as their fees for both this transaction as well as their fees for the largest possible Contract Execution Transaction. If this value is below the dust limit, then no change output is added for that party.
```
benthecarman: > If this value is below the dust limit

Should say what it is, `1000 sats` I believe

nkohen:
Related to #12

Transaction/DLC_Funding_Transaction.md
```diff
@@ -0,0 +1,109 @@
...
+```
+
+Multiplying non-witness data by 4 results in a weight of:
+
+```
+// 460 + 164 * num_inputs weight
+funding_transaction_weight = 4 * funding_transaction
+
+// 2 + 108 * num_inputs weight
+witness_wight = witness_header + witness * num_inputs
```
benthecarman: > witness_wight

Transaction/DLC_Funding_Transaction.md
```diff
@@ -0,0 +1,109 @@
...
+
+p2wsh: 34 bytes
+	- OP_0: 1 byte
+	- OP_DATA: 1 byte (witness_script_SHA256 length)
+	- witness_script_SHA256: 32 bytes
+
+change_output: 31 bytes
+	- value: 8 bytes
+	- var_int: 1 byte (pk_script length)
+	- pk_script (p2wpkh): 22 bytes
```
benthecarman: we should allow for any type of change output

Transaction/DLC_Funding_Transaction.md
```diff
@@ -0,0 +1,109 @@
...
+	- var_int: 1 byte (pk_script length)
+	- pk_script (p2wsh): 34 bytes
+
+funding_input: 41 bytes
+	- previous_out_point: 36 bytes
+		- hash: 32 bytes
+		- index: 4 bytes
+	- var_int: 1 byte (script_sig length)
+	- script_sig: 0 bytes
+	- witness <----	"witness" is used instead of "script_sig" for
```
benthecarman, line 77: We should support any time of inputs as well

nkohen:
Begins work on #46

Transaction/DLC_Funding_Transaction.md
nkohen: I did this and left a comment on #11 so that we can improve this in the future

Transaction/DLC_Funding_Transaction.md
nkohen: Opened #52 so that this can be changed later, this is out-of-scope for now because it requires the contract negotiation specification to exist first

Transaction/DLC_Funding_Transaction.md
nkohen: Made this variable, but left the 1 byte cap (see #53)

Transaction/DLC_Funding_Transaction.md
Tibo-lg: I'm ok to leave a constant for now, but 1000 seem a bit high. I think 500 would be good enough (I just worried that considering that you can get 10000 from a testnet faucet 1000 feels a bit a lot). Also just as a note, we've basically copy pastad the code from bitcoin core for determining if an output is dust: https://github.com/bitcoin/bitcoin/blob/6d8543504d8c5bde1d12a3c60407dee44d2c8e11/src/policy/policy.cpp#L14

benthecarman: In Bitcoin-S [we have a fixed threshold of 1000 for dust](https://github.com/bitcoin-s/bitcoin-s/blob/master/core/src/main/scala/org/bitcoins/core/policy/Policy.scala#L63), however, I do agree we should move to deterministic.

Tibo-lg: I see, well if we just put 1000 as a placeholder for now I don't really mind, but just I will probably not change our implementation to that ;)

Transaction/DLC_Funding_Transaction.md
```diff
@@ -0,0 +1,110 @@
...
+  * [Funding Inputs](#funding-inputs)
+  * [FundingOutput](#funding-output)
+  * [Change Outputs](#change-outputs)
+* [Fees](#fees)
+
+# Transaction
+
+## Transaction Input and Output Ordering
+
+The Offerer/Initiator's inputs come first and in the order that they were listed in the Offer message, the Accepter's inputs come after this in the order they are listed in the Accept message. The funding output comes first, followed by the Offerer's change output if applicable, and then the Accepter's change output if applicable.
```
ariard, line 26: I think sorting change output/multisig pubkey is leaking DLC opener/accepter to any onchain observer. And this heuristic will be undoubted if there is at least 3 outputs. 

Fine for taking for v0.1 but maybe open a v0.2 issue with any privacy/security debt we're aware of

Transaction/DLC_Funding_Transaction.md
```diff
@@ -0,0 +1,110 @@
...
+
+`2 <offer_funding_pubkey> <accept_funding_pubkey> 2 OP_CHECKMULTISIG`
+
+* Where both `pub_key`s are in compressed format.
+
+## Change Outputs
+
+The funding transaction's change outputs should pay to the address specified in the relevant offer/accept message. A change output's value should equal the total funding amount of that party subtracted by their total collateral as well as their fees for both this transaction as well as their fees for the largest possible Contract Execution Transaction. If this value is below the dust limit of `1000 satoshis`, then no change output is added for that party.
+
+# Fees
```
ariard, line 48: We should ensure that funding transaction feerate is above Core's default `minRelayTxFee` (1000sat per kB). Though maybe in contract negotiation at feerate announcement, rejecting frail feerate proposal.

Transaction/DLC_Funding_Transaction.md
```diff
@@ -0,0 +1,110 @@
...
+
+## Transaction Input and Output Ordering
+
+The Offerer/Initiator's inputs come first and in the order that they were listed in the Offer message, the Accepter's inputs come after this in the order they are listed in the Accept message. The funding output comes first, followed by the Offerer's change output if applicable, and then the Accepter's change output if applicable.
+
+Note that this will likely change in the future.
+
+## Funding Inputs
+
+All funding inputs must be Segwit or nested P2SH(Segwit) in order to protect against malleability attacks.
```
ariard, line 32: We may have a concern with malleable _witnesses_.

Let's say both Alice and Bob both announce P2WSH inputs. They exchange their inputs, sign the funding transactions, sign CETs timelocked to 1 week. Bob then broadcast the funding transaction with a valid witness deflating the feerate so low that the funding transaction is stuck in network mempools. After a week, Alice wins the contract but can't spend her CET as the funding has never been confirmed. After 2 weeks, when funding transaction expires from network mempools, Bob double-spend his collateral back to him, that way invalidating a loosing contract.

This assume some congestion in mempools. Miniscript might help there. Note, channel doesn't have this issue as absolute timelock aren't starting before the funding transaction confirmation.

LLFourn: >Bob then broadcast the funding transaction with a valid witness deflating the feerate so low that the funding transaction is stuck in network mempools.

how does this work? Are you saying that some scripts have malleable witnesses so Bob can chuck random crap in their to increase the size of it without Alice knowing after they've agreed to a different witness.

matthewjablack: > After 2 weeks, when funding transaction expires from network mempools, Bob double-spend his collateral back to him, that way invalidating a loosing contract.

It seems like congested mempools are a general edge case for funding DLC's. Perhaps clients should be designed to double spend in the case that the funding tx is not confirmed within x amount of time by default.

ariard: > how does this work? Are you saying that some scripts have malleable witnesses so Bob can chuck random crap in their to increase the size of it without Alice knowing after they've agreed to a different witness.

Yes, if my P2WSH witnessScript has control flow embedded (IF/NOTIF/...), I can announce you witness 1 with feerate 10 sat/vByte but broadcast a bigger valid witness 2 dropping feerate of the whole transaction to 1  sat/vByte thus downgrading its confirmation chance in case of high-congestion.

> It seems like congested mempools are a general edge case for funding DLC's. Perhaps clients should be designed to double spend in the case that the funding tx is not confirmed within x amount of time by default.

Either this solution or anchor output. I think we would like anchor outputs anyway to let DLC participants exercise unilateral fee-bumping in case of congestion, independently of malicious behavior. Given, we don't have previous state like LN for onchain DLC, we don't have the issue of blind-CPFP/pinning that package relay solves.

Also, anchor output let a single-party push the DLC state forward, whereas RBF'ing to double-spend after x amount of time, you loss timevalue-wrt-DLC-bid of your collateral during x.

Transaction/DLC_Funding_Transaction.md
nkohen: sgtm, and #18 is already open if you wanna leave thoughts there about this

Transaction/DLC_Funding_Transaction.md
nkohen: Since we are going to have to mitigate witness malleability and fee rate malleability generally, can we allow non-witness inputs or is there some way in which segwit v0 still ends up helping us here?

Transaction/DLC_Funding_Transaction.md
benthecarman: Maybe for v0.1 we can use BIP-69 and sort the mutlisig pubkeys lexicographically as well. That way we can obfuscate the ownership of inputs and outputs to outside observers, without having to negotiate how to shuffle these.

The downside being we add a fingerprint to DLCs, however, I think that is better than leaking data about which inputs and outputs belong to whom.

benthecarman:
This should delete the old `Transaction.md`

nkohen:
> This should delete the old `Transaction.md`

Issue is this only specs out the funding tx so I figured I'd leave that to the PR that adds the last tx stuff just in case there's anything useful in there.

Transaction/DLC_Funding_Transaction.md
ariard: @benthecarman If we assume that oracle signatures are public you can intersect them with onchain cooperative closing timing no ? I'm not sure using bip-69 as LN transactions using it always have one-input/N-outputs pattern IIRC. Really that would be better to negotiate on shuffling but we don't have to do it now.

benthecarman: I agree that we should eventually move to a way to negotiate the shuffling. I just think bip 69 is a decent temp solution so user's wallet clusters are harder to distinguish.

nkohen:
Replaced by #57

PR #56, opened 2020-08-11
Contract Execution Transaction Specification

nkohen:
Took my best shot at specifying the DLC Contract Execution Transaction as things stand right now.

The CET weight has been tested here: https://blockstream.info/testnet/tx/e2678c5646dc79ed0d7e0cc79ef5b0b3539d8ec9e32de1fbb3d854aaab1709f6

Related to #12 
Related to #46

TODO:
- [ ] CET Test Vectors

### DIFF
--- a/Transaction/DLC_Contract_Execution_Transaction.md
+++ b/Transaction/DLC_Contract_Execution_Transaction.md
@@ -1,0 +1,138 @@
+# DLC Contract Execution Transaction
+
+This details the exact format of the on-chain contract execution transaction (CET), which both sides need to agree on to ensure signatures and adaptor signatures are valid.
+
+# Table of Contents
+
+* [Transaction](#transaction)
+  * [CET Outputs](#cet-outputs)
+    * [Offerer Output](#offerer-output)
+    * [Accepter Output](#accepter-output)
+* [Fees](#fees)
+* [Test Vectors](#test-vectors)
+* [References](#references)
+* [Authors](#authors)
+
+# Transaction
+
+* version: 2
+* locktime: `contract_maturity`
+* txin count: 1
+  * `txin[0]` outpoint: `txid` of funding transaction and `output_index` 0
+  * `txin[0]` sequence: 0xFFFFFFFE
+  * `txin[0]` script bytes: 0
+  * `txin[0]` witness: `0 <signature_for_offer_pubkey> <signature_for_accept_pubkey>`
+
+The output script public keys and `contract_maturity` are negotiated in the offer and accept messages.
+
+There will be one CET for every possible outcome, the output values correspond to such an outcome and are negotiated in the offer message.
+
+## CET Outputs
+
+The Offerer/Initiator's output comes first, the Accepter's output comes second.
+
+Note that this will likely change in the future.
+
+If either party receives less than the dust limit of `1000 satoshis` for this outcome, then their output is not produced.
+
+### Offerer Output
+
+This output sends funds won by the offerer corresponding to this CET's outcome to the offerer's final address specified in the offer message.
+
+### Accepter Output
+
+This output sends funds won by the accepter corresponding to this CET's outcome to the accepter's final address specified in the accept message.
+
+# Fees
+
+Fees for the CET are paid for in the Funding Transaction, the following only shows the weight calculation for the largest possible CET.
+
+```
+multi_sig: 71 bytes
+	- OP_2: 1 byte
+	- OP_DATA: 1 byte (offer_pubkey length)
+	- offer_pubkey: 33 bytes
+	- OP_DATA: 1 byte (accept_pubkey length)
+	- accept_pubkey: 33 bytes
+	- OP_2: 1 byte
+	- OP_CHECKMULTISIG: 1 byte
+
+offer_output: 9 + offer_output_script length bytes
+	- value: 8 bytes
+	- var_int: 1 byte (pk_script length)
+	- pk_script: offer_output_script length
+
+accept_output: 9 + accept_output_script length bytes
+	- value: 8 bytes
+	- var_int: 1 byte (pk_script length)
+	- pk_script: accept_output_script length
+
+funding_tx_input: 41 bytes
+	- previous_out_point: 36 bytes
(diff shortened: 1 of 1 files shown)


Transaction/DLC_Contract_Execution_Transaction.md
```diff
@@ -0,0 +1,138 @@
...
+
+There will be one CET for every possible outcome, the output values correspond to such an outcome and are negotiated in the offer message.
+
+## CET Outputs
+
+The Offerer/Initiator's output comes first, the Accepter's output comes second.
+
+Note that this will likely change in the future.
+
+If either party receives less than the dust limit of `1000 satoshis` for this outcome, then their output is not produced.
```
benthecarman, line 36: Should note that both of these are likely to change

benthecarman: Also, I think this is another good case for where we could use bip-69 for input/output ordering for v0.1

nkohen:
I'm realizing that the `Refund Transaction` is exactly the same as a `CET` except it has a different locktime and output values are equal to the offerer and accepter's total collaterals from contract negotiation. As such, I am no starting to lean towards once again having just a single specification document for transactions which combines this PR and #51 into a single file and has a section on the Refund Transaction stating that it is specified as a special `CET` construction so that there isn't so much redundancy.

Transaction/DLC_Contract_Execution_Transaction.md
```diff
@@ -0,0 +1,138 @@
...
+multi_sig: 71 bytes
+	- OP_2: 1 byte
+	- OP_DATA: 1 byte (offer_pubkey length)
+	- offer_pubkey: 33 bytes
+	- OP_DATA: 1 byte (accept_pubkey length)
+	- accept_pubkey: 33 bytes
+	- OP_2: 1 byte
+	- OP_CHECKMULTISIG: 1 byte
+
+offer_output: 9 + offer_output_script length bytes
```
ariard, line 60: Aren't output scripts normalized and thus we know their length ?

Transaction/DLC_Contract_Execution_Transaction.md
```diff
@@ -0,0 +1,138 @@
...
+
+If either party receives less than the dust limit of `1000 satoshis` for this outcome, then their output is not produced.
+
+### Offerer Output
+
+This output sends funds won by the offerer corresponding to this CET's outcome to the offerer's final address specified in the offer message.
+
+### Accepter Output
+
+This output sends funds won by the accepter corresponding to this CET's outcome to the accepter's final address specified in the accept message.
```
ariard, line 44: Replace address by scriptpubkey. Address are application-level data structs.

Transaction/DLC_Contract_Execution_Transaction.md
```diff
@@ -0,0 +1,138 @@
...
+
+The Offerer/Initiator's output comes first, the Accepter's output comes second.
+
+Note that this will likely change in the future.
+
+If either party receives less than the dust limit of `1000 satoshis` for this outcome, then their output is not produced.
+
+### Offerer Output
+
+This output sends funds won by the offerer corresponding to this CET's outcome to the offerer's final address specified in the offer message.
```
ariard, line 40: Layout the witnessScript as they're not P2WPKH IIRC?

Transaction/DLC_Contract_Execution_Transaction.md
```diff
@@ -0,0 +1,138 @@
...
+
+### Accepter Output
+
+This output sends funds won by the accepter corresponding to this CET's outcome to the accepter's final address specified in the accept message.
+
+# Fees
+
+Fees for the CET are paid for in the Funding Transaction, the following only shows the weight calculation for the largest possible CET.
```
ariard, line 48: You mean the feerate is the same than the funding transaction ? Might be better like "The fee calculation for CETs is based on the negotiated feerate_per_kw and the expected weight of the transaction."

ariard (commented):
I think we should discuss if we add CET anchor outputs for v0.1. IMO, as predicting the feerate is even harder to be right for CET given your oracle signatures might be a month from now, network mempools will be far different. With the funding, the feerate negotiated is to be broadcast soon so less likely to be inaccurate.

nkohen:
Replaced by #57

Transaction/DLC_Contract_Execution_Transaction.md
nkohen: No, there is a final address field in both the offer and accept messages which will specify the SPK used, it does not even have to be segwit.

Transaction/DLC_Contract_Execution_Transaction.md
nkohen: addresses are what are currently specified in the offer and accept messages.

Transaction/DLC_Contract_Execution_Transaction.md
benthecarman: p2wpkh is `OP_0 <20-byte-hash>` vs p2wsh is `OP_0 <32-byte-hash>`

taproot will be `OP_1 <32-byte-pub-key>`

Transaction/DLC_Contract_Execution_Transaction.md
nkohen: not sure what you mean?

Transaction/DLC_Contract_Execution_Transaction.md
nkohen: No, I mean that the outputs of the CETs are the exact outcome values negotiated and this is accomplished by including their fees in the funding output

nkohen: I.e. all fee computation is done when constructing the funding transaction, none is necessary for the CETs/Refund Tx

Transaction/DLC_Contract_Execution_Transaction.md
ariard: Gotcha, they're freely picked up by offerer/accepter ?

Issue #54, opened 2020-08-07
Anchor Output Discussion


nkohen:
From LDK DLC Slack channel for posterity:

@benthecarman asked:
> Was thinking about anchor outputs for DLCs, @nkohen told me that @ariard and @Tibo-lg thought differently, but my understanding is that they should only be needed on the funding tx if one of the parties doesn't receive a change output, so they can CPFP to get it confirmed. Whereas for the closing tx they shouldn't be needed because you can always CPFP with your winnings and if you don't receive anything from the closing tx,  then you shouldn't actually care if it gets confirmed or not.
> Thoughts?

@thebluematt responded:
> Anchor isn’t quite sufficient to solve a whole host of relay vulnerabilities Antoine (and to some extent, I) have been researching lately, but it’s definitely required to have something the broadcasting party can use to make a commitment transaction confirm via CPFP, which may be important if your counterparty goes offline. 
> So, really, if you’re running a channel with a truly untrusted counterparty, you need anchor v2 + package relay. Until then, anchor is quite nice to have, in that if fees go up a bit, you may be completely unable to close your channel.
> We’ll get it upstream sooner or later, then hopefully it’ll be easy to have support for it in DLC-LDK without effort :)

@Tibo-Ig responded:
> If I remember our discussions correctly, I think we concluded that anchoring was not required with adaptor version since it's not possible to have a CET accepted that is not the one corresponding to the actual outcome, so your counter party cannot really do anything. I think the concern was with non-adaptor version, and I don't remember exactly how the attack we discussed with Antoine worked, but I think it was something like your counter party would use some bad CETs to prevent the correct one to get accepted. (That is for basic DLCs, not thinking about channels).

@ariard responded:
> Since then, I thought more, if you offer a refunding transaction with your DLC that's a source of concern. That said I think it's less vulnerable than Lightning if you go onchain far before reaching the timelock, like ~50 block before (note Lightning soon 18). Ideally we want package relay which woud let any contracting application (think LN or DLC) increase unilateally feerate whatever your counterparty is doing (either irresponsive or malicious)

ariard:
[...]

Issue #53, opened 2020-08-06
Restrictions on Funding Change Script Public Keys


nkohen:
Do we want any restrictions on what change Script Public Keys are allowed to be?

In the current state of the specification (see #51) both parties pay equal fees for the funding transaction so I'm currently leaning towards starting out with the restriction that specifying the length of the SPK requires only one byte (i.e. the var_int for the spk length takes up only one byte of the transaction).

In the future we will likely move to a model where each participant pays the fees for their own inputs and outputs rather than sharing this burden, at which point perhaps this restriction should be lifted.

ariard:
Is there any reason to have var_int superior to one if we restrain to Segwit outputs only ? DLC is a new sytem, so less concern with being compatible with pre-Segwit software ?

nkohen:
So currently I see no reason to restrict things only to Segwit (e.g. what if you're using DLC software on top of an older hardware wallet for cold stuff which wants to do P2SH or P2WPKH or something).

nkohen:
It seems like it comes at very little cost and I only have the var_int restriction right now to make things simpler. Once parties are paying for their own change output fees then I can't think of a good reason to really have any restriction (but I could be wrong).

benthecarman:
Having a mix of p2wpkh and p2sh(p2wpkh) inputs can reduce privacy. More than likely if there is a mix, one user has all p2wpkh inputs and the other use has all the p2sh(p2wpkh) inputs. However, I think that shouldn't be restricted by the protocol, if a user wants to protection from that they should be able to configure their wallet to do so.

Christewart:
My argument for restricting it is for simplifying the number of scripts library developers need to support when building on the DLC protocol. I don't think this is a concern for us per-se in bitcoin-s as we have pretty extensive Script support, but other library developers may not and may not want to spend time on supporting eccentric change output scripts.

An example of  real world problem of this the relay policy of bitcoin core nodes, I don't think bitcoin core will relay txs with non standard scripts. :man_shrugging:

benthecarman:
> An example of real world problem of this the relay policy of bitcoin core nodes, I don't think bitcoin core will relay txs with non standard scripts. 🤷‍♂️

That makes sense, we can say only standard scripts I think that is just
P2PK, Raw Multisig. P2PKH, P2SH, P2WPKH, P2WSH

Christewart:
[...]

Issue #52, opened 2020-08-06
Support for non-P2WPKH Funding Inputs


nkohen:
Specifically support for nested P2SH(Segwit) and native P2WSH because funding inputs must still be Segwit inputs.

As of #51 fee computation is made simpler by assuming that all funding inputs are P2WPKH. In the future we want to include input info fields in our contract negotiation (similar to the proposal for LN dual-funded channels) which contain fee information so that computation can be done more accurately while staying deterministic.

nkohen:
Closed by #81

PR #57, opened 2020-08-11
Adaptor DLC Transactions Specification

nkohen:
Re-wrote the Transactions specification to match the new adaptor signature based DLCs and to match the BOLT style.

Weights are tested in the two transactions associated with the following address: https://blockstream.info/testnet/address/tb1q6zgwt0ahs9jwxu27kl9whlym93vgrn3q6anymsl2lyr9g75wa4pqjwd867

Related to #12
Related to #46

TODO:

- [ ] Test Vectors

Replaces #51 and #56

### DIFF
--- a/Transactions.md
+++ b/Transactions.md
@@ -1,182 +1,292 @@
 # DLC Transactions
 
-## A Note on Key Derivation
+This details the exact format of all on-chain transactions, which both sides need to agree on to ensure signatures are valid.
 
-There is no strict constraint on how the two keys (Funding and ToLocal) and one address (Final Address) used in a DLC are generated. We do note that absent external considerations, it does seem reasonable to use [BIP 44](https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki) with three sequential address indices. We think this will usually be the best option for implementing key derivation because it is compatible with [normal wallet account discovery](https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki#account-discovery).
+# Table of Contents
 
-## Funding Transaction
-### <a name="FundingKnownValues">Known Values</a>
-  * Local Funding Inputs: `List[TransactionInput]`
-  * Local Change Address: `BitcoinAddress`
-  * Local Funding Public Key: `ECPublicKey`
-  * Remote Funding Inputs: `List[TransactionInput]`
-  * Remote Change Address: `BitcoinAddress`
-  * Remote Funding Public Key: `ECPublicKey`
-  * Total Local Collateral: `CurrencyUnit`
-  * Total Remote Collateral: `CurrencyUnit`
-  * Fee Rate: `FeeUnit`
+* [Funding Transaction](#funding-transaction)
+  * [Funding Transaction Input and Output Ordering](#funding-transaction-input-and-output-ordering)
+  * [Funding Inputs](#funding-inputs)
+  * [FundingOutput](#funding-output)
+  * [Change Outputs](#change-outputs)
+* [Contract Execution Transaction](#contract-execution-transaction)
+  * [CET Outputs](#cet-outputs)
+    * [Offerer Output](#offerer-output)
+    * [Accepter Output](#accepter-output)
+* [Refund Transaction](#refund-transaction)
+* [Fees](#fees)
+  * [Expected Weight of the Funding Transaction](#expected-weight-of-the-funding-transaction)
+  * [Expected Weight of the Contract Execution or Refund Transaction](#expected-weight-of-the-contract-execution-or-refund-transaction)
+* [Test Vectors](#test-vectors)
+* [References](#references)
+* [Authors](#authors)
 
-Where
-  - The sum of the values of each `Funding Inputs` value is at least that of its corresponding `Total Collateral`
-  - `Funding Public Key`s are both 33-byte compressed public keys
-### <a name="FundingGlobal">Global</a>
-  * nLockTime is `0`
-### <a name="FundingInputs">Inputs</a>
-  * Local Funding Inputs
-      * All should have nSequence of `0xffffffff`
-  * Remote Funding Inputs
-      * All should have nSequence of `0xffffffff`
-### <a name="FundingOutputs">Outputs</a>
-  * P2WSH(DLC Funding Output)
-  * Local Change Address
-  * Remote Change Address
+# Funding Transaction
 
-Where
-  - `P2WSH(DLC Funding Output)`'s value is `Total Local Collateral + Total Remote Collateral + Computed CET Fee + Computed ToLocal Closing Fee`
-  - `DLC Funding Output`'s script is
-   
-        OP_2 <Local Funding Public Key> <Remote Funding Public Key> OP_2 OP_CHECKMULTISIG
-   
-  - Each `Change Address`'s value is at most that of its respective `Sum(Funding Inputs) - Total Collateral - Computed Fees - (Computed CET Fee + Computed ToLocal Closing Fee)/2` with `Computed Fees` being proportional to each party's total input weight and `Computed CET Fee` being the estimated fee for a [Contract Execution Transaction](#contract-execution-transaction) and `Computed ToLocal Closing Fee` being the estimated fee for a [Unilateral Closing Transaction](#ClosingUnilateral)
+* version: 2
+* locktime: 0
 
-## Contract Execution Transaction
-### <a name="CETKnownValues">Known Values</a>
-  * Oracle Signature Point: `ECPublicKey`
-  * Local Funding Public Key: `ECPublicKey`
-  * Local Sweep Public Key: `ECPublicKey`
-  * Local Payout: `CurrencyUnit`
-  * Remote Sweep Public Key: `ECPublicKey`
-  * Remote Final Address: `BitcoinAddress`
-  * Remote Payout: `CurrencyUnit`
-  * nLockTime: `UInt32`
-  * Timeout: `UInt32`
-  * DLC Funding Output: `ScriptPubKey`
-  * Fee Rate: `FeeUnit`
+The funding inputs and change output script public keys are negotiated in the offer and accept messages.
 
-Where
-  - `Oracle Signature Point` is the 33-byte public key associated with this CET's outcome
-  - `Local Funding Public Key` is the local key from the [funding transaction](#funding-transaction)
-  - Both `Sweep Public Key`s are 33-byte compressed public keys
-  - `Local Payout + Remote Payout = (DLC Funding Output).value`
-  - `nLockTime` is set to the contract maturity time
-  - `Timeout` is a CSV locktime after which [penalty transactions](#ClosingPenalty) are valid
-  - `DLC Funding Output` is of the form [specified above](#FundingOutputs)
-### <a name="CETGlobal">Global</a>
-  * nLockTime
-### <a name="CETInputs">Inputs</a>
-  * Input Spending(P2WSH(DLC Funding Output))
-      * nSequence is `0xfffffffe`
-### <a name="CETOutputs">Outputs</a>
-  * P2WSH(ToLocalOutput)
-  * ToRemoteOutput
+## Funding Transaction Input and Output Ordering
 
-Where
-  - `P2WSH(ToLocalOutput).value = Local Payout + Computed ToLocal Closing Fee`
-  - `ToRemoteOutput.value = Remote Payout`
-  - `ToLocalOutput`'s script is:
-  
-        OP_IF
-          <Oracle Signature Point + Local Funding Public Key + SHA256(Local Sweep Public Key)*G>
-        OP_ELSE
-          <Timeout> OP_CHECKSEQUENCEVERIFY OP_DROP
-          <Remote Sweep Public Key>
-        OP_ENDIF
-        OP_CHECKSIG
-      
-      - Note that the addition in the `IF` case is elliptic curve point addition
-  - `ToRemoteOutput`'s script corresponds to `Remote CET Final Address`
+The Offerer/Initiator's inputs come first and in the order that they were listed in the Offer message, the Accepter's inputs come after this in the order they are listed in the Accept message. The funding output comes first, followed by the Offerer's change output if applicable, and then the Accepter's change output if applicable.
 
-## Refund Transaction
-### <a name="RefundKnownValues">Known Values</a>
-  * Local Final Address: `BitcoinAddress`
-  * Total Local Collateral: `CurrencyUnit`
-  * Remote Final Address: `BitcoinAddress`
-  * Total Remote Collateral: `CurrencyUnit`
-  * Timeout: `UInt32`
-  * DLC Funding Output: `ScriptPubKey`
-  * Fee Rate: `FeeUnit`
+Note that this will likely change in the future.
 
-Where
-  - Unlike CETs in a DLC, there is only one Refund Transaction that both parties share, similar to how there is only one [Funding Transaction](#funding-transaction)
-  - `Total Local Collateral + Total Remote Collateral = (DLC Funding Output).value`
-  - `Timeout` is a CLTV locktime set well after the contract maturity time
-  - `DLC Funding Output` is of the form [specified above](#FundingOutputs)
-### <a name="RefundGlobal">Global</a>
-  * nLockTime is `Timeout`
-### <a name="RefundInputs">Inputs</a>
-  * Input Spending(P2WSH(DLC Funding Output))
-      * nSequence is `0xfffffffe`
-### <a name="RefundOutputs">Outputs</a>
-  * ToLocalOutput
-  * ToRemoteOutput
+## Funding Inputs
 
-Where
-  - `ToLocalOutput`'s value is `Total Local Collateral + RefundFeeDelta/2`
-  - `ToRemoteOutput`'s value is `Total Remote Collateral + RefundFeeDelta/2`
-  - `RefundFeeDelta = Computed CET Fee + Computed ToLocal Closing Fee - Computed Refund Tx Fee` (note that the Refund Transaction is smaller than any CET)
-  - `ToLocalOutput`'s script is that of `Local Final Address`
-  - `ToRemoteOutput`'s script is that of `Remote Final Address`
+All funding inputs must be Segwit or nested P2SH(Segwit) in order to protect against malleability attacks.
 
-## Mutual Closing Transaction
-### <a name="MutualClosingKnownValues">Known Values</a>
-  * Local Final Address: `BitcoinAddress`
-  * Local Payout: `CurrencyUnit`
-  * Remote Final Address: `BitcoinAddress`
-  * Remote Payout: `CurrencyUnit`
-  * DLC Funding Output: `ScriptPubKey`
-  * Fee Rate: `FeeUnit`
+## Funding Output
 
-Where
-  - After the contract maturity time, Mutual Closing Transaction is created in cooperation for fee reduction and improvement in privacy 
-  - `Local Payout = (Contract Execution Transaction Local Payout).value`
-  - `Remote Payout = (Contract Execution Transaction Remote Payout).value`
-  - `DLC Funding Output` is of the form [specified above](#FundingOutputs)
-### <a name="MutualClosingGlobal">Global</a>
-  * nLockTime is `0`
-### <a name="MutualClosingInputs">Inputs</a>
-  * Input Spending(P2WSH(DLC Funding Output))
-      * nSequence is `0xffffffff`
-### <a name="MutualClosingOutputs">Outputs</a>
-  * ToLocalOutput
-  * ToRemoteOutput
+* The funding output script is a P2WSH to:
 
-Where
-  - `ToLocalOutput`'s value is `Local Payout + MutualClosingFeeDelta/2`
-  - `ToRemoteOutput`'s value is `Remote Payout + MutualClosingFeeDelta/2`
-  - `MutualClosingFeeDelta = Computed CET Fee + Computed ToLocal Closing Fee - Computed MutualClosing Tx Fee` (note that the Mutual Closing Transaction is smaller than any CET)
-  - `ToLocalOutput`'s script is that of `Local Final Address`
+```
+2 <offer_funding_pubkey> <accept_funding_pubkey> 2 OP_CHECKMULTISIG
+```
 
-  - `ToRemoteOutput`'s script is that of `Remote Final Address`
+* Where both `pub_key`s are in compressed format.
 
-## <a name="ClosingUnilateral">Closing Transaction (Unilateral)</a>
-### <a name="ClosingKnownValues">Known Values</a>
-  * Local Final Address: `BitcoinAddress`
-  * nLockTime: `UInt32`
-  * Local Payout: `CurrencyUnit`
-  * ToLocalOutput: `ScriptPubKey`
-  * Fee Rate: `FeeUnit`
+## Change Outputs
 
-Where
-  - `ToLocalOutput` is of the form [specified above](#CETOutputs)
-### <a name="ClosingGlobal">Global</a>
-  * nLockTime is `0`
-### <a name="ClosingInputs">Inputs</a>
-  * Input Spending(P2WSH(ToLocalOutput))
-      * nSequence is `0xffffffff`
-### <a name="ClosingOutputs">Outputs</a>
-  * One output corresponding to `Local Final Address` with value `Local Payout`
+The funding transaction's change outputs should pay to the address specified in the relevant offer/accept message. A change output's value should equal the total funding amount of that party subtracted by their total collateral as well as their fees for both this transaction as well as their fees for the largest possible Contract Execution Transaction. If this value is below the dust limit of `1000 satoshis`, then that party must include additional funding in order to ensure they have a valid anchor output.
 
-## <a name="ClosingPenalty">Closing Transaction (Penalty)</a>
-### <a name="ClosingKnownValues">Known Values</a>
-  * Local Address: `BitcoinAddress`
-  * Remote's ToLocalOutput: `ScriptPubKey`
-  * Fee Rate: `FeeUnit`
+# Contract Execution Transaction
 
-Where
-  - `Local Address` is any unused local address
-  - `Remote's ToLocalOutput` is of the form [specified above](#CETOutputs)
-### <a name="ClosingInputs">Inputs</a>
-  * Input Spending(P2WSH(Remote's ToLocalOutput))
-### <a name="ClosingOutputs">Outputs</a>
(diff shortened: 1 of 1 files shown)


nkohen:
Comments from the replaced PRs:

- https://github.com/discreetlogcontracts/dlcspecs/pull/51#discussion_r466601354
- https://github.com/discreetlogcontracts/dlcspecs/pull/51#discussion_r467585736
- https://github.com/discreetlogcontracts/dlcspecs/pull/51#discussion_r467589135
- https://github.com/discreetlogcontracts/dlcspecs/pull/56#pullrequestreview-465241478

nkohen:
> I think we should discuss if we add CET anchor outputs for v0.1. IMO, as predicting the feerate is even harder to be right for CET given your oracle signatures might be a month from now, network mempools will be far different. With the funding, the feerate negotiated is to be broadcast soon so less likely to be inaccurate.

@ariard I thought we had decided that anchor outputs for parties that win nothing for a given outcome are not needed?

Transactions.md
```diff
@@ -1,182 +1,292 @@
...
+```
+Funding Transaction weight:    286 + 4 * total_change_length + 272 * num_inputs
+CET/Refund Transaction weight: 500 + 4 * total_output_length
+```
+
+### Fee Payment
+
+All fees are paid for in the [Funding Transaction](#funding-transaction) so that the funding output's value is inflated so that the outputs of CETs and the refund transaction are the exact amounts specified in the offer message's contract information (or total collateral specified in the offer and accept messages for the refund transaction) with no fees subtracted.
+
+All fees are currently paid evenly between the two parties, though this will change in a future version.
```
ariard, line 120: One party can make the cost of aggregating its utxos spread unevenly by the other one. Alice declares 10 inputs, Bob only 2, assuming spends are equal, Bob will pays for 4 of Alice spends ?

A partial counter-measure, left to implementations, is to reject DLC offer if number of inputs are blatantly unbalanced.

Transactions.md
```diff
@@ -1,182 +1,292 @@
...
+This yields the following *expected weights* (details of the computation below):
+
+```
+Funding Transaction weight:    286 + 4 * total_change_length + 272 * num_inputs
+CET/Refund Transaction weight: 500 + 4 * total_output_length
+```
+
+### Fee Payment
+
+All fees are paid for in the [Funding Transaction](#funding-transaction) so that the funding output's value is inflated so that the outputs of CETs and the refund transaction are the exact amounts specified in the offer message's contract information (or total collateral specified in the offer and accept messages for the refund transaction) with no fees subtracted.
```
ariard: Ah I finally understood, "The funding output value is composed of the total_collateral (as sum up from offer/accept messages) plus base_fee of second-stage transactions, either CETs or refund" sounds better ?

ariard:
@nkohen Of course I agree you don't care about CET if you don't have funds at stake, but if you do you want to bump feerate to avoid CET stucking in mempools in case of congestion, and thus unilaterally.

I can write down the anchor output part later, to avoid slow down this PR

nkohen:
> @nkohen Of course I agree you don't care about CET if you don't have funds at stake, but if you do you want to bump feerate to avoid CET stucking in mempools in case of congestion, and thus unilaterally.
> 
> I can write down the anchor output part later, to avoid slow down this PR

Doesn't the output which contains the payout act as an anchor output that can be spent using CPFP? Or am I missing something

Transactions.md
nkohen: So as I've kind of alluded to here, I think the plan for v0.1 is to not have users pay fees evenly but rather to have each user pay for their own inputs and change output (and the rest evenly), but I didn't want to clutter this PR with that discussion so I'm hoping to open a PR once this gets merged with that part of the specification, I'll open an issue for this so I don't forget.

nkohen: Here it is including some explanation of what change will be needed :) https://github.com/discreetlogcontracts/dlcspecs/issues/58

Transactions.md
```diff
@@ -1,182 +1,292 @@
...
+
+## Funding Output
+
+* The funding output script is a P2WSH to:
+
+```
+2 <offer_funding_pubkey> <accept_funding_pubkey> 2 OP_CHECKMULTISIG
+```
+
+* Where both `pub_key`s are in compressed format.
```
benthecarman: nit: above you say `pubkey` but then here you say `pub_key`

Transactions.md
benthecarman: I think `base_fee` should be replaced with `max_fee` since we want to accommodate for the largest CET, not a base CET

benthecarman (approved):

Transactions.md
```diff
@@ -1,182 +1,292 @@
...
+
+## Change Outputs
+
+The funding transaction's change outputs should pay to the address specified in the relevant offer/accept message. A change output's value should equal the total funding amount of that party subtracted by their total collateral as well as their fees for both this transaction as well as their fees for the largest possible Contract Execution Transaction. If this value is below the dust limit of `1000 satoshis`, then that party must include additional funding in order to ensure they have a valid anchor output.
```
Tibo-lg: nit: "as well" is repeated it sounds strange (to me at least, you can ignore if it sounds ok to you).

Transactions.md
```diff
@@ -1,182 +1,292 @@
...
+
+# Contract Execution Transaction
+
+Also known as a CET.
```
Tibo-lg, line 57: Maybe link to the glossary?

Transactions.md
```diff
@@ -1,182 +1,292 @@
...
+
+This output sends funds won by the offerer corresponding to this CET's outcome to the offerer's final address specified in the offer message.
+
+### Accepter Output
+
+This output sends funds won by the accepter corresponding to this CET's outcome to the accepter's final address specified in the accept message.
+
+# Refund Transaction
+
+The refund transaction is exactly the same as a [Contract Execution Transaction](#contract-execution-transaction) except that its locktime is `contract_timeout` (as negotiated in the offer message) instead of `contract_maturity` and the output values for the offerer and the accepter are their total collateral values in the offer and accept message respectively.
```
Tibo-lg: nit: "respective total collateral values"?

Transactions.md
```diff
@@ -1,182 +1,292 @@
...
+
+* Bitcoin uses DER-encoded signatures, which vary in size.
+* Bitcoin also uses variable-length integers, so a large number of outputs will take 3 bytes to encode rather than 1.
+* The offerer output may be below the dust limit.
+* The accepter output may be below the dust limit.
+
+Thus, a simplified formula for *expected weight* is used, which assumes:
+
+* Signatures are 72 bytes long.
+* There are a small number of outputs (thus 1 byte to count them).
```
Tibo-lg, line 107: Isn't the number of outputs fixed anyway? (4 for fund tx and max 2 for CET)

Transactions.md
```diff
@@ -1,182 +1,292 @@
...
+* There are always both an offerer output and an accepter output.
```
Tibo-lg: IIUC that's only true for the fund transaction, since for the CET you indicate that it can be droped?

Transactions.md
benthecarman: technically for the funding tx, you could batch a bunch of them

Tibo-lg: Good point, hadn't thought of that :)

Transactions.md
nkohen: whoops, yes I was going back and forth because of my misunderstanding earlier of anchor outputs but I will fix this

Transactions.md
```diff
@@ -1,182 +1,292 @@
...
+## Change Outputs
+
+The funding transaction's change outputs should pay to the address specified in the relevant offer/accept message. A change output's value should equal the total funding amount of that party subtracted by their total collateral as well as their fees for both this transaction as well as their fees for the largest possible Contract Execution Transaction. If this value is below the dust limit of `1000 satoshis`, then that party must include additional funding in order to ensure they have a valid anchor output.
+
+# Contract Execution Transaction
+
+Also known as a CET.
+
+* version: 2
+* locktime: `contract_maturity`
```
Christewart: What is the reasoning for having this again? It seems to me, if we are relying on an oracle, the contract "matures" whenever the Oracle publishes. I think things like this are better handled at the application layer to provide an "estimate" of when that will be rather than in the protocol

As far as I can tell, this doesn't provide any security guarantees, but does add unnecessary complexity when thinking about the protocol? For instance (I'm not 100% on this) this may effect relay policies of CETs on the network depending, it can also effect fee bumping logic (cc @ariard)

Transactions.md
```diff
@@ -1,182 +1,292 @@
...
+* The offerer output may be below the dust limit.
+* The accepter output may be below the dust limit.
+
+Thus, a simplified formula for *expected weight* is used, which assumes:
+
+* Signatures are 72 bytes long.
+* There are a small number of outputs (thus 1 byte to count them).
+* There are always both an offerer output and an accepter output.
+
+This yields the following *expected weights* (details of the computation below):
```
Christewart, line 109: nit: Link to the section in your comment here

Transactions.md
```diff
@@ -1,182 +1,292 @@
...
+	- OP_0: 1 byte
+	- OP_DATA: 1 byte (public_key_HASH160 length)
+	- public_key_HASH160: 20 bytes
+
+p2wsh: 34 bytes
+	- OP_0: 1 byte
+	- OP_DATA: 1 byte (witness_script_SHA256 length)
+	- witness_script_SHA256: 32 bytes
+
+change_output: 9 + change_spk_script length bytes
```
Christewart: I think it's reasonable to restrict this to 

1. p2sh script
2. p2wsh script
3. p2wpkh script 

I only say p2sh because the funding transaction accepts p2sh(segwit) scripts

Transactions.md
```diff
@@ -1,182 +1,292 @@
...
+	- value: 8 bytes
+	- var_int: 1 byte (pk_script length)
+	- pk_script: change_spk_script length
+
+funding_output: 43 bytes
+	- value: 8 bytes
+	- var_int: 1 byte (pk_script length)
+	- pk_script (p2wsh): 34 bytes
+
+funding_input: 41 bytes
```
Christewart, line 139: nit: Inputs should be before outputs IMO.

Transactions.md
```diff
@@ -1,182 +1,292 @@
...
+	- previous_out_point: 36 bytes
+		- hash: 32 bytes
+		- index: 4 bytes
+	- var_int: 1 byte (script_sig length)
+	- script_sig: 0 bytes
```
Christewart, line 144: Since this is allowed to be p2sh, scriptSig is not necessarily 0 bytes because for p2sh(segwit) you need to play the segwit scriptPubKey in the scriptSig IIRC.

Transactions.md
```diff
@@ -1,182 +1,292 @@
...
+ 			separately, and the cost for its size is smaller. So,
+ 		    the calculation of ordinary data is separated
+ 			from the witness data.
+	- sequence: 4 bytes
+
+witness_header: 2 bytes
+	- flag: 1 byte
+	- marker: 1 byte
+
+witness: 108 bytes
```
Christewart, line 156: so if I understand this correctly, we are limiting the funding of our transaction to be a p2sh(p2wpkh) or p2wpkh tx? It seems that are weight calculations are implying this any way.

Transactions.md
```diff
@@ -1,182 +1,292 @@
...
+
+witness: 222 bytes
+	- number_of_witness_elements: 1 byte
+	- nil_length: 1 byte
+	- sig_offer_length: 1 byte
+	- sig_offer: 73 bytes
+	- sig_accept_length: 1 byte
+	- sig_accept: 73 bytes
+	- witness_script_length: 1 byte
+	- witness_script (multi_sig): 71 bytes
```
Christewart, line 239: I think this may be in the wrong order? If this is the "top" of the stack, that means the first sig "popped" off is `sig_accept`, however in funding output specification you said the order has the `offer_pubkey` first?

>multi_sig: 71 bytes
	- OP_2: 1 byte
	- OP_DATA: 1 byte (offer_pubkey length)
	- offer_pubkey: 33 bytes
	- OP_DATA: 1 byte (accept_pubkey length)
	- accept_pubkey: 33 bytes
	- OP_2: 1 byte
	- OP_CHECKMULTISIG: 1 byte

Transactions.md
nkohen: I see no reason to restrict the change output https://github.com/discreetlogcontracts/dlcspecs/issues/53

Transactions.md
nkohen: https://github.com/lightningnetwork/lightning-rfc/blob/master/03-transactions.md#expected-weight-of-the-commitment-transaction

Transactions.md
nkohen: So perhaps this field needs to be renamed but I strongly believe it must be included.

I agree that the decision of what this value should be should happen at the application layer but it definitely still needs to be specified here. This value can be set to 0 for example. It can be set to the current block height, it can be set to the time at which you are expected to be back online. Regardless, it definitely must be specified to avoid ambiguity.

It doesn't provide any guarantees about the oracle but it certainly gives parties (should they set this) time to cooperate to change oracles if an oracle becomes compromised.

I don't believe that this time-lock is associated with fee concerns.

Christewart: >This value can be set to 0 for example. It can be set to the current block height, it can be set to the time at which you are expected to be back online. Regardless, it definitely must be specified to avoid ambiguity.

Ok, I agree with you on the need for it to be specified, I think it just offers faux security guarantees if we start setting it to be value in the future. It makes it harder for us as protocol developers to test, and doesn't add any security guarantees to the end user because once the oracle publishes signatures related to the event your leverage is gone.

Transactions.md
Christewart: Let's continue discussion on the issue: https://github.com/discreetlogcontracts/dlcspecs/issues/53#issuecomment-674574509

Transactions.md
Christewart: I'll assume they got it right then :+1:

Transactions.md
Tibo-lg: I'm with @Christewart on this one as I tried to express in https://github.com/discreetlogcontracts/dlcspecs/pull/59#discussion_r470360648 It's a real pain for testing, can affect user experience (in many cases you might not be able to broadcast your TX even after the oracle signature is vailable_ and I agree it doesn't provide anything valuable:

> It doesn't provide any guarantees about the oracle but it certainly gives parties (should they set this) time to cooperate to change oracles if an oracle becomes compromised.

I don't see that as a valid argument, if both parties are cooperative and/or honest, they can change oracle. If one of the party thinks the compromise of the oracle is favorable to them, they'll just take advantage of it whether there is a time lock or not, the time lock really doesn't change anything.

If the point is to set it in the past or to current block height I'm ok with it, but then I feel we should specify how that should work in the protocol (e.g. which values a client should accept or reject).

Transactions.md
ariard: I meaned base_fee as described in BOLT 3 as it describes weight of transaction multiplied by negotiated feerate before to trim dust outputs, effectively increasing the fee. But that's different with onchain DLC we don't have a intermediary commitment transaction from which to fine-grain withdraw fee base on number of HTLCs outputs. So we need to provision fee for the worst-case scenario, and I agree with you for our context `max_fee` is a good label :)

Transactions.md
ariard: A non-mature transaction isn't going to propagate (where maturity is defined as relative/absolute timelock not expiring at next block). If your transaction isn't in network mempools you can't CPFP it. So it may complexify a bit your broadcast logic as you might learn a valid witness (oracle scalar) to authorize a spend but have to cache broadcast until reaching unix/block height.

I don't think it provides any security for the reason mentioned by Tibo, unless you can get with a proof of oracle compromise and thus active immediately the refund path. Security with regards to oracle compromise is interesting to think of but beyond current convervsation IMO.

Wheres it does provide setting this value to non-far-from-current-blockheight is both as a anti-fee snipping mechanism and avoiding a fingerprint. Bitcoin Core wallet implements fingeprint but set back from few blocks ~10% of transactions to avoid fingerprinting pre-signed Coinjoins to account for slow multi-party signing rounds. It would be great if we can mask DLC in this set. We don't care about the anti-fee snipping as our fee model is likely to be a just-in-time CPFP and won't be interesting for a miner to fee snip a poor-feerate parent.

ariard:
@nkohen 
     
> Doesn't the output which contains the payout act as an anchor output that can be spent using CPFP? Or am I missing something

You right, I did review with the old model where CETs where relatively timelocked to enable punishment in case of fraudulent broadcast.

Transactions.md
nkohen: https://github.com/discreetlogcontracts/dlcspecs/issues/52

Transactions.md
nkohen: https://github.com/discreetlogcontracts/dlcspecs/issues/52

Transactions.md
```diff
@@ -1,182 +1,291 @@
...
+## Funding Inputs
+
+All funding inputs must be Segwit or nested P2SH(Segwit) in order to protect against malleability attacks.
+
+## Funding Output
+
+* The funding output script is a P2WSH to:
+
+```
+2 <offer_funding_pubkey> <accept_funding_pubkey> 2 OP_CHECKMULTISIG
```
benthecarman, line 46: Maybe we can sort these lexicographically, it should be an easy compute for each party + doesn't reveal which key is being used by each party. IIRC there was some bitcoin proposal to have all `OP_CHECKMULTISIG` keys sorted this way before, might have just been something wallets were doing for easier backups.

benthecarman: Looks like this is what LN does, would definitely be good to do this then so we can share anon set with them better.

https://github.com/lightningnetwork/lightning-rfc/blob/master/03-transactions.md#funding-transaction-output

Transactions.md
```diff
@@ -1,182 +1,291 @@
...
+```
+
+* Where both `pubkey`s are in compressed format.
+
+## Change Outputs
+
+The funding transaction's change outputs should pay to the address specified in the relevant offer/accept message. A change output's value should equal the total funding amount of that party subtracted by their total collateral, their fees for this transaction, and their fees for the largest possible closing transaction. If this value is below the dust limit of `1000 satoshis`, then that party must include additional funding in order to ensure they have a valid anchor output.
```
ariard, line 53: Ah I see you already wrote the requirement. Should we copy-paste anchor output scripts and requirements from BOLT or referring them, especially with regards to utxo set pollution prevention ? 

At least you should mention to not set the witnessScript with some third-party spendable path to avoid this one interfering with your funding confirmation.

benthecarman: See https://github.com/discreetlogcontracts/dlcspecs/issues/54#issuecomment-675853144

I don't think we need to actually specify anything for anchor outputs

Christewart:
An excellent post from @ariard on tx propogation policy development in bitcoin core, linking here so others can find and make sure we are all on the same page with things we need to consider: https://github.com/bitcoin/bitcoin/issues/19820

Transactions.md
ariard: @benthecarman You mean we consider DLC parties to always spend their change outputs even when it's a low-economic value of 1000 satoshis ? 

It just a blockchain usage gentleman's rule to avoid utxo set pollution in case of multi-party protocol, we don't have to hold to it.

Transactions.md
```diff
@@ -1,182 +1,291 @@
...
+
+## Funding Transaction Input and Output Ordering
+
+The Offerer/Initiator's inputs come first and in the order that they were listed in the Offer message, the Accepter's inputs come after this in the order they are listed in the Accept message. The funding output comes first, followed by the Offerer's change output if applicable, and then the Accepter's change output if applicable.
+
+Note that this will likely change in the future.
+
+## Funding Inputs
+
+All funding inputs must be Segwit or nested P2SH(Segwit) in order to protect against malleability attacks.
```
ariard, line 39: Even guaranteeing a party always has an anchor output to prevent witness-inflation pinning, I think the malicious party can just double-spend its announced outpoint with a high-fee, low-feerate transaction, stucking in network mempools and thus preventing confirmation of the funding.

If absolute fee of this transaction is higher than funding, it would be rejected by network mempools thus preventing any CPFP bump.

AFAICT, it's a pinning variant of the one LN is currently exposed to, low-feerate package based on concurrent states.

Transactions.md
benthecarman: Not that they always spend them but they have the option to for CPFP

Transactions.md
Christewart: What exactly is the downside of this? It seems that if this happens on the funding tx, the counter party can just double spend their outpoint the same to exit the transaction?

If your counterparty is being malicious during the funding process, you probably don't want to be in a contract with them?

Transactions.md
NicolasDorier: I think sorting need to be done as well.

Transactions.md
```diff
@@ -1,182 +1,291 @@
...
-  * Fee Rate: `FeeUnit`
-
-Where
-  - The sum of the values of each `Funding Inputs` value is at least that of its corresponding `Total Collateral`
-  - `Funding Public Key`s are both 33-byte compressed public keys
-### <a name="FundingGlobal">Global</a>
-  * nLockTime is `0`
-### <a name="FundingInputs">Inputs</a>
-  * Local Funding Inputs
-      * All should have nSequence of `0xffffffff`
```
NicolasDorier, line 26: is it still true after your changes?

Transactions.md
NicolasDorier: All those fee calculations are really a rabbit hole of petty details that can go wrong.
In the [payjoin proposal](https://github.com/bitcoin/bips/blob/master/bip-0078.mediawiki), I had a similar issue, and the fee calculation was going to take like 60% of the complexity of the protocol, and responsible for 90% of the bikeshedding during review.

I ended up managing to remove the problem entirely by letting the flexibility of choosing what fee to pay for all parties. All parties decide what they want to pay.

In the case nobody wants to pay any fee, and that would make the transaction unbroadcastable, then there is just an error.

In most case, there is no need to maintain the same feerate across funding, CET and refund. Why forcing it? This make the protocol more complicated.
On top of it, it can actually be dangerous! You want the CET to have a higher fee rate than the refund. If the mempool is full and they both have the same feerate and the CET can't propagate on time, the refund would cancel it.

Transactions.md
```diff
@@ -1,182 +1,291 @@
...
+
+Note that this will likely change in the future.
+
+## Funding Inputs
+
+All funding inputs must be Segwit or nested P2SH(Segwit) in order to protect against malleability attacks.
+
+## Funding Output
+
+* The funding output script is a P2WSH to:
```
NicolasDorier, line 43: what about calculation of the value?

Transactions.md
benthecarman: Turns out this is a BIP

https://github.com/bitcoin/bips/blob/master/bip-0067.mediawiki

Transactions.md
Tibo-lg: I have the same thought as @Christewart I don't really see the big issue here. You shouldn't consider the contract established until the fund transaction is confirmed.

ariard: @Christewart, this happens post-funding process, when all signatures have been exchanged.

> It seems that if this happens on the funding tx, the counter party can just double spend their outpoint the same to exit the transaction?

In theory yes. In practice:
* Bob finds Alice full-node
* Bob mempools-split Alice's full-node from the rest of the network, feeding his high-fee, low-feerate double-spend to Alice, the funding_tx to others
* Bob's double spend doesn't propagate to the rest of network (just provoke a conflict on one of the input)
* Alice broadcast the funding tx, it's rejected _by her mempool_ as Bob's double spend doesn't signal RBF
* After few blocks, Alice try to double-spend her collateral input, her transaction is accepted locally, rejected by the rest of the network
* If DLC outcome isn't favorable to Bob, Bob wait until refund tx maturation and confirm funding_tx with a bump at that time

This assume full-node mapping and rapid feerate change to stuck the funding so it's assuming a lot on attacks resources. But AFAICT it formally works.

A good mitigation is to relative-timelock the refund tx to give opportunity to a honest party to commit CETs in case of malicious funding tx mempool-pinning. Also we should ensure that funding_tx feerate is competitive to hinder pinning.

Tibo-lg: @ariard Thanks for the explanation, it's really interesting, but I still don't understand everything:
* How does having anchor output solves the problem here? If Alice's mempool rejected the fund transaction, how could she use her anchor output to bump the fees? Could you describe what she should do assuming that she notices the issue?
* How come is Alice's double spend on the fund transaction rejected by the rest of the network, while a potential double spend from Bob is not (I assume that's what bob would do if outcome is unfavorable?).

Transactions.md
```diff
@@ -1,182 +1,291 @@
...
+
+Also known as a CET.
+
+* version: 2
+* locktime: `contract_maturity_bound`
+* txin count: 1
+  * `txin[0]` outpoint: `txid` of funding transaction and `output_index` 0
+  * `txin[0]` sequence: 0xFFFFFFFE
+  * `txin[0]` script bytes: 0
+  * `txin[0]` witness: `0 <signature_for_offer_pubkey> <signature_for_accept_pubkey>`
```
benthecarman, line 65: When doing the pubkey ordering, make sure to change here as well

Transactions.md
```diff
@@ -1,182 +1,291 @@
...
+
+If either party receives less than the dust limit of `1000 satoshis` for this outcome, then their output is not produced.
+
+### Offerer Output
+
+This output sends funds won by the offerer corresponding to this CET's outcome to the offerer's final address specified in the offer message.
+
+### Accepter Output
+
+This output sends funds won by the accepter corresponding to this CET's outcome to the accepter's final address specified in the accept message.
```
benthecarman, line 85: Should probably make some reference to the `contract_info` here for where to get the amounts

benthecarman (approved):
I think this is ready to merge and be iterated on

Transactions.md
Tibo-lg: From our last meeting, my understanding of @ariard explanation is that the main reason to enforce change output was to enable fee bumping a fund tx with CPFP without requiring to resign a tx which would have higher fee. Assuming that my understanding is correct, I would propose not in including this requirement in the first version of these specification since we have already so much to implement.

nkohen:
Relevant open issues:

https://github.com/discreetlogcontracts/dlcspecs/issues/71
https://github.com/discreetlogcontracts/dlcspecs/issues/60
https://github.com/discreetlogcontracts/dlcspecs/issues/58
https://github.com/discreetlogcontracts/dlcspecs/issues/53
https://github.com/discreetlogcontracts/dlcspecs/issues/52
https://github.com/discreetlogcontracts/dlcspecs/issues/18
https://github.com/discreetlogcontracts/dlcspecs/issues/12
https://github.com/discreetlogcontracts/dlcspecs/issues/11

PR #55, opened 2020-08-07
oracle proposal

Tibo-lg:
This is a proposal for oracle specification.

PR #59, opened 2020-08-13
Peer Protocol for Contract Negotiation Specification

nkohen:
Re-wrote the Transactions specification to match the new adaptor signature based DLCs and to match the BOLT style.

It is not in its final state but it should be a good starting point to get merged in!

Related to #46

TODO:

- [ ] Test Vectors

### DIFF
--- a/Messaging.md
+++ b/Messaging.md
@@ -1,0 +1,63 @@
+# Messaging Protocol
+
+## Overview
+
+This protocol assumes an underlying authenticated and ordered transport mechanism that takes care of framing individual messages.
+[BOLT #8](08-transport.md) specifies the canonical transport layer used in Lightning, though it can be replaced by any transport that fulfills the above guarantees.
+
+All data fields are unsigned big-endian unless otherwise specified.
+
+## Table of Contents
+
+* [Connection Handling and Multiplexing](#connection-handling-and-multiplexing)
+  * [Message Format](#message-format)
+  * [Fundamental Types](#fundamental-types)
+  * [Authors](#authors)
+
+## Connection Handling and Multiplexing
+
+Implementations MUST use a single connection per peer; contract messages (which include a contract ID) are multiplexed over this single connection.
+
+## Message Format
+
+We reuse the [Lightning Message Format](01-messaging.md#lightning-message-format) and the [Type-Length-Value Format](01-messaging.md#type-length-value-format)
+
+## Fundamental Types
+
+Various fundamental types are referred to in the message specifications:
+
+* `byte`: an 8-bit byte
+* `u16`: a 2 byte unsigned integer
+* `u32`: a 4 byte unsigned integer
+* `u64`: an 8 byte unsigned integer
+
+Inside TLV records which contain a single value, leading zeros in
+integers can be omitted:
+
+* `tu16`: a 0 to 2 byte unsigned integer
+* `tu32`: a 0 to 4 byte unsigned integer
+* `tu64`: a 0 to 8 byte unsigned integer
+
+The following convenience types are also defined:
+
+* `chain_hash`: a 32-byte chain identifier (see [BOLT #0](00-introduction.md#glossary-and-terminology-guide))
+* `contract_id`: a 32-byte contract_id (see [Protocol Specification](Protocol.md))
+* `sha256`: a 32-byte SHA2-256 hash
+* `signature`: a 64-byte bitcoin Elliptic Curve signature
+* `point`: a 33-byte Elliptic Curve point (compressed encoding as per [SEC 1 standard](http://www.secg.org/sec1-v2.pdf#subsubsection.2.3.3))
+* `spk`: A bitcoin script public key encoded as ASM prefixed with a CompactSize
+* `short_contract_id`: an 8 byte value identifying a contract (see [BOLT #7](07-routing-gossip.md#definition-of-short-channel-id))
+* `bigsize`: a variable-length, unsigned integer similar to Bitcoin's CompactSize encoding, but big-endian.  Described in [BigSize](01-messaging.md#appendix-a-bigsize-test-vectors).
+* `contract_info`: ???
+* `oracle_info`: ???
+* `funding_input`: ???
+* `cet_signatures`: ???
+* `funding_signatures`: ???
+
+## Authors
+
+Nadav Kohen <nadavk25@gmail.com>
+
+![Creative Commons License](https://i.creativecommons.org/l/by/4.0/88x31.png "License CC-BY")
+<br>
+This work is licensed under a [Creative Commons Attribution 4.0 International License](http://creativecommons.org/licenses/by/4.0/).
--- a/Messaging.md
+++ b/Messaging.md
@@ -3,7 +3,7 @@
 ## Overview
 
 This protocol assumes an underlying authenticated and ordered transport mechanism that takes care of framing individual messages.
-[BOLT #8](08-transport.md) specifies the canonical transport layer used in Lightning, though it can be replaced by any transport that fulfills the above guarantees.
+[BOLT #8](https://github.com/lightningnetwork/lightning-rfc/blob/master/08-transport.md) specifies the canonical transport layer used in Lightning, though it can be replaced by any transport that fulfills the above guarantees.
 
 All data fields are unsigned big-endian unless otherwise specified.
 
@@ -40,13 +40,13 @@ integers can be omitted:
 
 The following convenience types are also defined:
 
-* `chain_hash`: a 32-byte chain identifier (see [BOLT #0](00-introduction.md#glossary-and-terminology-guide))
+* `chain_hash`: a 32-byte chain identifier (see [BOLT #0](https://github.com/lightningnetwork/lightning-rfc/blob/master/00-introduction.md#chain_hash))
 * `contract_id`: a 32-byte contract_id (see [Protocol Specification](Protocol.md))
 * `sha256`: a 32-byte SHA2-256 hash
 * `signature`: a 64-byte bitcoin Elliptic Curve signature
 * `point`: a 33-byte Elliptic Curve point (compressed encoding as per [SEC 1 standard](http://www.secg.org/sec1-v2.pdf#subsubsection.2.3.3))
 * `spk`: A bitcoin script public key encoded as ASM prefixed with a Bitcoin CompactSize unsigned integer
-* `short_contract_id`: an 8 byte value identifying a contract funding transaction on-chain (see [BOLT #7](07-routing-gossip.md#definition-of-short-channel-id))
+* `short_contract_id`: an 8 byte value identifying a contract funding transaction on-chain (see [BOLT #7](https://github.com/lightningnetwork/lightning-rfc/blob/master/07-routing-gossip.md#definition-of-short-channel-id))
 * `bigsize`: a variable-length, unsigned integer similar to Bitcoin's CompactSize encoding, but big-endian.  Described in [BigSize](01-messaging.md#appendix-a-bigsize-test-vectors).
 * `contract_info`: ???
 * `oracle_info`: ???
--- a/Messaging.md
+++ b/Messaging.md
@@ -12,7 +12,7 @@ All data fields are unsigned big-endian unless otherwise specified.
 * [Connection Handling and Multiplexing](#connection-handling-and-multiplexing)
   * [Message Format](#message-format)
   * [Fundamental Types](#fundamental-types)
-  * [Authors](#authors)
+* [Authors](#authors)
 
 ## Connection Handling and Multiplexing
 
@@ -45,8 +45,8 @@ The following convenience types are also defined:
 * `sha256`: a 32-byte SHA2-256 hash
 * `signature`: a 64-byte bitcoin Elliptic Curve signature
 * `point`: a 33-byte Elliptic Curve point (compressed encoding as per [SEC 1 standard](http://www.secg.org/sec1-v2.pdf#subsubsection.2.3.3))
-* `spk`: A bitcoin script public key encoded as ASM prefixed with a CompactSize
-* `short_contract_id`: an 8 byte value identifying a contract (see [BOLT #7](07-routing-gossip.md#definition-of-short-channel-id))
+* `spk`: A bitcoin script public key encoded as ASM prefixed with a Bitcoin CompactSize unsigned integer
+* `short_contract_id`: an 8 byte value identifying a contract funding transaction on-chain (see [BOLT #7](07-routing-gossip.md#definition-of-short-channel-id))
 * `bigsize`: a variable-length, unsigned integer similar to Bitcoin's CompactSize encoding, but big-endian.  Described in [BigSize](01-messaging.md#appendix-a-bigsize-test-vectors).
 * `contract_info`: ???
 * `oracle_info`: ???
--- a/Protocol.md
+++ b/Protocol.md
@@ -1,16 +1,37 @@
-# DLC Protocol
+# Peer Protocol for Contract Negotiation
 
-This document describes a protocol for two parties to set up a Discreet Log Contract (DLC).
+# Table of Contents
 
-## Preliminaries
+  * [Contract](#contract)
+      * [Definition of `contract_id`](#definition-of-contract_id)
+      * [Contract Negotiation](#contract-negotiation)
+          * [The `offer_dlc` Message](#the-offer_dlc-message)
+          * [The `accept_dlc` Message](#the-accept_dlc-message)
+          * [The `sign_dlc` Message](#the-sign_dlc-message)
+* [Authors](#authors)
 
-The following protocol assumes that both parties have already agreed on the contract terms (the possible outputs of the contract).
-Negotiation of the contract, or decisions on the underlying asset are thus not part of the exchanged information.
+# Contract
 
-## Description
+## Definition of `contract_id`
 
-The two parties exchange a set of three messages, and terminates when the non-initiating party broadcasts the Fund Transaction to the Bitcoin network.
-The following figure gives a brief overview of the protocol:
+Most messages use a `contract_id` to identify the contract. It's
+derived from the funding transaction by combining the `funding_txid`
+and the `funding_output_index`, using big-endian exclusive-OR
+(i.e. `funding_output_index` alters the last 2 bytes).
+
+Prior to a contract being accepted, a `temporary_contract_id` is used,
+which is the SHA256 hash of the offer message.
+
+## Contract Negotiation
+
+Contract Negotiation consists of the initiator (aka offerer) sending an `offer_dlc` message,
+followed by the responding node (aka accepter) sending `accept_dlc`. With the
+contract parameters locked in, both parties are able to create the funding
+transaction and subsequently all contract execution transactions (CETs) and the refund transaction, as described in the [transaction specification](Transactions.md). As such, the accepter includes its signatures of the CETs and refund transaction in the `accept_dlc` message.
+The initiator now able to generate signatures for all CETs and the refund transaction, as well as the funding transaction, and send it over using the `sign_dlc` message.
+
+Once the accepter receives the `sign_dlc` message, it
+must broadcast the funding transaction to the Bitcoin network.
 
     +-------+                    +-------+
     |       |                    |       |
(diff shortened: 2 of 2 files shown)


Messaging.md
```diff
@@ -0,0 +1,63 @@
...
+* `chain_hash`: a 32-byte chain identifier (see [BOLT #0](00-introduction.md#glossary-and-terminology-guide))
+* `contract_id`: a 32-byte contract_id (see [Protocol Specification](Protocol.md))
+* `sha256`: a 32-byte SHA2-256 hash
+* `signature`: a 64-byte bitcoin Elliptic Curve signature
+* `point`: a 33-byte Elliptic Curve point (compressed encoding as per [SEC 1 standard](http://www.secg.org/sec1-v2.pdf#subsubsection.2.3.3))
+* `spk`: A bitcoin script public key encoded as ASM prefixed with a CompactSize
+* `short_contract_id`: an 8 byte value identifying a contract (see [BOLT #7](07-routing-gossip.md#definition-of-short-channel-id))
+* `bigsize`: a variable-length, unsigned integer similar to Bitcoin's CompactSize encoding, but big-endian.  Described in [BigSize](01-messaging.md#appendix-a-bigsize-test-vectors).
+* `contract_info`: ???
+* `oracle_info`: ???
```
benthecarman, line 52: Can probably port over the way we do these in bitcoin-s?

Protocol.md
```diff
@@ -23,188 +44,182 @@ The following figure gives a brief overview of the protocol:
...
+  - set `contract_maturity_cltv` and `contract_timeout_cltv` to either both be UNIX timestamps, or both be block heights as distinguished in [BIP 65](https://github.com/bitcoin/bips/blob/master/bip-0065.mediawiki).
+  - set `contract_maturity_cltv` to be less than `contract_timeout_cltv`.
 
-#### Refund locktime
-The locktime to be used for the refund transaction.
-It should be set at a later date than the maturity date of the contract.
+The sending node SHOULD:
 
----
+  - set `feerate_per_vb` to at least the rate it estimates would cause the transaction to be immediately included in a block.
```
benthecarman, line 122: Lightning uses [`feerate_per_kw`](https://github.com/lightningnetwork/lightning-rfc/blob/master/03-transactions.md#fee-calculation). Maybe we should do the same here, especially since for the txs we are giving specifications in weights. It should hopefully require less conversions when reasoning about then

Protocol.md
```diff
@@ -23,188 +44,182 @@ The following figure gives a brief overview of the protocol:
...
-#### B's addresses
+The receiving node MUST reject the contract if:
 
-The following addresses:
+  - the `chain_hash` value is set to a hash of a chain that is unknown to the receiver.
+  - the `contract_info` refers to events unknown to the receiver.
+  - the `oracle_info` refers to an oracle unknown to the receiver.
+  - it considers `feerate_per_vb` too small for timely processing or unreasonably large.
+  - `funding_pubkey` is not a valid secp256k1 pubkey in compressed format.
+  - the funder's amount for the funding transaction is not sufficient for full [fee payment](Transactions.md#fee-payment).
```
benthecarman: > not sufficient for full fee payment

Maybe "not sufficient for their full fee payment", otherwise it could imply they need to be able to pay the entirety of fees

Protocol.md
```diff
@@ -23,188 +44,182 @@ The following figure gives a brief overview of the protocol:
...
-#### Refund signature
-A signature from A for the refund transaction input.
+Other fields have the same requirements as their counterparts in `offer_dlc`.
 
-### Broadcast
-Upon receiving the `sign` message from A, B [validates](#sign-validation) its content, signs the fund transaction and broadcasts it to the Bitcoin network.
+### The `sign_dlc` Message
 
-## Validation
+This message gives all of the initiator's the signatures, which allows the receiver
```
benthecarman: > This message gives all of the initiator's the signatures

This message gives all of the initiator's signatures

Protocol.md
```diff
@@ -23,188 +44,182 @@ The following figure gives a brief overview of the protocol:
...
-* Refund locktime: confirm that it corresponds to what was agreed upon.
+#### Requirements
 
-#### Machine validation
+The sender MUST:
 
-* A fund amount: given this and B fund amount, need to validate that the amount incorporates the required fees for the CET or refund transaction as well as the closing transaction, and that the total input amount is greater than the payout amounts plus the fees to be paid.
-* A public key: validate that the public key is well formed.
-* Alice inputs: validate that the referenced outputs are in fact unspent and are either P2WSH or P2WPKH.
+  - set `channel_id` by exclusive-OR of the `funding_txid` and the `funding_output_index` from the `offer_dlc` and `accept_dlc` messages.
```
benthecarman: > `channel_id`

`contract_id`

Protocol.md
```diff
@@ -23,188 +44,182 @@ The following figure gives a brief overview of the protocol:
...
-### Sign
+The sender MUST:
 
-After receiving and [validating](#sign-validation) the content of the `offer` message, A replies with a `sign` message containing:
-1. Fund transaction signatures
-1. CET signatures
-1. Refund signature
+  - set `total_collateral_satoshis` sufficiently large so that the sum of both parties' total collaterals is at least as large as the largest payout in the `offer_dlc`'s `contract_info`.
+  - set `cet_signatures` to valid adaptor signatures, using its `funding_pubkey` for each CET, as defined in the [transaction specification](Transactions.md#contract-execution-transaction) and using signature public keys computed using the `offer_dlc`'s `contract_info` and `oracle_info` as adaptor points.
+  - include an adaptor signature in `cet_signatures` for every event specified in the `offer_dlc`'s `contract_info`.
```
benthecarman, line 177: should add the `refund_signature`. The one from `sign_dlc` can just be copy-pasta'd

Protocol.md
```diff
@@ -23,188 +44,182 @@ The following figure gives a brief overview of the protocol:
...
-### Accept validation
+The recipient:
 
-#### User validation
-* B fund amount: confirm that the fund amount matches what was agreed upon.
+  - if any `signature` is incorrect:
+    - MUST reject the contract.
+  - MUST NOT broadcast the funding transaction before receipt of a valid `sign_dlc`.
+  - on receipt of a valid `sign_dlc`:
+    - SHOULD broadcast the funding transaction.
```
benthecarman, line 229: Maybe add something about if the Initiator doesn't see the funding tx in a timely manner, they can exit the DLC by self-spending one of their funding inputs.

Protocol.md
```diff
@@ -23,188 +44,182 @@ The following figure gives a brief overview of the protocol:
...
-If this transaction gets included in the blockchain prior to the fund transaction, the contract will be canceled as this will have rendered the fund transaction invalid.
-In the event that the fund transaction gets included first, both parties will be bound to the contract.
+Nadav Kohen <nadavk25@gmail.com>
 
-# Authors
-Takatoshi Nakagawa
-Ichiro Kuwahara
-Thibaut Le Guilly
-Nadav Kohen
+[ FIXME: Add Authors ]
```
benthecarman, line 234: still need the FIXME?

Messaging.md
```diff
@@ -0,0 +1,63 @@
...
+[BOLT #8](08-transport.md) specifies the canonical transport layer used in Lightning, though it can be replaced by any transport that fulfills the above guarantees.
+
+All data fields are unsigned big-endian unless otherwise specified.
+
+## Table of Contents
+
+* [Connection Handling and Multiplexing](#connection-handling-and-multiplexing)
+  * [Message Format](#message-format)
+  * [Fundamental Types](#fundamental-types)
+  * [Authors](#authors)
```
Tibo-lg: I don't think you need indent here?

Messaging.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Connection Handling and Multiplexing
+
+Implementations MUST use a single connection per peer; contract messages (which include a contract ID) are multiplexed over this single connection.
```
Tibo-lg, line 19: I can see that comes from BOLT 1, but in my experience it's not that trivial to achieve in practice if you have two peers that both try to connect with each others. For example, is there anything specified on what happens if they end up with two connections? How do they decide which one to drop?

Messaging.md
```diff
@@ -0,0 +1,63 @@
...
+
+* `chain_hash`: a 32-byte chain identifier (see [BOLT #0](00-introduction.md#glossary-and-terminology-guide))
+* `contract_id`: a 32-byte contract_id (see [Protocol Specification](Protocol.md))
+* `sha256`: a 32-byte SHA2-256 hash
+* `signature`: a 64-byte bitcoin Elliptic Curve signature
+* `point`: a 33-byte Elliptic Curve point (compressed encoding as per [SEC 1 standard](http://www.secg.org/sec1-v2.pdf#subsubsection.2.3.3))
+* `spk`: A bitcoin script public key encoded as ASM prefixed with a CompactSize
+* `short_contract_id`: an 8 byte value identifying a contract (see [BOLT #7](07-routing-gossip.md#definition-of-short-channel-id))
+* `bigsize`: a variable-length, unsigned integer similar to Bitcoin's CompactSize encoding, but big-endian.  Described in [BigSize](01-messaging.md#appendix-a-bigsize-test-vectors).
+* `contract_info`: ???
```
Tibo-lg, line 51: I'm not sure that should be a convenience type, it seems to me more something to be represented as a set of TLV.

Messaging.md
Tibo-lg: I think it might depend on the discussion on oracle specifications. It could end up just being a URI.

Messaging.md
```diff
@@ -0,0 +1,63 @@
...
+* `oracle_info`: ???
+* `funding_input`: ???
+* `cet_signatures`: ???
+* `funding_signatures`: ???
```
Tibo-lg, line 55: I'm not sure these should be convenience types either. Especially the last 2 are just array of `signature` aren't they?

Protocol.md
```diff
@@ -1,16 +1,37 @@
-# DLC Protocol
+# Peer Protocol for Contract Negotiation
 
-This document describes a protocol for two parties to set up a Discreet Log Contract (DLC).
+# Table of Contents
 
-## Preliminaries
+  * [Contract](#contract)
```
Tibo-lg, line 5: detail, but I'm not sure I see the point of this top level section.

Protocol.md
```diff
@@ -1,16 +1,37 @@
...
 
-The two parties exchange a set of three messages, and terminates when the non-initiating party broadcasts the Fund Transaction to the Bitcoin network.
-The following figure gives a brief overview of the protocol:
+Most messages use a `contract_id` to identify the contract. It's
+derived from the funding transaction by combining the `funding_txid`
+and the `funding_output_index`, using big-endian exclusive-OR
+(i.e. `funding_output_index` alters the last 2 bytes).
+
+Prior to a contract being accepted, a `temporary_contract_id` is used,
+which is the SHA256 hash of the offer message.
```
Tibo-lg: I guess it makes sense to reuse the same although it's a bit of a pain...

Protocol.md
```diff
@@ -1,16 +1,37 @@
...
+
+## Contract Negotiation
+
+Contract Negotiation consists of the initiator (aka offerer) sending an `offer_dlc` message,
+followed by the responding node (aka accepter) sending `accept_dlc`. With the
+contract parameters locked in, both parties are able to create the funding
+transaction and subsequently all contract execution transactions (CETs) and the refund transaction, as described in the [transaction specification](Transactions.md). As such, the accepter includes its signatures of the CETs and refund transaction in the `accept_dlc` message.
+The initiator now able to generate signatures for all CETs and the refund transaction, as well as the funding transaction, and send it over using the `sign_dlc` message.
```
Tibo-lg: nit: should it be "is now"?

Tibo-lg: and "sends them over"? (I assume you are referring to the signatures?)

Protocol.md
Tibo-lg: Although I wonder if it's a good idea not to have anything related to the contract properties as part of the computation. One thing I can think of is two parties start to establish a contract, set the contract id, but then something fails prior to the signing message being exchanged (maybe on party gets disconnected and the protocol times out). This contract is no marked as failed in both parties application. Later they try to establish a new contract, which has a bit different parameters (maybe because BTC/USD price has changed). But because they use exactly the same UTXOs, the contract id ends up being the same.  It means that the application will overwrite the previously failed contract, even though it was different. Isn't that an issue? I feel for lightning it wouldn't matter because the channel will end up being the same, but in our case it seems to me like this is not ideal.

Protocol.md
```diff
@@ -23,188 +44,182 @@ The following figure gives a brief overview of the protocol:
...
-If both parties already have this information, transmission is unnecessary.
+Note that multiple contracts can be open in parallel, as all DLC
+messages are identified by either a `temporary_contract_id` (before the
+funding transaction is created) or a `contract_id` (derived from the
+funding transaction).
+
+### The `offer_dlc` Message
+
+This message contains information about a node and indicates its
+desire to open a new contract. This is the first step toward creating
```
Tibo-lg: nit: "enter in a new contract"?

Messaging.md
Tibo-lg: Maybe instead of `funding_input` it should be just `input` or `utxo`? And I guess it can be defined as txid+vout?

Protocol.md
```diff
@@ -23,188 +44,182 @@ The following figure gives a brief overview of the protocol:
...
+`funding_pubkey` is the public key in the 2-of-2 multisig script of
+the funding transaction output. `payout_spk` specifies the script
+pubkey that CETs and the refund transaction should use in the sender's output.
 
-The following addresses:
+`total_collateral_satoshis` is the amount the sender is putting into the
+contract. `num_funding_inputs` is the number of funding inputs contributed by
+the sender and `funding_inputs` contains outputs, outpoints, and expected weights
+of the sender's funding inputs. `change_spk` specifies the script pubkey that funding
+change should be sent to.
```
Tibo-lg, line 101: I could be easier to define `input` as `txid+vout` and then give an expected weight over the set of inputs?

Protocol.md
```diff
@@ -23,188 +44,182 @@ The following figure gives a brief overview of the protocol:
...
-The number to use as input to the CSV in the [CETs](Transactions.md#outputs-1).
+The sending node MUST:
 
-#### CET lock time
-The number to use as `nLockTime` for the CET transactions.
-This should correspond to the maturity date of the contracts, invalidating early CET broadcasting.
+  - set undefined bits in `contract_flags` to 0.
+  - ensure the `chain_hash` value identifies the chain it wishes to open the contract within.
+  - set `funding_pubkey` to a valid secp256k1 pubkey in compressed format.
+  - set `total_collateral_satoshis` greater than or equal to 1000.
```
Tibo-lg: nit: "to a value greater or..."

Protocol.md
```diff
@@ -23,188 +44,182 @@ The following figure gives a brief overview of the protocol:
...
+   * [`oracle_info`:`oracle_info`]
+   * [`point`:`funding_pubkey`]
+   * [`spk`:`payout_spk`]
+   * [`u64`:`total_collateral_satoshis`]
+   * [`u16`:`num_funding_inputs`]
+   * [`num_funding_inputs*funding_input`:`funding_inputs`]
+   * [`spk`:`change_spk`]
+   * [`u64`:`feerate_per_vb`]
+   * [`u32`:`contract_maturity_cltv`]
+   * [`u32`:`contract_timeout_cltv`]
```
Tibo-lg, line 81: Could be good to add a TLV stream as well like they do in LN?

Protocol.md
```diff
@@ -23,188 +44,182 @@ The following figure gives a brief overview of the protocol:
...
```
Tibo-lg: should these be `nLockTime`? And didn't we discuss not putting lock time on CETs at the end?

Protocol.md
```diff
@@ -23,188 +44,182 @@ The following figure gives a brief overview of the protocol:
...
+  - it does not want to use the oracle(s) specified in `oracle_info`.
+  - `total_collateral_satoshis` is too small.
 
-#### B's addresses
+The receiving node MUST reject the contract if:
 
-The following addresses:
+  - the `chain_hash` value is set to a hash of a chain that is unknown to the receiver.
+  - the `contract_info` refers to events unknown to the receiver.
+  - the `oracle_info` refers to an oracle unknown to the receiver.
```
Tibo-lg: At some point we'll need to clarify what goes in which. IMHO, event info should go in `oracle info`.

Protocol.md
```diff
@@ -23,188 +44,182 @@ The following figure gives a brief overview of the protocol:
...
+acceptance of the new DLC, as well as its CET and refund transaction
+signatures. This is the second step toward creating the funding transaction
+and closing transactions.
 
-#### CET signatures
-A set of signatures from B, one for each CET input.
+1. type: ??? (`accept_dlc`)
+2. data:
+   * [`32*byte`:`temporary_contract_id`]
+   * [`u64`:`total_collateral_satoshis`]
```
Tibo-lg, line 160: Isn't that implicit from the offer message? I would say it should be equal to `outcome amount - offerer collateral`?

Protocol.md
```diff
@@ -23,188 +44,182 @@ The following figure gives a brief overview of the protocol:
...
+   * [`point`:`funding_pubkey`]
+   * [`spk`:`payout_spk`]
+   * [`u16`:`num_funding_inputs`]
+   * [`num_funding_inputs*funding_input`:`funding_inputs`]
+   * [`spk`:`change_spk`]
+   * [`cet_signatures`:`cet_signatures`]
```
Tibo-lg, line 166: again I feel here the type should be `nb_cets*signature`.

Protocol.md
```diff
@@ -23,188 +44,182 @@ The following figure gives a brief overview of the protocol:
...
+The `temporary_contract_id` MUST be the SHA256 hash of the `offer_dlc` message.
 
-### Sign
+The sender MUST:
 
-After receiving and [validating](#sign-validation) the content of the `offer` message, A replies with a `sign` message containing:
-1. Fund transaction signatures
-1. CET signatures
-1. Refund signature
+  - set `total_collateral_satoshis` sufficiently large so that the sum of both parties' total collaterals is at least as large as the largest payout in the `offer_dlc`'s `contract_info`.
```
Tibo-lg, line 175: Again I'm quite confused by this. I feel the required collateral should be clear from the offer message, I don't really see in which case you would want it to be greater?

Protocol.md
```diff
@@ -23,188 +44,182 @@ The following figure gives a brief overview of the protocol:
...
+### The `sign_dlc` Message
 
-## Validation
+This message gives all of the initiator's the signatures, which allows the receiver
+to broadcast the funding transaction with both parties being fully committed to
+all closing transactions.
 
-This section describes how each party needs to validate the messages received.
-We distinguish between user validation, performed by the user (through a user interface), and machine validation, that can be done without user interaction.
+This message introduces the `contract_id` to identify the contract. It's derived from the funding transaction by combining the `funding_txid` and the `funding_output_index`, using big-endian exclusive-OR (i.e. `funding_output_index` alters the last 2 bytes).
```
Tibo-lg: You can probably just refer to the first section? This seems like repeated.

Tibo-lg (commented):
Looks nice! Put quite some comments but maybe many things are just me being confused so sorry in advance if that's the case.

Protocol.md
matthewjablack: This assumes there can't be more than one `contract_maturity` / `contract_timeout` (such as for open ended positions)

Also I guess `contract_timeout_cltv` should be optional, since it wouldn't be necessary for adapter sig DLC's

Protocol.md
```diff
@@ -23,188 +44,182 @@ The following figure gives a brief overview of the protocol:
...
+desire to open a new contract. This is the first step toward creating
+the funding transaction and CETs.
+
+1. type: ??? (`offer_dlc`)
+2. data:
+   * [`byte`:`contract_flags`]
+   * [`chain_hash`:`chain_hash`]
+   * [`contract_info`:`contract_info`]
+   * [`oracle_info`:`oracle_info`]
+   * [`point`:`funding_pubkey`]
```
matthewjablack, line 72: I'm guessing `offer_dlc` assumes counterparty has already been established (whether off-chain or through an orderbook) since counterparty pubkey is required to derive `funding_pubkey`

Messaging.md
matthewjablack: Do we need to think about multi-oracle support here?

nkohen:
Note to self: This was just updated: https://github.com/niftynei/lightning-rfc/pull/1

Messaging.md
nkohen: So I'm thinking that will be the version 0 of them but I actually want these to be true TLV types that can very between number of oracles, kinds of oracles (multi-R values for example) etc. Same with contract info and others

Protocol.md
nkohen: I agree that should be specified but I believe it belongs in its own doc similar to [BOLT 5](https://github.com/lightningnetwork/lightning-rfc/blob/master/05-onchain.md)

nkohen: Should also include fee-bumping and such in that doc

Protocol.md
nkohen: Yeah, all the BOLTs have this so I figured in this case I would put it here as unlike the tx doc others have worked on this pretty extensively previously

Messaging.md
nkohen: Yes, that is my intention (https://github.com/discreetlogcontracts/dlcspecs/pull/59#discussion_r470900331) not really sure where that stuff should go so I just put that here for now, when a PR opens that specifies v0 of this this will likely move :)

Messaging.md
nkohen: so the issue is that CET signatures may need some kind of fancy map depending on how events are structured, as such I am hoping to version/type all of these things using TLV, as I said above, not sure where those definitions are best to put yet :) (also re above comment on oracle info, I could imagine many different types existing for different kinds of contracts and compositions of oracles

Protocol.md
nkohen: This is a good point, perhaps we should include the `temporary_contract_id` in the `contract_id`? Or just part of it? Thoughts?

Protocol.md
nkohen: Note to self, look up with that new dual funding does with this

Protocol.md
nkohen: I'll have to read up more on this, if I don't get around to becoming educated on this soon I'll likely make an issue and do it on a separate dedicated PR

Protocol.md
nkohen: I vaguely remember us deciding that it could be nice to have from a user perspective to allow update/transfer if something happens to the oracle instead of incentivizing liveliness, but perhaps I am mistaken ... it was a while ago, I'll see if I can find something in meeting notes when we discussed this.
That said, they are certainly not needed but in case people want them, I think we should include it and if someone doesn't want one they can simply set the value here to 0 or the most recent block or whatever rather than making it an Option

Protocol.md
nkohen: My only concern with that is a situation where either party would intentionally want to over-collateralize (say to improve privacy in their lose cases) in which case it doesn't hurt to have this around and just use this rule you mention as the default verification?

Protocol.md
nkohen: But we need a label on each signature to point it to what outcome it signs for no? Or I guess we could use ordering explicitly (like copying the structure of the contract info) but this would need to be added to the specification if this were the case. I'll probably hold off judgement until the PR where `cet_signatures` as a TLV is defined.

Protocol.md
benthecarman: Maybe just adding a nonce to the offer can fix this

Protocol.md
benthecarman: I agree with nadav here, it should definitely be an option

Protocol.md
nkohen: I'm not sure what you mean by

> counterparty pubkey is required to derive `funding_pubkey`

as I don't believe this is true.

Regardless though, yes this entire document assumes the existence of some kind of communication layer as is stated at the top of the Messaging doc

Messaging.md
nkohen: Also funding signatures is going to have to support script signatures as opposed to just EC digital signatures so that is the complexity that still needs to be dealt with there (not a list of sigs)

Messaging.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Message Format
+
+We reuse the [Lightning Message Format](01-messaging.md#lightning-message-format) and the [Type-Length-Value Format](01-messaging.md#type-length-value-format)
+
+## Fundamental Types
+
+Various fundamental types are referred to in the message specifications:
+
+* `byte`: an 8-bit byte
```
Christewart, line 29: signed or unsigned?

Messaging.md
```diff
@@ -0,0 +1,63 @@
...
+* `tu64`: a 0 to 8 byte unsigned integer
+
+The following convenience types are also defined:
+
+* `chain_hash`: a 32-byte chain identifier (see [BOLT #0](00-introduction.md#glossary-and-terminology-guide))
+* `contract_id`: a 32-byte contract_id (see [Protocol Specification](Protocol.md))
+* `sha256`: a 32-byte SHA2-256 hash
+* `signature`: a 64-byte bitcoin Elliptic Curve signature
+* `point`: a 33-byte Elliptic Curve point (compressed encoding as per [SEC 1 standard](http://www.secg.org/sec1-v2.pdf#subsubsection.2.3.3))
+* `spk`: A bitcoin script public key encoded as ASM prefixed with a CompactSize
```
Christewart: nit : `bitcoin protocol CompactSizeUInt`

Messaging.md
```diff
@@ -0,0 +1,63 @@
...
+* `short_contract_id`: an 8 byte value identifying a contract (see [BOLT #7](07-routing-gossip.md#definition-of-short-channel-id))
```
Christewart: nit: Identifying a contract on chain?

Protocol.md
```diff
@@ -23,188 +44,182 @@ The following figure gives a brief overview of the protocol:
...
+
+This message contains information about a node and indicates its
+desire to open a new contract. This is the first step toward creating
+the funding transaction and CETs.
+
+1. type: ??? (`offer_dlc`)
+2. data:
+   * [`byte`:`contract_flags`]
+   * [`chain_hash`:`chain_hash`]
+   * [`contract_info`:`contract_info`]
```
Christewart, line 70: So how/where is "contract_info" specified? There is an assumption here that peers know how to parse/interpret contract_info? Do we need a similar spec for this just like we are talking about oracles in #55 ?

Protocol.md
Christewart: Hmm, Matt does have a good point. It might be nice to invalidate refund transactions when contract terms are updated. Perhaps we can use the Lightning penalty mechanism?

Protocol.md
```diff
@@ -23,188 +44,182 @@ The following figure gives a brief overview of the protocol:
...
 
-#### A's addresses
+`funding_pubkey` is the public key in the 2-of-2 multisig script of
+the funding transaction output. `payout_spk` specifies the script
+pubkey that CETs and the refund transaction should use in the sender's output.
 
-The following addresses:
+`total_collateral_satoshis` is the amount the sender is putting into the
+contract. `num_funding_inputs` is the number of funding inputs contributed by
+the sender and `funding_inputs` contains outputs, outpoints, and expected weights
```
Christewart, line 99: nit: Let's be consistent with `sender`/`offerer`/`maker` term (I like `maker` personally)

Protocol.md
```diff
@@ -23,188 +44,182 @@ The following figure gives a brief overview of the protocol:
...
-   - The address to use to send the change if the sum of A's inputs is greater than the fund amount plus the fees.
-2. Final Address
-   - The address to which funds will be sent in [unilateral contract execution](Transactions.md#CETOutputs), [refund](Transactions.md#refund-transaction) and [mutual closing](Transactions.md#mutual-closing-transaction) transactions.
+`feerate_per_vb` indicates the fee rate in satoshi per virtual byte that both
+sides will use to compute fees in the funding transaction, as described in the
+[transaction specification](Transactions.md).
 
-#### A's inputs
-The set of UTXOs to be used as input to the fund transaction.
+`contract_maturity_cltv` is the CLTV timelock to be put on CETs. `contract_timeout_cltv` is the CLTV timelock to be put on the refund transaction.
```
Christewart: I believe `contract_timeout_cltv` is actually a nLockTime? The funding output is a multsig output so there isn't a CLTV op in that script.

Protocol.md
```diff
@@ -23,188 +44,182 @@ The following figure gives a brief overview of the protocol:
...
+The sending node MUST:
 
-#### CET lock time
-The number to use as `nLockTime` for the CET transactions.
-This should correspond to the maturity date of the contracts, invalidating early CET broadcasting.
+  - set undefined bits in `contract_flags` to 0.
+  - ensure the `chain_hash` value identifies the chain it wishes to open the contract within.
+  - set `funding_pubkey` to a valid secp256k1 pubkey in compressed format.
+  - set `total_collateral_satoshis` greater than or equal to 1000.
+  - set `contract_maturity_cltv` and `contract_timeout_cltv` to either both be UNIX timestamps, or both be block heights as distinguished in [BIP 65](https://github.com/bitcoin/bips/blob/master/bip-0065.mediawiki).
```
Christewart: I've expressed this before but I want to re-iterate again, I think the usage of wall clock time is a bad idea and makes the protocol harder to reason about. I'm of the opinion we should only allow block heights and let application developers figure out the best representation for their users of the concept of time on a distributed system (blockchain).

This also makes our jobs easier in reasoning about the safety, security and testing the protocol.

Protocol.md
```diff
@@ -23,188 +44,182 @@ The following figure gives a brief overview of the protocol:
...
+  - set `contract_maturity_cltv` to be less than `contract_timeout_cltv`.
 
-#### Refund locktime
-The locktime to be used for the refund transaction.
-It should be set at a later date than the maturity date of the contract.
+The sending node SHOULD:
 
----
+  - set `feerate_per_vb` to at least the rate it estimates would cause the transaction to be immediately included in a block.
+  - set `contract_maturity_cltv` to the earliest expected oracle signature time.
```
Christewart: As I said in https://github.com/discreetlogcontracts/dlcspecs/pull/57#discussion_r471023249 i think this is a bad idea as it adds complexity and no real security benefits IMO.

Protocol.md
```diff
@@ -23,188 +44,182 @@ The following figure gives a brief overview of the protocol:
...
+  - `total_collateral_satoshis` is too small.
 
-#### B's addresses
+The receiving node MUST reject the contract if:
 
-The following addresses:
+  - the `chain_hash` value is set to a hash of a chain that is unknown to the receiver.
+  - the `contract_info` refers to events unknown to the receiver.
+  - the `oracle_info` refers to an oracle unknown to the receiver.
+  - it considers `feerate_per_vb` too small for timely processing or unreasonably large.
```
Christewart, line 146: Maybe this is implied here, but if the number of `funding_inputs` is something ridiculous you should reject as well.

Protocol.md
Christewart: >My only concern with that is a situation where either party would intentionally want to over-collateralize (say to improve privacy in their lose cases) in which case it doesn't hurt to have this around and just use this rule you mention as the default verification?

I think this is a debate for something after v0.1, it adds unnecessary complexity for v0.1 and makes the rest of the spec harder to reason about IMO. Can we agree to tabling this for now and filing an issue to come back to it later?

Protocol.md
Christewart: To add to this, we should probably say "refers to an oracle that is unknown to the receiver, or the provided oracle information is invalid".

Protocol.md
Christewart: nit,: `initiator` -> `offerer`/`maker` etc

Protocol.md
```diff
@@ -23,188 +44,182 @@ The following figure gives a brief overview of the protocol:
...
+  - include a signature in `funding_signatures` for every funding input specified in the`offer_dlc` message.
 
-### Accept validation
+The recipient:
 
-#### User validation
-* B fund amount: confirm that the fund amount matches what was agreed upon.
+  - if any `signature` is incorrect:
+    - MUST reject the contract.
+  - MUST NOT broadcast the funding transaction before receipt of a valid `sign_dlc`.
```
Christewart, line 226: what happens if this is not followed?

Messaging.md
nkohen: this question doesn't apply to `byte`, it isn't a number, just 8 bits.

Protocol.md
nkohen: see https://github.com/discreetlogcontracts/dlcspecs/pull/59#discussion_r470261234

Protocol.md
nkohen: I don't know what you mean ... updating contract terms requires an on-chain transaction spending the funding transaction which invalidates everything?

Protocol.md
nkohen: I disagree, I don't perceive any real addition to complexity, there's no reason not to include this in the message if there are use cases that have this determined implicitly and those that don't. Making it explicit is not an added complexity.

Protocol.md
nkohen: I think `maker` is semantically a higher-level term which would be out of place here.

Regardless, In the BOLTs they use different terminology at the beginning when speaking about the protocol and then use sender and receiver terminology per message definition, I have followed this convention here.

Protocol.md
nkohen: Firstly, I'll note that this has come up in multiple spec meetings and we have always ended up agreeing that wall clock time is the only way to go for practical applications. (e.g. reasoning about the price of bitcoin in USD at a given block height is pretty silly)

Regardless this is not an actual security concern since this timelock is just to make UX and requirements better but in no exact way (e.g. they don't have to be online and also have time to update terms if an oracle loses its signing keys).

Protocol.md
nkohen: This has to be specified implicitly or explicitly regardless. I'm open to changing the name but I see no reason for this not to be included in an offer.

Protocol.md
nkohen: The only time this would lead to a rejection is if the resulting funding transaction would not be valid so I'm not sure I need to say this? Will double check against Lisa's spec.

Protocol.md
nkohen: I think "the provided oracle information is invalid" is too ambiguous, I have no clue what this means.

Protocol.md
nkohen: A free option

Protocol.md
Christewart: I don't either as the oracle spec isn't complete :-). But I think we definitely want to at least indicate that you need to validate the oracle information on your own and not trust your counterparty. We can fill in the details after #55 is resolved?

Protocol.md
Christewart: I should have been more specific about what I meant by "valid". In this case I meant valid in terms of relay policy on the bitcoin p2p network. If you want more background on this you look at the [`AcceptToMemoryPool`](https://github.com/bitcoin/bitcoin/blob/b75f2ad72db6db93665c66279a4c9e8d5d89f027/src/validation.cpp#L1062) (ATMP) logic in bitcoin core.

and you should definitely checkout this excellent discussion from @ariard about relay policies on the bitcoin network in the future

https://github.com/bitcoin/bitcoin/issues/14895#issuecomment-665874764

Messaging.md
Tibo-lg: I see, makes sense thanks for the explanation!

Protocol.md
Tibo-lg: Ok I think I got what you mean and why you put it. Though how does that affect the contract outcomes then? Let's say offerer sent contract with collaterals: 1 for offerer and 1 for receiver. Outcomes are all 2. But receiver would like to put 2 as collateral for privacy as you mention. Should offerer assume that the outcomes are now 3 when constructing the CETs? If so I guess it should be specified explicitely.

Protocol.md
Tibo-lg: Yes I was assuming that the order of signatures and outcomes was corresponding, but I feel this is an acceptable assumption, as I can't think of a reason not to do that.

Protocol.md
Tibo-lg: I'm not sure I'm interpreting what @matthewjablack wrote properly, but I can think of a case with multiple sets of CETs with different maturity time for leveraged contract with margin call (if price touches a certain value the contract is closed). In that case it becomes quite messy to specify an nLockTime that would correspond to contract maturity. But I think there would still be only a single refund though. But tbh I'm not sure we should think about this too much for v0.1, as it changes quite a number of things in other places.

Protocol.md
Tibo-lg: I guess both proposals are equivalent since IIUC `temporary_contract_id` is a random nonce? 

I would really like to have an ID that can be set only by the offerer, but I can't come up with something that really works properly so I guess reusing the temporary id is the best.

Protocol.md
nkohen: I plan to solve this by having `contract_info` be a one-sided POV data structure, i.e. contains the payouts to the offerer, thus the remainder of the funds (total_funding - outcome) goes to the accepter where total_funding is the sum of total_collaterals.

Thus, in the case you specified the offerer has a contract info with outcomes of two, and posts a total_collateral of 1. Now the accepter can choose either to post 1 and have things be like the offerer may be expecting or they are allowed to post 2, in which case their outputs will be either 1 or 3.

Protocol.md
matthewjablack: `funding_pubkey` is a 2-of-2 multisig Script right? So don't you need pubkey of both alice and bob to derive `funding_pubkey`?

nkohen: `funding_pubkey` is just one party's public key (which is one of the two that goes into the funding script public key)

Protocol.md
```diff
@@ -23,188 +45,183 @@ The following figure gives a brief overview of the protocol:
...
-The set of UTXOs to be used as input to the fund transaction.
+#### Requirements
 
-#### Estimatesmartfee (satoshi/vbyte)
-The fee rate to be used when computing fees for the different transactions.
+The sending node MUST:
 
-#### CET CSV Delay
-The number to use as input to the CSV in the [CETs](Transactions.md#outputs-1).
+  - set undefined bits in `contract_flags` to 0.
```
benthecarman, line 113: Above you say they should be ignored, here you say they should be set to 0. Maybe change above to say that it should expect 0 and fail otherwise?

Protocol.md
```diff
@@ -23,188 +45,183 @@ The following figure gives a brief overview of the protocol:
...
+  - ensure the `chain_hash` value identifies the chain it wishes to open the contract within.
+  - set `funding_pubkey` to a valid secp256k1 pubkey in compressed format.
+  - set `total_collateral_satoshis` to a value greater than or equal to 1000.
+  - set `contract_maturity_bound` and `contract_timeout` to either both be UNIX timestamps, or both be block heights as distinguished in [BIP 65](https://github.com/bitcoin/bips/blob/master/bip-0065.mediawiki).
```
benthecarman: BIP 65 isn't realy relevant here as we aren't using OP_CLTV maybe link to this instead https://en.bitcoin.it/wiki/NLockTime

Protocol.md
```diff
@@ -23,188 +45,183 @@ The following figure gives a brief overview of the protocol:
...
-The number to use as `nLockTime` for the CET transactions.
-This should correspond to the maturity date of the contracts, invalidating early CET broadcasting.
+The sending node SHOULD:
 
-#### Refund locktime
-The locktime to be used for the refund transaction.
-It should be set at a later date than the maturity date of the contract.
+  - set `feerate_per_kw` to at least the rate it estimates would cause the transaction to be immediately included in a block.
+  - set `contract_maturity_bound` to the earliest expected oracle signature time.
+  - set `contract_timeout` sufficiently long after `contract_maturity_bound` to allow for late oracle signatures and other delays to closing the contract.
```
benthecarman: > set `contract_timeout` sufficiently long after `contract_maturity_bound`

Since we defined as `contract_maturity_bound` "the earliest expected oracle signature time". Then `contract_timeout` should be defined as "sufficiently long after the latest expected oracle signature time"

Protocol.md
```diff
@@ -23,188 +45,183 @@ The following figure gives a brief overview of the protocol:
...
 
-### Accept
-After receiving the `offer` message from A, and after [validating](#offer-validation) the provided information, B creates and sends back an `accept` message containing:
-1. B's fund amount (satoshi)
-1. B's public keys
-1. B's addresses
-1. B's inputs
-1. CET signatures
-1. Refund signature
+  - ignore undefined bits in `channel_flags`.
```
benthecarman: ditto `channel_flags` comment from before

Protocol.md
```diff
@@ -23,188 +45,183 @@ The following figure gives a brief overview of the protocol:
...
-   - B's public key used in the multi-signature [funding output](Transactions.md#FundingOutputs)
-2. Sweep Public Key
-   - B's public key used in the creation of the CETs [output script](Transactions.md#CETOutputs).
+The receiving node MUST reject the contract if:
 
-#### B's addresses
+  - the `chain_hash` value is set to a hash of a chain that is unknown to the receiver.
+  - the `contract_info` refers to events unknown to the receiver.
+  - the `oracle_info` refers to an oracle unknown to the receiver.
+  - it considers `feerate_per_kw` too small for timely processing or unreasonably large.
```
benthecarman: Maybe add

- Does not have access to the oracle(s)
- Oracle has not committed to the given R value

Protocol.md
```diff
@@ -23,188 +45,183 @@ The following figure gives a brief overview of the protocol:
...
-1. CET signatures
-1. Refund signature
+The receiver:
 
-#### Fund transaction signatures
-A set of signatures from A, one (in case of P2WPKH) or several (in case of multisig P2WSH) for each of A's input to the fund transaction.
+  - if `total_collateral_satoshis` is not large enough:
+    - MAY reject the contract.
+  - if `cet_signatures` or `refund_signature` fail validation:
+    - MUST reject the contract.
```
benthecarman, line 190: Let's not do what LN did, should specify that they must verify the funding inputs have enough to pay the `total_collateral` + fees

Protocol.md
```diff
@@ -23,188 +45,183 @@ The following figure gives a brief overview of the protocol:
...
-   - B's public key used in the creation of the CETs [output script](Transactions.md#CETOutputs).
+The receiving node MUST reject the contract if:
 
-#### B's addresses
+  - the `chain_hash` value is set to a hash of a chain that is unknown to the receiver.
+  - the `contract_info` refers to events unknown to the receiver.
+  - the `oracle_info` refers to an oracle unknown to the receiver.
+  - it considers `feerate_per_kw` too small for timely processing or unreasonably large.
+  - `funding_pubkey` is not a valid secp256k1 pubkey in compressed format.
+  - the funder's amount for the funding transaction is not sufficient for their full [fee payment](Transactions.md#fee-payment).
```
benthecarman: Maybe be more explicit and say the full fee payment + their total collateral

Protocol.md
```diff
@@ -23,188 +45,183 @@ The following figure gives a brief overview of the protocol:
...
-* Refund locktime: confirm that it corresponds to what was agreed upon.
+The sender MUST:
 
-#### Machine validation
+  - set `contract_id` by exclusive-OR of the `funding_txid` and the `funding_output_index` from the `offer_dlc` and `accept_dlc` messages.
+  - set `cet_signatures` to valid adaptor signatures, using its `funding_pubkey` for each CET, as defined in the [transaction specification](Transactions.md#contract-execution-transaction) and using signature public keys computed using the `offer_dlc`'s `contract_info` and `oracle_info` as adaptor points.
+  - include an adaptor signature in `cet_signatures` for every event specified in the `offer_dlc`'s `contract_info`.
+  - set `refund_signature` to the valid signature, using its `funding_pubkey` for the refund transaction, as defined in the [transaction specification](Transactions.md#refund-transaction).
+  - set `funding_signatures` to valid input signatures.
+  - include a signature in `funding_signatures` for every funding input specified in the`offer_dlc` message.
```
benthecarman: > the`offer_dlc`

the `offer_dlc`

benthecarman: missing a space

Protocol.md
```diff
@@ -23,188 +45,183 @@ The following figure gives a brief overview of the protocol:
...
-#### B's fund amount (satoshi)
-How much B inputs into the contract.
+The receiving node MAY reject the contract if:
 
-#### B's public keys
-The following public keys:
+  - it does not agree to the terms in `contract_info`.
+  - the `contract_info` is missing relevant events.
+  - it does not want to use the oracle(s) specified in `oracle_info`.
+  - `total_collateral_satoshis` is too small.
```
benthecarman, line 140: Above you say the sending node must

> set `feerate_per_kw` to at least the rate it estimates would cause the transaction to be immediately included in a block

Should tell the receiving node to verify this

Messaging.md
```diff
@@ -0,0 +1,63 @@
...
+Inside TLV records which contain a single value, leading zeros in
+integers can be omitted:
+
+* `tu16`: a 0 to 2 byte unsigned integer
+* `tu32`: a 0 to 4 byte unsigned integer
+* `tu64`: a 0 to 8 byte unsigned integer
+
+The following convenience types are also defined:
+
+* `chain_hash`: a 32-byte chain identifier (see [BOLT #0](00-introduction.md#glossary-and-terminology-guide))
```
benthecarman: Link doesn't work, use

https://github.com/lightningnetwork/lightning-rfc/blob/master/00-introduction.md#chain_hash

benthecarman: Definition should be moved into this repo eventually however

Protocol.md
Tibo-lg: I think what you wrote above should be written in an issue so we remember it.

Messaging.md
Tibo-lg: Tracked here if I'm not mistaken: https://github.com/discreetlogcontracts/dlcspecs/issues/61

Protocol.md
Tibo-lg: Tracked here: https://github.com/discreetlogcontracts/dlcspecs/issues/62

Messaging.md
Tibo-lg: Tracked in https://github.com/discreetlogcontracts/dlcspecs/issues/69

Protocol.md
Tibo-lg: It would be nice to decide something on this. One reason I can think of to follow your approach is that we might have to split them in several messages in which case having the index can be practical.

Protocol.md
```diff
@@ -23,188 +45,183 @@ The following figure gives a brief overview of the protocol:
...
+   * [`byte`:`contract_flags`]
+   * [`chain_hash`:`chain_hash`]
+   * [`contract_info`:`contract_info`]
+   * [`oracle_info`:`oracle_info`]
+   * [`point`:`funding_pubkey`]
+   * [`spk`:`payout_spk`]
+   * [`u64`:`total_collateral_satoshis`]
+   * [`u16`:`num_funding_inputs`]
+   * [`num_funding_inputs*funding_input`:`funding_inputs`]
+   * [`spk`:`change_spk`]
```
ariard, line 77: We should make it mandatory that `payout_spk` and `change_spk` must be fresh for every DLC contract. Maybe obvious but it's always good to avoid address reuse.

Protocol.md
```diff
@@ -23,188 +45,183 @@ The following figure gives a brief overview of the protocol:
...
 
-#### A's addresses
+`total_collateral_satoshis` is the amount the sender is putting into the
+contract. `num_funding_inputs` is the number of funding inputs contributed by
+the sender and `funding_inputs` contains outputs, outpoints, and expected weights
+of the sender's funding inputs. `change_spk` specifies the script pubkey that funding
+change should be sent to.
 
-The following addresses:
+`feerate_per_kw` indicates the fee rate in satoshi per 1000-weight that both
```
ariard: An implementation should reject the feerare if it's evaluated unreasonable. Let's avoid someone asking you to burn more in fee than its own collateral.

Protocol.md
nkohen: https://github.com/discreetlogcontracts/dlcspecs/issues/72

Protocol.md
nkohen: I think that in the current state it is easiest to implement this way but it should change before we have an official v0.1: https://github.com/discreetlogcontracts/dlcspecs/issues/74

Protocol.md
nkohen: https://github.com/discreetlogcontracts/dlcspecs/issues/76

nkohen:
Relevant open issues:

https://github.com/discreetlogcontracts/dlcspecs/issues/76
https://github.com/discreetlogcontracts/dlcspecs/issues/74
https://github.com/discreetlogcontracts/dlcspecs/issues/73
https://github.com/discreetlogcontracts/dlcspecs/issues/72
https://github.com/discreetlogcontracts/dlcspecs/issues/69
https://github.com/discreetlogcontracts/dlcspecs/issues/62
https://github.com/discreetlogcontracts/dlcspecs/issues/61
https://github.com/discreetlogcontracts/dlcspecs/issues/60
https://github.com/discreetlogcontracts/dlcspecs/issues/53

PR #63, opened 2020-08-18
URI format for event ids

LLFourn:
Here's an initial proposal for event ids that are URIs.
In many respects it can seen as a counter proposal to: https://github.com/discreetlogcontracts/dlcspecs/pull/55 although I have moved closer to some of the ideas there is some respects (not adding `price` event types but just having more generic event types).

It doesn't include anything about HTTP APIs yet. It probably will eventually because it obviously lends itself to making easy HTTP APIs.

This is just for discussion of the idea and if this were to become a specification it would have to be re-structured and more consistently formatted. Hopefully it's readable!

### DIFF
--- a/oracle-uri-proposal.md
+++ b/oracle-uri-proposal.md
@@ -1,31 +1,31 @@
 # Oracle Event URI proposal
 
 This is a proposal for a _Uniform Resource Identifier_ (URI) format for oracle _event identifiers_ (event id).
-We take URI to mean a character string that unqiuely defines a resource without implying its location or how to access it.
+We take URI to mean a character string that uniquely defines a resource without implying its location or how to access it.
 From the URI style event id, a client can determine all the event outcomes attach semantics to each.
 
 ## Motivation
 
-When executing a _Discreet Log Contract_ two parties agree on an event and a propotion of coins each party will recieve for each event outcome.
+When executing a _Discreet Log Contract_ two parties agree on an event and a proportion of coins each party will receive for each event outcome.
 In addition they must also agree on an _oracle_ who will cryptographically attest to the outcome of an event.
 To do this correctly, the oracle and the two parties have precisely the same understanding of the event.
 
 A single URI string that uniquely identifies an event has the following benefits:
 
-1. Copy pasteable: A user can easily copy paste the event id into a chat app/DLC software/browerser url.
-2. Trvially serializable: It is easy to byte encode the event id for signing purposes.
-3. Human readbale: A user who is suffienctly familiar with event namespace does not need software to understand the event.
-4. Portable: If two oracles support the same event id then you can use either of the oracles without having to query the orcales to determine if they are attesting to the same event, and just as importantly, in the same way.
+1. Copy pasteable: A user can easily copy paste the event id into a chat app/DLC software/browser address bar.
+2. Trivially serializable: It is easy to byte encode the event id for signing purposes.
+3. Human readable: A user who is sufficiently familiar with event namespace does not need software to understand the event.
+4. Portable: If two oracles support the same event id then you can use either of the oracles without having to query the oracles to determine if they are attesting to the same event, and just as importantly, in the same way.
 
 ## Specification
 
 An event identifier (_event id_) is a relative URI reference (a URI without the scheme and authority) with a query string.
 The path of the URI is interpreted as a description of an event in time and space.
-The query string is interrpeted as an _measurement_ and describes a certain fact that will be observed by an oracle on the object described by the path.
+The query string is interpreted as an _measurement_ and describes a certain fact that will be observed by an oracle on the object described by the path.
 
-If applicable, the path should start with the human organisation responsible for the event.
+If applicable, the path should start with the human organization responsible for the event.
 If it is a price then the name of the exchange where the price is being measured e.g. `binance` or if it's a sporting competition then the name of the competition e.g. `NBA`.
-If there is no applicable organisation because the measurement is being taken on some natural phenomenon like the weather or the passage of time then the first segment of the path should just be the name of this phenomenon e.g. `weather` or `time`.
+If there is no applicable organization because the measurement is being taken on some natural phenomenon like the weather or the passage of time then the first segment of the path should just be the name of this phenomenon e.g. `weather` or `time`.
 
 Each following segment, should narrow the event id until it focuses on a particular event in space and time. 
 Oracles are free to define the structure for this themselves with the following considerations:
@@ -37,19 +37,19 @@ Oracles are free to define the structure for this themselves with the following 
 
 Oracles must announce their intention to attest to the outcome of an event ahead of time.
 This is done by signing a message which includes the event id.
-Additionally it usually involves commiting to some auxialliary cryptographic data (e.g. nonces) which allows the clients to _anticipate_ the final signature for any message.
+Additionally it usually involves committing to some auxiliary cryptographic data (e.g. nonces) which allows the clients to _anticipate_ the final signature for any message.
 The announcement signature is made by suffixing the event id with `!` and appending the cryptographic data.
 
 _how to encode the data is left for a future revision_
 
 ### Attestation
 
-When attesting to a particular outcome for the event the orcale appends `=` and the outcome text onto the event id and signs the result.
+When attesting to a particular outcome for the event the oracle appends `=` and the outcome text onto the event id and signs the result.
 For example, if the outcome of `NBA/s2019/2020-08-11/PHI_PHX.vs` is `PHX_win` then the oracle signs `NBA/s2019/2020-08-11/PHI_PHX.vs=PHX_win`.
 
 ### Source Identifiers
 
-If the first segment of a path contains an `@` symbol then the text that preceeds it is a _source identifier_.
+If the first segment of a path contains an `@` symbol then the text that precedes it is a _source identifier_.
 The source identifier describes a third party that the oracle is using to decide the outcome.
 This allows an oracle to explicitly state that it is vulnerable to manipulation from this party and therefore abdicate part of its usual responsibility.
 The oracle is conveying that it may blindly attest to whatever the source tells it without verifying it. 
@@ -62,22 +62,22 @@ Clients should interpret the event that the event id is describing as if the sou
 Examples:
 
 - `espn.com@NBA/s2019/2020-08-11/PHI_PHX.vs` indicates the oracle is using `espn.com` to determine the outcome of the `NBA` match (it is not watching the match itself).
-- `wttr.in@Earth/Paris/2020-08-19?weather` indicates that the oracle is using `wttr.in` to determine the weather in Paris on the date (it is not using its own weather satelite).
+- `wttr.in@Earth/Paris/2020-08-19?weather` indicates that the oracle is using `wttr.in` to determine the weather in Paris on the date (it is not using its own weather satellite).
 
 ### Event Types
 
 The _event type_ occupies the _query string_ part of the URI.
 Event types will be introduced in their own separate RFCs and all event types mentioned here are just an example.
-The event type is a kebab-case identifier which may be followed by _paramters_ enclosed in parenthesis.
+The event type is a kebab-case identifier which may be followed by _parameters_ enclosed in parenthesis.
 The parameters are a list of comma separated values.
 Each event type defines its parameters and together they determine the outcomes for the event.
 
-The event type will usually define how to interpret the preceeding path segments.
+The event type will usually define how to interpret the preceding path segments.
 For example, for a `vs` event type may require that the final path segment be in the form `<team1>_<team2>` or a `weather` event type may require the last two segments to be in the form `<location>/<date>`.
 
 We explicitly disallow the oracle to communicate what the outcomes are for a particular event explicitly.
 Outcomes are always determined implicitly from the event type.
-This is done to avoid oracles indepedently creating events with similar but not precisely the same semantics and structure.
+This is done to avoid oracles independently creating events with similar but not precisely the same semantics and structure.
 For example, one oracle might have a weather event that has outcomes `sunny`, `cloudy`, `rainy` while another may have `light-rain` in addition.
 The hope is that this restriction will encourage the public collaboration on new event types and rich client support for them.
 
@@ -88,12 +88,12 @@ Oracles who wish to control their own event type namespace partially or entirely
 
 An event that describes a cardinal quantity like a price will have a large number of discrete outcomes.
 Normally, users will have to arrange a transactions to cover each outcome even if they only care about a small subset of outcomes.
-It is possible to optimize this by splitting up the quantitiy into some function of multiple sub-outcomes.
+It is possible to optimize this by splitting up the quantity into some function of multiple sub-outcomes.
 
-For example, a price between \$0 and \$99,999 could be expresseed as `10000×e + 1000×d + 100×c + b×10 + a` wherea a..e are in the range 0..10.
-This means that two users who want to make a BTC/USDT trade with boundries of \$4,000 and \$20,000 and with increments of \$100 would only have to compute contracts for 7 + 5 + 3×5×9 = 147 possible relevant outcomes instead of the usual 100,000 outcomes. 
+For example, a price between \$0 and \$99,999 could be expressed as `10000×e + 1000×d + 100×c + b×10 + a` whereas a..e are in the range 0..10.
+This means that two users who want to make a BTC/USDT trade with boundaries of \$4,000 and \$20,000 and with increments of \$100 would only have to compute contracts for 7 + 5 + 3×5×9 = 147 possible relevant outcomes instead of the usual 100,000 outcomes. 
 
-The orcale will have to declare cryptogrpahically its intention to release siganture for each of the decimal decomposed values a..e.
+The oracle will have to declare cryptographically its intention to release signature for each of the decimal decomposed values a..e.
 We treat all these values as related "fragments" of an event rather than separate events as they are semantically related to the same object.
 
 An event type may determine that an event is made up of several event fragments.
@@ -109,12 +109,12 @@ When attesting to a notional outcome (which is composed of the outcomes of each 
 
 ### `occur`
 
-This trvial event type only has one outcome `true`.
-It can be used to mark the occurance of some event (that may or may not ever happen).
+This trial event type only has one outcome `true`.
+It can be used to mark the occurrence of some event (that may or may not ever happen).
 For example, it can be useful to mark the passage of time itself.
 
 - Example event id: `time/2020-08-16T08:00:00?occur`
-- Example outcome stirng:  `time/2020-08-16T08:00:00?occur=true` (after the event has happened)
+- Example outcome sting:  `time/2020-08-16T08:00:00?occur=true` (after the event has happened)
 
 ### `weather`
 
@@ -134,9 +134,9 @@ It has two parameters:
 - `n_digits`: the number of digits to the left of the lowest digit.
 - `least_significant_digit`: (default=0) the position the lowest digit e.g. -1 for a number like `12.3` and 0 for `123`.
 
-If the actual value is above or equal to the maximum number expressble by the digits then the oracle must attest to the maximum number.
+If the actual value is above or equal to the maximum number expressible by the digits then the oracle must attest to the maximum number.
 
-For exmaple, to attest to the bid price of Bitcoin on Binance between \$0 and \$99,999 the oracle would announce `binance/BTC_USDT/bid/2021-05-16T08:00:00?digits(5)`.
+For example, to attest to the bid price of Bitcoin on Binance between \$0 and \$99,999 the oracle would announce `binance/BTC_USDT/bid/2021-05-16T08:00:00?digits(5)`.
 If the bid price on the date is \$20,421 the oracle would release five signatures on the following messages:
 
   1. `binance/BTC_USDT/bid-price/2021-05-16T08:00:00?digits(5).0=1`
@@ -156,18 +156,18 @@ This binary event type asserts that the quantity measured will be above a certai
 
 ### `vs`
 
-This event type is for competive matches where the one team wins and one team loses or there is a draw.
+This event type is for competitive matches where the one team wins and one team loses or there is a draw.
 Competitions where drawing is impossible can still use the `vs` event type and just use the `draw` outcome to indicate the match being cancelled due to some act of god.
 
-The `vs` event type requires that the path segment preceeding it be in the form `<team1>_<team2>` where `<team1>_<team2>` are the competing teams.
+The `vs` event type requires that the path segment preceding it be in the form `<team1>_<team2>` where `<team1>_<team2>` are the competing teams.
 For example, `NBA/s2019/2020-08-11/PHI_PHX.vs` describes a team named `PHI` is playing against another team named `PHX`.
 Additionally the client can infer the date of the match.
 
 Clients that are familiar with the `NBA` namespace will be able to infer that the game is part of the 2019 season and that `PHI` refers to the Phoenix Suns and `PHX` refers to the Philadelphia 76ers.
 
 The outcomes for this event are `PHI_win`, `PHX_win` or `draw`. 
-Note that the team names are somewhat reduntantly included to make the winning team explicit rather than using something like `left-team-won`.
-Although `left-team-won` is not ambigious, it is easy to imagine a a programmer error in some part of an oracle system that does something like:
+Note that the team names are somewhat redundantly included to make the winning team explicit rather than using something like `left-team-won`.
+Although `left-team-won` is not ambiguous, it is easy to imagine a a programmer error in some part of an oracle system that does something like:
 
 ```
 if winning_team == "PHI" {
@@ -191,7 +191,7 @@ The goal of the format is to allow extension of the event space through new even
 The idea is that new event types will generally involve a lot formal collaboration between oracles through the authoring of new specification documents.
 This is because clients and servers must all implement the specific logic of the event type.
 
-On the other hand, oracles can create new namespaces indpendently.
+On the other hand, oracles can create new namespaces independently.
 Obviously it would be ideal if the namespaces are consistent across oracles but we hope that this can happen through ad-hoc cooperation.
-We expect that useful namespace conventions will emerge accross oracles and if those conventions becomes widespread enough and clients rely on them then they can be specified later.
+We expect that useful namespace conventions will emerge across oracles and if those conventions becomes widespread enough and clients rely on them then they can be specified later.
 
--- a/oracle-uri-proposal.md
+++ b/oracle-uri-proposal.md
@@ -2,15 +2,15 @@
 
 This is a proposal for a _Uniform Resource Identifier_ (URI) format for oracle _event identifiers_ (event id).
 We take URI to mean a character string that unqiuely defines a resource without implying its location or how to access it.
-From the URI style event id, a client can determine all the event outcomes attach some meaning to each event outcome.
+From the URI style event id, a client can determine all the event outcomes attach semantics to each.
 
 ## Motivation
 
-When executing a _Discreet Log Contract_ two parties agree on an event and a propotion of coins each party will recieve on each possible event outcome.
+When executing a _Discreet Log Contract_ two parties agree on an event and a propotion of coins each party will recieve for each event outcome.
 In addition they must also agree on an _oracle_ who will cryptographically attest to the outcome of an event.
-To do this correctly, the oracle and the two parties have precisely the same understanding of the event and the rules the oracle will decide each outcome.
+To do this correctly, the oracle and the two parties have precisely the same understanding of the event.
 
-If we have a single URI string that uniquely identifies an event we get the following benefits:
+A single URI string that uniquely identifies an event has the following benefits:
 
 1. Copy pasteable: A user can easily copy paste the event id into a chat app/DLC software/browerser url.
 2. Trvially serializable: It is easy to byte encode the event id for signing purposes.
@@ -19,7 +19,7 @@ If we have a single URI string that uniquely identifies an event we get the foll
 
 ## Specification
 
-An event identifier (_event id_) a relative URI reference (a URI without the scheme and authority) with a query string.
+An event identifier (_event id_) is a relative URI reference (a URI without the scheme and authority) with a query string.
 The path of the URI is interpreted as a description of an event in time and space.
 The query string is interrpeted as an _measurement_ and describes a certain fact that will be observed by an oracle on the object described by the path.
 
@@ -30,14 +30,14 @@ If there is no applicable organisation because the measurement is being taken on
 Each following segment, should narrow the event id until it focuses on a particular event in space and time. 
 Oracles are free to define the structure for this themselves with the following considerations:
 
-1. The path should follow the strucuture defined by the event type if it specifies one.
-2. Rules and conventions may be attached to namespaces formally or informally and oracles should try and follow them.
+1. The event type of the event may require the path has a certain structure.
+2. Rules and conventions may be attached to namespaces and in general oracles should try and follow them.
 
 ### Announcement
 
 Oracles must announce their intention to attest to the outcome of an event ahead of time.
 This is done by signing a message which includes the event id.
-Usually involve commiting to some auxialliary cryptographic data which allows the clients to _anticipate_ the final signature for any message.
+Additionally it usually involves commiting to some auxialliary cryptographic data (e.g. nonces) which allows the clients to _anticipate_ the final signature for any message.
 The announcement signature is made by suffixing the event id with `!` and appending the cryptographic data.
 
 _how to encode the data is left for a future revision_
@@ -57,7 +57,7 @@ The oracle is conveying that it may blindly attest to whatever the source tells 
 An oracle who does not provide a source identifier is asserting that it will take the measurement of the event itself or otherwise verify it somehow before attesting to it. 
 It is completely acceptable from the point of view of this specification for an oracle to omit a source identifier even if in reality it is only using one online source to decide the outcome; it is simply a question of whether the oracle finds it advantageous to inform its users of this fact.
 
-Clients should determine the event that the event id is describing as if the source identifier were absent.
+Clients should interpret the event that the event id is describing as if the source identifier were absent.
 
 Examples:
 
@@ -73,7 +73,7 @@ The parameters are a list of comma separated values.
 Each event type defines its parameters and together they determine the outcomes for the event.
 
 The event type will usually define how to interpret the preceeding path segments.
-For example, for a `vs` event type may imply that the final path segment be in the form `<team1>_<team2>` or a `weather` event type may require the last two segments to be in the form `<city>/<date>`.
+For example, for a `vs` event type may require that the final path segment be in the form `<team1>_<team2>` or a `weather` event type may require the last two segments to be in the form `<location>/<date>`.
 
 We explicitly disallow the oracle to communicate what the outcomes are for a particular event explicitly.
 Outcomes are always determined implicitly from the event type.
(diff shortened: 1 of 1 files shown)


oracle-uri-proposal.md
```diff
@@ -0,0 +1,197 @@
...
+When executing a _Discreet Log Contract_ two parties agree on an event and a propotion of coins each party will recieve for each event outcome.
+In addition they must also agree on an _oracle_ who will cryptographically attest to the outcome of an event.
+To do this correctly, the oracle and the two parties have precisely the same understanding of the event.
+
+A single URI string that uniquely identifies an event has the following benefits:
+
+1. Copy pasteable: A user can easily copy paste the event id into a chat app/DLC software/browerser url.
+2. Trvially serializable: It is easy to byte encode the event id for signing purposes.
+3. Human readbale: A user who is suffienctly familiar with event namespace does not need software to understand the event.
+4. Portable: If two oracles support the same event id then you can use either of the oracles without having to query the orcales to determine if they are attesting to the same event, and just as importantly, in the same way.
```
benthecarman: > orcales

oracles

oracle-uri-proposal.md
```diff
@@ -0,0 +1,197 @@
...
+
+## Motivation
+
+When executing a _Discreet Log Contract_ two parties agree on an event and a propotion of coins each party will recieve for each event outcome.
+In addition they must also agree on an _oracle_ who will cryptographically attest to the outcome of an event.
+To do this correctly, the oracle and the two parties have precisely the same understanding of the event.
+
+A single URI string that uniquely identifies an event has the following benefits:
+
+1. Copy pasteable: A user can easily copy paste the event id into a chat app/DLC software/browerser url.
```
benthecarman: > browerser

browser

oracle-uri-proposal.md
```diff
@@ -0,0 +1,197 @@
...
+3. Human readbale: A user who is suffienctly familiar with event namespace does not need software to understand the event.
+4. Portable: If two oracles support the same event id then you can use either of the oracles without having to query the orcales to determine if they are attesting to the same event, and just as importantly, in the same way.
+
+## Specification
+
+An event identifier (_event id_) is a relative URI reference (a URI without the scheme and authority) with a query string.
+The path of the URI is interpreted as a description of an event in time and space.
+The query string is interrpeted as an _measurement_ and describes a certain fact that will be observed by an oracle on the object described by the path.
+
+If applicable, the path should start with the human organisation responsible for the event.
```
benthecarman: > organisation

organization

oracle-uri-proposal.md
```diff
@@ -0,0 +1,197 @@
...
+If it is a price then the name of the exchange where the price is being measured e.g. `binance` or if it's a sporting competition then the name of the competition e.g. `NBA`.
+If there is no applicable organisation because the measurement is being taken on some natural phenomenon like the weather or the passage of time then the first segment of the path should just be the name of this phenomenon e.g. `weather` or `time`.
```
benthecarman: > organisation

organization

oracle-uri-proposal.md
```diff
@@ -0,0 +1,197 @@
+# Oracle Event URI proposal
+
+This is a proposal for a _Uniform Resource Identifier_ (URI) format for oracle _event identifiers_ (event id).
+We take URI to mean a character string that unqiuely defines a resource without implying its location or how to access it.
```
benthecarman: > unqiuely

uniquely

oracle-uri-proposal.md
```diff
@@ -0,0 +1,197 @@
...
+From the URI style event id, a client can determine all the event outcomes attach semantics to each.
+
+## Motivation
+
+When executing a _Discreet Log Contract_ two parties agree on an event and a propotion of coins each party will recieve for each event outcome.
```
benthecarman: > propotion

proportion

benthecarman: > recieve

receive

oracle-uri-proposal.md
```diff
@@ -0,0 +1,197 @@
...
+In addition they must also agree on an _oracle_ who will cryptographically attest to the outcome of an event.
+To do this correctly, the oracle and the two parties have precisely the same understanding of the event.
+
+A single URI string that uniquely identifies an event has the following benefits:
+
+1. Copy pasteable: A user can easily copy paste the event id into a chat app/DLC software/browerser url.
+2. Trvially serializable: It is easy to byte encode the event id for signing purposes.
+3. Human readbale: A user who is suffienctly familiar with event namespace does not need software to understand the event.
```
benthecarman: > readbale

readable

benthecarman: > suffienctly

sufficiently

oracle-uri-proposal.md
```diff
@@ -0,0 +1,197 @@
...
+Oracles are free to define the structure for this themselves with the following considerations:
+
+1. The event type of the event may require the path has a certain structure.
+2. Rules and conventions may be attached to namespaces and in general oracles should try and follow them.
+
+### Announcement
+
+Oracles must announce their intention to attest to the outcome of an event ahead of time.
+This is done by signing a message which includes the event id.
+Additionally it usually involves commiting to some auxialliary cryptographic data (e.g. nonces) which allows the clients to _anticipate_ the final signature for any message.
```
benthecarman: > auxialliary

auxiliary

oracle-uri-proposal.md
```diff
@@ -0,0 +1,197 @@
...
+_how to encode the data is left for a future revision_
+
+### Attestation
+
+When attesting to a particular outcome for the event the orcale appends `=` and the outcome text onto the event id and signs the result.
+For example, if the outcome of `NBA/s2019/2020-08-11/PHI_PHX.vs` is `PHX_win` then the oracle signs `NBA/s2019/2020-08-11/PHI_PHX.vs=PHX_win`.
+
+### Source Identifiers
+
+If the first segment of a path contains an `@` symbol then the text that preceeds it is a _source identifier_.
```
benthecarman: > preceeds

precedes

oracle-uri-proposal.md
```diff
@@ -0,0 +1,197 @@
...
+
+### Event Types
+
+The _event type_ occupies the _query string_ part of the URI.
+Event types will be introduced in their own separate RFCs and all event types mentioned here are just an example.
+The event type is a kebab-case identifier which may be followed by _paramters_ enclosed in parenthesis.
+The parameters are a list of comma separated values.
+Each event type defines its parameters and together they determine the outcomes for the event.
+
+The event type will usually define how to interpret the preceeding path segments.
```
benthecarman: > preceeding

preceding

oracle-uri-proposal.md
```diff
@@ -0,0 +1,197 @@
...
+For example, for a `vs` event type may require that the final path segment be in the form `<team1>_<team2>` or a `weather` event type may require the last two segments to be in the form `<location>/<date>`.
+
+We explicitly disallow the oracle to communicate what the outcomes are for a particular event explicitly.
+Outcomes are always determined implicitly from the event type.
+This is done to avoid oracles indepedently creating events with similar but not precisely the same semantics and structure.
```
benthecarman: > indepedently

independently

oracle-uri-proposal.md
```diff
@@ -0,0 +1,197 @@
...
+Oracles should communicate to clients which specifications they are following for their event types but this is outside the scope of this document.
+Oracles who wish to control their own event type namespace partially or entirely may therefore link to their own specifications defining their own event types.
+
+### Event Fragments
+
+An event that describes a cardinal quantity like a price will have a large number of discrete outcomes.
+Normally, users will have to arrange a transactions to cover each outcome even if they only care about a small subset of outcomes.
+It is possible to optimize this by splitting up the quantitiy into some function of multiple sub-outcomes.
+
+For example, a price between \$0 and \$99,999 could be expresseed as `10000×e + 1000×d + 100×c + b×10 + a` wherea a..e are in the range 0..10.
```
benthecarman: > expresseed

expressed

benthecarman: > wherea

where

oracle-uri-proposal.md
```diff
@@ -0,0 +1,197 @@
...
+This means that two users who want to make a BTC/USDT trade with boundries of \$4,000 and \$20,000 and with increments of \$100 would only have to compute contracts for 7 + 5 + 3×5×9 = 147 possible relevant outcomes instead of the usual 100,000 outcomes.
```
benthecarman: > boundries

boundaries

oracle-uri-proposal.md
```diff
@@ -0,0 +1,197 @@
...
+### Event Fragments
+
+An event that describes a cardinal quantity like a price will have a large number of discrete outcomes.
+Normally, users will have to arrange a transactions to cover each outcome even if they only care about a small subset of outcomes.
+It is possible to optimize this by splitting up the quantitiy into some function of multiple sub-outcomes.
+
+For example, a price between \$0 and \$99,999 could be expresseed as `10000×e + 1000×d + 100×c + b×10 + a` wherea a..e are in the range 0..10.
+This means that two users who want to make a BTC/USDT trade with boundries of \$4,000 and \$20,000 and with increments of \$100 would only have to compute contracts for 7 + 5 + 3×5×9 = 147 possible relevant outcomes instead of the usual 100,000 outcomes. 
+
+The orcale will have to declare cryptogrpahically its intention to release siganture for each of the decimal decomposed values a..e.
```
benthecarman: > orcale

oracle

benthecarman: > cryptogrpahically

cryptographically

benthecarman: > siganture

signature

oracle-uri-proposal.md
```diff
@@ -0,0 +1,197 @@
...
+
+When attesting to a notional outcome (which is composed of the outcomes of each fragment) each fragment message is determined by the message:
+
+`<event_id>.<i>=<fragment_outcome_i>`
+
+## Example Event Types
+
+### `occur`
+
+This trvial event type only has one outcome `true`.
```
benthecarman: > trvial

trivial

oracle-uri-proposal.md
```diff
@@ -0,0 +1,197 @@
...
+It can be used to mark the occurance of some event (that may or may not ever happen).
```
benthecarman: > occurance

occurrence

oracle-uri-proposal.md
```diff
@@ -0,0 +1,197 @@
...
+For example, it can be useful to mark the passage of time itself.
+
+- Example event id: `time/2020-08-16T08:00:00?occur`
+- Example outcome stirng:  `time/2020-08-16T08:00:00?occur=true` (after the event has happened)
```
benthecarman: > stirng

string

oracle-uri-proposal.md
```diff
@@ -0,0 +1,197 @@
...
+Each decimal digit is an event fragment i.e. the oracle will release a signature for each digit.
+
+It has two parameters:
+
+- `n_digits`: the number of digits to the left of the lowest digit.
+- `least_significant_digit`: (default=0) the position the lowest digit e.g. -1 for a number like `12.3` and 0 for `123`.
+
+If the actual value is above or equal to the maximum number expressble by the digits then the oracle must attest to the maximum number.
+
+For exmaple, to attest to the bid price of Bitcoin on Binance between \$0 and \$99,999 the oracle would announce `binance/BTC_USDT/bid/2021-05-16T08:00:00?digits(5)`.
```
benthecarman: > exmaple

example

oracle-uri-proposal.md
```diff
@@ -0,0 +1,197 @@
...
+
+A positive number decomposed into its decimal digits.
+Each decimal digit is an event fragment i.e. the oracle will release a signature for each digit.
+
+It has two parameters:
+
+- `n_digits`: the number of digits to the left of the lowest digit.
+- `least_significant_digit`: (default=0) the position the lowest digit e.g. -1 for a number like `12.3` and 0 for `123`.
+
+If the actual value is above or equal to the maximum number expressble by the digits then the oracle must attest to the maximum number.
```
benthecarman: > expressble

expressible

oracle-uri-proposal.md
```diff
@@ -0,0 +1,197 @@
...
+
+If a path ends in `<name1>_<name2>/bid-price/<datetime>` and has an event type of `digits` or `int` then the events value represents the bid price of `<name1>` in terms of `<name2>` at `<datetime>`.
+
+### `above(threshold)`
+
+This binary event type asserts that the quantity measured will be above a certain threshold.
+
+### `vs`
+
+This event type is for competive matches where the one team wins and one team loses or there is a draw.
```
benthecarman: > competive

competitive

oracle-uri-proposal.md
```diff
@@ -0,0 +1,197 @@
...
+Competitions where drawing is impossible can still use the `vs` event type and just use the `draw` outcome to indicate the match being cancelled due to some act of god.
+
+The `vs` event type requires that the path segment preceeding it be in the form `<team1>_<team2>` where `<team1>_<team2>` are the competing teams.
```
benthecarman: > preceeding

preceding

oracle-uri-proposal.md
```diff
@@ -0,0 +1,197 @@
...
+For example, `NBA/s2019/2020-08-11/PHI_PHX.vs` describes a team named `PHI` is playing against another team named `PHX`.
+Additionally the client can infer the date of the match.
+
+Clients that are familiar with the `NBA` namespace will be able to infer that the game is part of the 2019 season and that `PHI` refers to the Phoenix Suns and `PHX` refers to the Philadelphia 76ers.
+
+The outcomes for this event are `PHI_win`, `PHX_win` or `draw`. 
+Note that the team names are somewhat reduntantly included to make the winning team explicit rather than using something like `left-team-won`.
```
benthecarman: > reduntantly

redundantly

oracle-uri-proposal.md
```diff
@@ -0,0 +1,197 @@
...
+Although `left-team-won` is not ambigious, it is easy to imagine a a programmer error in some part of an oracle system that does something like:
```
benthecarman: > ambigious

ambiguous

oracle-uri-proposal.md
```diff
@@ -0,0 +1,197 @@
...
+These are just like `vs` except they assert a particular team will win and are therefore binary.
+For the event `NBA/s2019/2020-08-11/PHI_PHX.left-win` the outcomes are `PHI_win` or `PHX_win-or-draw`.
+
+## Extensibility
+
+The goal of the format is to allow extension of the event space through new event types and new namespaces.
+The idea is that new event types will generally involve a lot formal collaboration between oracles through the authoring of new specification documents.
+This is because clients and servers must all implement the specific logic of the event type.
+
+On the other hand, oracles can create new namespaces indpendently.
```
benthecarman: > indpendently

independently

oracle-uri-proposal.md
```diff
@@ -0,0 +1,197 @@
...
+Obviously it would be ideal if the namespaces are consistent across oracles but we hope that this can happen through ad-hoc cooperation.
+We expect that useful namespace conventions will emerge accross oracles and if those conventions becomes widespread enough and clients rely on them then they can be specified later.
```
benthecarman: > accross

across

benthecarman (changes requested):
Typo pass, found a bunch so I ran it through a word processor to find the rest

LLFourn:
@benthecarman Ahh Thanks for that. Apparently I don’t have spellchecking on my fresh emacs setup (I am also bad at spelling).

oracle-uri-proposal.md
LLFourn: using murican english now.

oracle-uri-proposal.md
```diff
@@ -0,0 +1,197 @@
...
+When executing a _Discreet Log Contract_ two parties agree on an event and a proportion of coins each party will receive for each event outcome.
+In addition they must also agree on an _oracle_ who will cryptographically attest to the outcome of an event.
+To do this correctly, the oracle and the two parties have precisely the same understanding of the event.
+
+A single URI string that uniquely identifies an event has the following benefits:
+
+1. Copy pasteable: A user can easily copy paste the event id into a chat app/DLC software/browser address bar.
+2. Trivially serializable: It is easy to byte encode the event id for signing purposes.
+3. Human readable: A user who is sufficiently familiar with event namespace does not need software to understand the event.
+4. Portable: If two oracles support the same event id then you can use either of the oracles without having to query the oracles to determine if they are attesting to the same event, and just as importantly, in the same way.
```
Tibo-lg: I'm not sure I understand this. How do you know they are attesting to the same event if you don't query them?

oracle-uri-proposal.md
```diff
@@ -0,0 +1,197 @@
...
+_how to encode the data is left for a future revision_
+
+### Attestation
+
+When attesting to a particular outcome for the event the oracle appends `=` and the outcome text onto the event id and signs the result.
+For example, if the outcome of `NBA/s2019/2020-08-11/PHI_PHX.vs` is `PHX_win` then the oracle signs `NBA/s2019/2020-08-11/PHI_PHX.vs=PHX_win`.
+
+### Source Identifiers
+
+If the first segment of a path contains an `@` symbol then the text that precedes it is a _source identifier_.
```
Tibo-lg: `@` is already a reserved character in URI syntax, it would probably be better not to reuse it otherwise it would make it difficult to use existing parsing libraries.

oracle-uri-proposal.md
```diff
@@ -0,0 +1,197 @@
...
+The oracle is conveying that it may blindly attest to whatever the source tells it without verifying it. 
+
+An oracle who does not provide a source identifier is asserting that it will take the measurement of the event itself or otherwise verify it somehow before attesting to it. 
+It is completely acceptable from the point of view of this specification for an oracle to omit a source identifier even if in reality it is only using one online source to decide the outcome; it is simply a question of whether the oracle finds it advantageous to inform its users of this fact.
+
+Clients should interpret the event that the event id is describing as if the source identifier were absent.
+
+Examples:
+
+- `espn.com@NBA/s2019/2020-08-11/PHI_PHX.vs` indicates the oracle is using `espn.com` to determine the outcome of the `NBA` match (it is not watching the match itself).
```
Tibo-lg: Why do you use a `.` instead of a `/` at the end? Has the `.` any special meaning?

oracle-uri-proposal.md
```diff
@@ -0,0 +1,197 @@
...
+Oracles should communicate to clients which specifications they are following for their event types but this is outside the scope of this document.
+Oracles who wish to control their own event type namespace partially or entirely may therefore link to their own specifications defining their own event types.
+
+### Event Fragments
+
+An event that describes a cardinal quantity like a price will have a large number of discrete outcomes.
+Normally, users will have to arrange a transactions to cover each outcome even if they only care about a small subset of outcomes.
+It is possible to optimize this by splitting up the quantity into some function of multiple sub-outcomes.
+
+For example, a price between \$0 and \$99,999 could be expressed as `10000×e + 1000×d + 100×c + b×10 + a` whereas a..e are in the range 0..10.
```
Tibo-lg, line 77: I considered this as well, but I think the mantissa exponent as described in the original paper is actually better because you only need 2 R-values for any case and you don't have issues of going out of bounds.

oracle-uri-proposal.md
```diff
@@ -0,0 +1,197 @@
...
+For example, to attest to the bid price of Bitcoin on Binance between \$0 and \$99,999 the oracle would announce `binance/BTC_USDT/bid/2021-05-16T08:00:00?digits(5)`.
+If the bid price on the date is \$20,421 the oracle would release five signatures on the following messages:
+
+  1. `binance/BTC_USDT/bid-price/2021-05-16T08:00:00?digits(5).0=1`
+  2. `binance/BTC_USDT/bid-price/2021-05-16T08:00:00?digits(5).1=2`
+  3. `binance/BTC_USDT/bid-price/2021-05-16T08:00:00?digits(5).2=4`
+  4. `binance/BTC_USDT/bid-price/2021-05-16T08:00:00?digits(5).3=0`
+  5. `binance/BTC_USDT/bid-price/2021-05-16T08:00:00?digits(5).4=2`
+
+If `digits(5)` turns out to be insufficient later on because the bitcoin price appreciates too much the oracle can always announce a new `digits(6)` event on the same path.
```
Tibo-lg, line 131: How can a client make use of this?

oracle-uri-proposal.md
```diff
@@ -0,0 +1,197 @@
...
+
+This binary event type asserts that the quantity measured will be above a certain threshold.
+
+### `vs`
+
+This event type is for competitive matches where the one team wins and one team loses or there is a draw.
+Competitions where drawing is impossible can still use the `vs` event type and just use the `draw` outcome to indicate the match being cancelled due to some act of god.
+
+The `vs` event type requires that the path segment preceding it be in the form `<team1>_<team2>` where `<team1>_<team2>` are the competing teams.
+For example, `NBA/s2019/2020-08-11/PHI_PHX.vs` describes a team named `PHI` is playing against another team named `PHX`.
```
Tibo-lg: Should it be a `?` instead of `.` IIUC?

Tibo-lg (commented):
I think at the end it's not really in opposition to https://github.com/discreetlogcontracts/dlcspecs/pull/55 and seems more like a layer above to me. I think it will be nice to have common standard for event types, but I also feel like it is too restrictive to assume that for an oracle to publish an event it first has to make it standard. IIUC, it means clients need to be aware of all the event types out there, so if a new event type comes out, the client needs to be updated. I'm not sure I see why that would be better than having the client receive information on how to handle the event.

oracle-uri-proposal.md
LLFourn: Sorry this is just me changing notations inconsistently it should be `?`. Thanks.

oracle-uri-proposal.md
LLFourn: This is a good point.  I think it's only reserved in the authority section strictly speaking? I'll have to experiment to see if it confuses parsers.

oracle-uri-proposal.md
LLFourn: Err woops this is wrong. I wanted to say you don't have to inspect query response bodies to check this (rather just if it's 200 OK rather than 404).

oracle-uri-proposal.md
LLFourn: I made an issue to discuss further: https://github.com/discreetlogcontracts/dlcspecs/issues/65

oracle-uri-proposal.md
LLFourn: Clients wanting to make a bet after `digits(6)` has been announced can use it.  Clients who have already made bets on `digits(5)` have to just carry on with that and its possibly truncated outcome.

oracle-uri-proposal.md
LLFourn: I ended up removing this claim for now. I think portability is a concern but I wouldn't say the event id uri is the only way to achieve it.

LLFourn:
@Tibo-lg Thanks for your review

> I think at the end it's not really in opposition to #55 and seems more like a layer above to me.

I think our ideas are less opposed than I originally thought.
This does specify a different way to serialize and combine events with other data when announcing and attesting. In your proposal it's a TBD TLV encoding where as here it's a single string for the event with things appended onto the end. This seems to be the important difference.

> I think it will be nice to have common standard for event types, but I also feel like it is too restrictive to assume that for an oracle to publish an event it first has to make it standard

Note that this restriction isn't really to do with the event id URI idea per se. You could easily have an event type `enum` which works like `Earth/Paris/2020-08-15/weather?enum(sunny,rainy,cloudy)`. The reason I don't like it is that it encourages the proliferation of bespoke events by oracles. There can now be another oracle that does `Earth/Paris/2020-08-15/weather?enum(clear,rainy,cloudy)` which just happened to choose a different identifier for the same thing. Is this not a concern for you? (would be interested to know other's opinions as well.

I tried to allow oracles to create their own even types without go through some public process. They just have to say in a (to be specified) machine readable way that "event types under this namespace may be from either the standard spec or my custom spec". Then they can define a new  `weather` event type to mean `["sunny", "rainy", "cloudy"]`. If we have a standard way to define enumerated event types then this could all be automatic. This is intentionally more involved than just returning an array in a HTTP response. On top of that it's more efficient.

oracle-uri-proposal.md
Tibo-lg: Yeah it might depend on implementations, but from my understanding of [the rfc](https://tools.ietf.org/html/rfc3986#section-2.2) if you want to use "@" it needs to be percentage encoded.

Tibo-lg:
> There can now be another oracle that does Earth/Paris/2020-08-15/weather?enum(clear,rainy,cloudy) which just happened to choose a different identifier for the same thing. Is this not a concern for you? (would be interested to know other's opinions as well.

Personally I don't really see a big issue there. For me the purpose of the specification is to facilitate the development of client applications that can make use of the information provided by oracles to let users set up contracts. In that respect, there is not really any difference between both, and if the end user is not happy with the result values that they're proposed they can just chose a different oracle. I can also imagine that some oracles might want to sign even outcomes in different languages, in that case, does it mean we should standardize for every language? 

> I tried to allow oracles to create their own even types without go through some public process. They just have to say in a (to be specified) machine readable way that "event types under this namespace may be from either the standard spec or my custom spec". Then they can define a new weather event type to mean ["sunny", "rainy", "cloudy"]. If we have a standard way to define enumerated event types then this could all be automatic. This is intentionally more involved than just returning an array in a HTTP response. On top of that it's more efficient.

I'm not sure I understand what you mean here. I guess the user (or client app) would have to get hold of the event type definition somehow, so wouldn't that end up being the same at the end?

oracle-uri-proposal.md
```diff
@@ -0,0 +1,196 @@
...
+3. Human readable: A user who is sufficiently familiar with event namespace does not need software to understand the event.
+
+## Specification
+
+An event identifier (_event id_) is a relative URI reference (a URI without the scheme and authority) with a query string.
+The path of the URI is interpreted as a description of an event in time and space.
+The query string is interpreted as an _measurement_ and describes a certain fact that will be observed by an oracle on the object described by the path.
+
+If applicable, the path should start with the human organization responsible for the event.
+If it is a price then the name of the exchange where the price is being measured e.g. `binance` or if it's a sporting competition then the name of the competition e.g. `NBA`.
```
nkohen: I'm a little worried about capitalization

oracle-uri-proposal.md
```diff
@@ -0,0 +1,196 @@
...
+### Event Types
+
+The _event type_ occupies the _query string_ part of the URI.
+Event types will be introduced in their own separate RFCs and all event types mentioned here are just an example.
+The event type is a kebab-case identifier which may be followed by _parameters_ enclosed in parenthesis.
+The parameters are a list of comma separated values.
+Each event type defines its parameters and together they determine the outcomes for the event.
+
+The event type will usually define how to interpret the preceding path segments.
+For example, for a `vs` event type may require that the final path segment be in the form `<team1>_<team2>` or a `weather` event type may require the last two segments to be in the form `<location>/<date>`.
```
nkohen, line 60: I think it may be beneficial to add a version to the URI to make things more extensible

oracle-uri-proposal.md
```diff
@@ -0,0 +1,196 @@
...
+### Event Fragments
+
+An event that describes a cardinal quantity like a price will have a large number of discrete outcomes.
+Normally, users will have to arrange a transactions to cover each outcome even if they only care about a small subset of outcomes.
+It is possible to optimize this by splitting up the quantity into some function of multiple sub-outcomes.
+
+For example, a price between \$0 and \$99,999 could be expressed as `10000×e + 1000×d + 100×c + b×10 + a` whereas a..e are in the range 0..10.
+This means that two users who want to make a BTC/USDT trade with boundaries of \$4,000 and \$20,000 and with increments of \$100 would only have to compute contracts for 7 + 5 + 3×5×9 = 147 possible relevant outcomes instead of the usual 100,000 outcomes. 
+
+The oracle will have to declare cryptographically its intention to release signature for each of the decimal decomposed values a..e.
```
nkohen, line 80: Any reason we shouldn't have the oracle specify its base (e.g. 10) to allow for binary or hex in the future should there be events where this makes sense to do?

oracle-uri-proposal.md
LLFourn: I think that proper nouns should be in capitals. So `Binance` and `NBA`. What do you think?

oracle-uri-proposal.md
LLFourn: No the digits event type could indeed take a `base` parameter.

oracle-uri-proposal.md
LLFourn: It also might be annoying on twitter. Perhaps the url can just have a fragment and after the `#` we just serialize metadata like this that doesn't affect the outcomes in some way. So it doesn't appear in the URL but is signed in the announcement message.

LLFourn:
I pushed some changes. The biggest one was that the event ids now start with `/` because I found it easier to get url parsers to work this way. I also removed the source identifier idea in favor of just signing metadata after a `#`.

@Tibo-lg 
> I can also imagine that some oracles might want to sign even outcomes in different languages, in that case, does it mean we should standardize for every language?

[edit] I read this wrong and responded to what I imagined I read. Sorry. I don't think that oracles will want to sign things in different languages for the same event. That would be redundant in a bad way.

Easy localization in general seems like an argument in favor of this proposal. If the oracle is allowed to specify an array of arbitrary strings with unspecified semantics then there's no easy way to create a localized client. If you have a `weather` event type now the client knows all the outcomes and their semantics can translate them for the user.

Although the event types are formally specified in English oracles are free to use paths how they like. For example a _Nippon Professional Baseball_ game could have an event id `/NPB/2020-09-01/東京ヤクルトスワローズ_阪神タイガース?vs` and have an outcome `東京ヤクルトスワロ_win` etc. 

I don't think you have an English version of a NPB game or a Japanese versions of NBA events. Localization is the job of clients.

> I'm not sure I understand what you mean here. I guess the user (or client app) would have to get hold of the event type definition somehow, so wouldn't that end up being the same at the end?

The difference is the clients can get the (immutable) event definition they which resolves the ambiguity of having `[ "sunny", "rainy", "cloudy" ]` in the event data (nonces) the client fetches each time.

Tibo-lg:
> This seems like an argument in favor of this proposal. If the oracle is allowed to specify an array of arbitrary strings with unspecified semantics then there's no easy way to create a localized client. If you have a weather event type now the client knows all the outcomes and their semantics can translate them for the user.

You've got a point, and my example was poorly chosen. But to summarize my opinion, I like (most of) your proposal, and I can definitely see standardization of events as useful and even necessary when it comes to using multiple oracles for a single event.

The point I disagree or maybe don't fully understand, is considering that all the information can be provided with just the URI.

> The difference is the clients can get the (immutable) event definition they which resolves the ambiguity of having [ "sunny", "rainy", "cloudy" ] in the event data (nonces) the client fetches each time.

I don't really get what you are saying here. Where do they get the event definition from? How do they know it is the proper one?

LLFourn:
> The point I disagree or maybe don't fully understand, is considering that all the information can be provided with just the URI.

So far it seems like it will work since it is should not be that hard to describe a set of strings in a single human readable compact form.

To motivate this further I wanted to add "Trump vs Biden" event which is here: http://h00.ooo/Earth/USA/presidential-election/2020/Trump_Biden?vs
To add the event I just made a small cli and gave it:  `/Earth/USA/presidential-election/2020/Trump_Biden?vs`
I can then complete it by giving it: `/Earth/USA/presidential-election/2020/Trump_Biden?vs=Trump_win`.
There is no annoying passing of lists of outcomes via the cli etc. There is also no need to access the database to check if `Trump_win` is a valid outcome for the event.

>I don't really get what you are saying here.

Sorry! that should read.

> The difference is the clients can get the (immutable) event **type** definition they which resolves the ambiguity of having [ "sunny", "rainy", "cloudy" ] in the event data (nonces) the client fetches each time.

"ambiguity" is not what I really wanted there. I think "brittleness" is more what I meant.

This is just an intuition and it may be wrong. My hope is that oracles will never need to do this and we can avoid even specifying anything like this. I have a feeling this is a bridge that can be crossed later. Perhaps allowing oracles to specify their own event types is the solution or maybe just an `enumerate` event type could be added so that an oracles can just list the outcomes in the event id.

Tibo-lg:
> There is no annoying passing of lists of outcomes via the cli etc. There is also no need to access the database to check if Trump_win is a valid outcome for the event.

Correct me if I'm wrong, but the reason you are able to do that is because you hard coded the event type and how to handle it right? If that's the case, I don't think it can scale to a large number of event types, and it means having to upgrade your code every type you want to add a new event, which is very inflexible IMHO.

oracle-uri-proposal.md
```diff
@@ -0,0 +1,179 @@
...
+Each event type defines its parameters and together they determine the outcomes for the event.
+
+The event type will usually define how to interpret the preceding path segments.
+For example, for a `vs` event type may require that the final path segment be in the form `<team1>_<team2>` or a `weather` event type may require the last two segments to be in the form `<location>/<date>`.
+
+We explicitly disallow the oracle to communicate what the outcomes are for a particular event explicitly.
+Outcomes are always determined implicitly from the event type.
+This is done to avoid oracles independently creating events with similar but not precisely the same semantics and structure.
+For example, one oracle might have a weather event that has outcomes `sunny`, `cloudy`, `rainy` while another may have `light-rain` in addition.
+The hope is that this restriction will encourage the public collaboration on new event types and rich client support for them.
```
nkohen, line 66: I think that there should be a built-in (so as to encourage use) versioning of event types in the URI.

oracle-uri-proposal.md
```diff
@@ -0,0 +1,179 @@
...
+An event type may determine that an event is made up of several event fragments.
+When announcing an event with `n` event fragments the oracle creates only one announcement signature which signs all the cryptographic data together:
+
+`<event_id>#<fragment_nonce_1>..<fragment_nonce_n>`
+
+When attesting to a notional outcome (which is composed of the outcomes of each fragment) each fragment message is determined by the message:
+
+`<event_id>.<i>=<fragment_outcome_i>`
+
+## Example Event Types
```
nkohen, line 92: It might be nice to have test vectors for parsing these examples, and in the future when encoding is specified it might also be nice to have serialization and de-serialization test vectors along with signature point computation ones.

nkohen:
So I think on a high-level I agree with Tibo that there is an issue of determining the outcomes should multiple oracles accidentally (or maliciously) decide to spec things slightly (or entirely) differently for the same event. As such, I think it should be important for the oracle to provide an endpoint where you can either query it directly for all possible outcomes (perhaps serialized in some descriptor fashion) or else at the very least you should be able to demand a signature of a hash of the specification they are following in some sense otherwise it may be very hard to generate a fraud proof for the oracle that is reliable.

oracle-uri-proposal.md
LLFourn: I would hope explicit versions will not be necessary. If we screw up specifying `weather`, you can always make `weather2` or perhaps `weather.2` if we want to make visually clear it's a version.

(as side not weather is a bad example since I think it is not the type of thing people will actually use this for)

oracle-uri-proposal.md
LLFourn: Sorry I missed this comment.
In my experience, versions in general make things less extensible because as soon as a client sees a version it does not understand it fails to parse whatever comes next. What kind of extensibility do you have in mind?

oracle-uri-proposal.md
LLFourn: Yes. I think there could be some basic ones in the URI format. Each document that specifies new event types should have have some more e.g. a `vs` parser should extract the team names.

I would like to do this soon but I want to avoid doing this inside this not very well focused document. I think each event type should be in its own document with test etc.

oracle-uri-proposal.md
```diff
@@ -0,0 +1,179 @@
...
+This binary event type asserts that the quantity measured will be above a certain threshold.
+
+### `vs`
+
+This event type is for competitive matches where the one team wins and one team loses or there is a draw.
+Competitions where drawing is impossible can still use the `vs` event type and just use the `draw` outcome to indicate the match being cancelled due to some act of god.
+
+The `vs` event type requires that the path segment preceding it be in the form `<team1>_<team2>` where `<team1>_<team2>` are the competing teams.
+For example, `/NBA/s2019/2020-08-11/PHI_PHX?vs` describes a team named `PHI` is playing against another team named `PHX`.
+Additionally the client can infer the date of the match.
```
LLFourn, line 147: @Tibo-lg made a point last meeting that you can have competitions with multiple entities. e.g. Top scorer in a competition. For this you could do `/FIFA-world-cup/2021/top-scoring-player/Messi_Mbappe_Ronaldo?vs(3)`.

LLFourn:
@nkohen 
> So I think on a high-level I agree with Tibo that there is an issue of determining the outcomes should multiple oracles accidentally (or maliciously) decide to spec things slightly (or entirely) differently for the same event. As such, I think it should be important for the oracle to provide an endpoint where you can either query it directly for all possible outcomes (perhaps serialized in some descriptor fashion) 

I don't think making another descriptor is a solution to misalignment between the oracle and client. If the client is misaligned to the oracle then they can be misaligned in the implementation of this second descriptor as well.

> or else at the very least you should be able to demand a signature of a hash of the specification they are following in some sense otherwise it may be very hard to generate a fraud proof for the oracle that is reliable.

Signing a digest of the specification they are following for the event type is a good idea. However, signing an entire spec is a bit much since if you change a typo then you're in trouble. Perhaps each event type can have a formal specification attached and that can be signed. I really like that idea and I think how sensitive the protocol is to correctness in this respect warrants it. It remains to choose a language.

oracle-uri-proposal.md
Christewart: Related to #88

LLFourn:
I'll close this now.

The things that were concluded in the last meeting were:

1. Most people want the oracle to explicitly list all the outcomes when there are a relatively small number of them.
2. I can still follow this idea informally. All that I need to do is to transform the URI into the "full" descriptor in the announcement signature.

Issue #76, opened 2020-09-08
Specify Signautre TLVs


nkohen:
CET signatures and funding signatures need TLVs

Issue #72, opened 2020-09-08
Funding Input Communication Specification (TLV)


nkohen:
We will need a minimal way to communicate funding inputs along with fee bound information.

nkohen:
Closed by #81

Issue #62, opened 2020-08-17
Contract Info TLV Specification


nkohen:
We need standard ways of serializing Contract Info in our P2P messages (see #59).

I foresee that there will be many different contract info types including

- Map from outcome hashes to values
- Ranges of outcomes to values
- Map from multiple outcome hashes and/or ranges to values

and likely more in the future. As such, I propose we use different TLV types for these different cases.

To allow test vectors to be written and development to begin without being blocked on this, I also propose type 0 contract information to be a the number of outcomes, followed by `(32 byte hash || 8 byte value)` pairs for each outcome.

benthecarman:
To do #40 we will eventually want to be able to define or reference oracle infos with each outcome

nkohen:
https://github.com/discreetlogcontracts/dlcspecs/pull/59#discussion_r471774033

Issue #61, opened 2020-08-17
Oracle Info TLV Specification


nkohen:
We need standard ways of serializing Oracle Info in our P2P messages (see #59).

I foresee that there will be many different oracle info types including

- Oracle with single R value
- Oracle with multiple R values
- Multiple oracles (n-of-n)
- Multiple oracles (m-of-n)

and likely more in the future. As such, I propose we use different TLV types for these different cases.

To allow test vectors to be written and development to begin without being blocked on this, I also propose type `0` oracle information to be a single oracle with a single R value serialized as the concatenation of the oracle signing key and the R value (32 bytes each for a total of 64 bytes + type and length prefix).

LLFourn:
I am not following the design of the p2p layer closely but I have some queries on this.

The oracle signing key is to identify the oracle right (It is not actually used by the other party)? Why would you send the R values; Shouldn't they just fetch them themselves?
If you are saying parties may send the oracle announcement message (signed by the oracle themselves) with its R values(s) then that's a nice idea but then I think the serialization should just be the announcement message itself.

nkohen:
The current `v0` type is to help get development off the ground while oracle discussions are still in flux, it allows for the construction of DLCs for testing and manual play, in the future I assume better types will be introduced that make more sense and are more robust

PR #81, opened 2020-09-09
Proposed v0 DLC TLV messages and Deterministic Fee Computation

nkohen:
Begins work on implementing #47, #61, #62, #72, #76

Implements #12

EDIT: Used to implement static test vectors but these have been moved to another PR


nkohen:
I should note that the type numbers were picked randomly and are likely subject to change

benthecarman:
> I should note that the type numbers were picked randomly and are likely subject to change

Maybe should put that in the doc

Messaging.md
```diff
@@ -44,15 +55,102 @@ The following convenience types are also defined:
...
+
+#### Version 0 `cet_signatures`
+
+1. type: 42774 (`cet_signatures_v0`)
+2. data:
+   * [`adaptor_signature`:`signature_1`]
+   * ...
+   * [`adaptor_signature`:`signature_n`]
+
+This type should be used with [`contract_info_v0`](#version-0-contract_info) where each indexed signature in the data corresponds to the outcome of the same index. And as in [`contract_info_v0`](#version-0-contract_info), the number of signatures is omitted as it can be derived from the length field of the TLV.
```
benthecarman: >. And as in contract_info_v0

Shouldn't start a sentence with `And` maybe change to `Following contract_info_v0, the ...`

Messaging.md
```diff
@@ -44,15 +55,102 @@ The following convenience types are also defined:
 * `contract_id`: a 32-byte contract_id (see [Protocol Specification](Protocol.md))
 * `sha256`: a 32-byte SHA2-256 hash
 * `signature`: a 64-byte bitcoin Elliptic Curve signature
+* `adaptor_signature`: a 162-byte ECDSA adaptor signature
+* `x_point`: a 32-byte x-only public key (with implicit y-coordinate being even)
```
benthecarman: Maybe link to BIP 340

Messaging.md
```diff
@@ -44,15 +55,102 @@ The following convenience types are also defined:
 * `contract_id`: a 32-byte contract_id (see [Protocol Specification](Protocol.md))
 * `sha256`: a 32-byte SHA2-256 hash
 * `signature`: a 64-byte bitcoin Elliptic Curve signature
+* `adaptor_signature`: a 162-byte ECDSA adaptor signature
```
benthecarman: Should add a todo link to the future doc defining this

make link to #50

benthecarman: Also maybe name it `ecdsa_adaptor_signature` so it is more clear when we have schnorr as well.

Messaging.md
```diff
@@ -44,15 +55,102 @@ The following convenience types are also defined:
...
-* `oracle_info`: ???
-* `funding_input`: ???
-* `cet_signatures`: ???
-* `funding_signatures`: ???
+
+## DLC Specific Types
+
+The following DLC-specific types are used throughout the specification:
+
+### The contract_info Message
```
benthecarman: nit:

> ### The contract_info Message

### The `contract_info` Message

benthecarman (changes requested):

benthecarman (approved):

Messaging.md
```diff
@@ -44,15 +55,102 @@ The following convenience types are also defined:
...
-* `cet_signatures`: ???
-* `funding_signatures`: ???
+
+## DLC Specific Types
+
+The following DLC-specific types are used throughout the specification. All type numbers are placeholders subject to change both here and in [Protocol.md](Protocol.md).
+
+### The `contract_info` Message
+
+This message contains information about a contracts outcomes and their corresponding payouts. To save space, only one side's POV is included in this message as the other can be derived using `remote_payout = total_collateral - local_payout`.
```
Tibo-lg: Nit: no need for plural

Otherwise, we can go with that for now, but I think in the future it might be good to have a way to express the payouts as a function (I think that's what @NicolasDorier was alluding to in #70). For example, it seems wasteful to transmit 1000 values to say `9000, 9001...10000`. You can easily reduce that using a range with `start=9000, stop=10000, step=1`. I guess a good solution would be to support both enumeration (if the payouts cannot be expressed as a range) and range.
For the outcomes, they will also very often be expressible as a range, so it would seem a better idea to have them "in clear". It might also be related to how the oracle represents outcomes.

Messaging.md
```diff
@@ -44,15 +55,102 @@ The following convenience types are also defined:
...
+2. data:
+   * [`sha256`:`outcome_1`]
+   * [`u64`:`outcome_1_local_payout`]
+   * ...
+   * [`sha256`:`outcome_n`]
+   * [`u64`:`outcome_n_local_payout`]
+
+This type of contract info is a simple enumeration where the value `n` is omitted from being explicitly included as it can be derived from the length field of the TLV.
+
+### The `oracle_info` Message
```
Tibo-lg: Still related to oracle specifications, but I am unsure that we need to transmit more than an event id (which would identify both an oracle and an event). As I think I mentioned during our last meeting, if you are unable to get this info from the oracle (directly or indirectly), you should probably not accept the contract. (I'm ok to use that for now though until we have decided the oracle specs).

Messaging.md
```diff
@@ -44,15 +55,102 @@ The following convenience types are also defined:
...
+
+This message contains information about a specific input to be used in a funding transaction, as well as its corresponding on-chain UTXO.
+
+#### Temporary `funding_input`
+
+1. type: 42772 (`funding_input_temp`)
+2. data:
+   * [`sha256`:`prev_txid`]
+   * [`u32`:`vout`]
+   * [`u64`:`value`]
```
Tibo-lg: I mentioned this previously but it would be more efficient to use aggregated values of all inputs, unless there is a reason not to do so?

Messaging.md
```diff
@@ -44,15 +55,102 @@ The following convenience types are also defined:
...
+   * [`spk`:`spk`]
```
Tibo-lg: What's the reason for transmitting the spk? I think we can only work with four types: P2WPKH, P2WSH, P2SH-P2WPKH and P2SH-P2WSH, wouldn't it make more sense to define them in an enum and pass the value here?

Messaging.md
```diff
@@ -44,15 +55,102 @@ The following convenience types are also defined:
...
+
+This type is not actually being proposed and will be replaced with a more suitable type in the near future. This is however needed to get a start on static test vectors and compatibility between implementations so it is currently included here for use in test vectors (and will later be replaced both here and in test vectors).
+
+`prev_txid` is the little-endian TXID of the transaction on which this UTXO resides at `vout`.
+`value` is the output amount of the UTXO and `spk` is the script public key that must be satisfied by this input.
+
+Note that `prev_txid || vout` form a serialized Transaction Outpoint and ` value || spk` form a serialized Transaction Output
+so that the data on this TLV can be parsed by most libraries conveniently as `outpoint || output`.
+
+### The `cet_signatures` Message
```
Tibo-lg: nit: Should we call that `cet_adaptor_signatures`? It would be more explicit IMHO.

Messaging.md
```diff
@@ -44,15 +55,102 @@ The following convenience types are also defined:
...
+
+This type should be used with [`contract_info_v0`](#version-0-contract_info) where each indexed signature in the data corresponds to the outcome of the same index. As in [`contract_info_v0`](#version-0-contract_info), the number of signatures is omitted as it can be derived from the length field of the TLV.
+
+### The `funding_signatures` Message
+
+This message contains signatures of the funding transaction and any necessary information linking the signatures to their inputs.
+
+#### Version 0 `funding_signatures`
+
+1. type: 42776 (`funding_signatures_v0`)
```
Tibo-lg, line 152: Couldn't we do some lexicographical ordering of tx inputs and give the signatures in that order? (I'm ok to keep it as is for now, but could be more efficient).

Messaging.md
```diff
@@ -44,15 +55,102 @@ The following convenience types are also defined:
 * `contract_id`: a 32-byte contract_id (see [Protocol Specification](Protocol.md))
 * `sha256`: a 32-byte SHA2-256 hash
 * `signature`: a 64-byte bitcoin Elliptic Curve signature
+* `ecdsa_adaptor_signature`: a 162-byte ECDSA adaptor signature (TODO: link to doc once [#50](https://github.com/discreetlogcontracts/dlcspecs/issues/50) is done)
```
Tibo-lg: IIUC there is a proof and a signature. Is that the concatenation of the two? IMHO it would be clearer to separate and indicate how they are concatenated (since the secp implementation outputs them separately and I guess this is somehow the reference implementation).

Messaging.md
```diff
@@ -44,15 +55,102 @@ The following convenience types are also defined:
...
+* `x_point`: a 32-byte x-only public key with implicit y-coordinate being even as in [BIP 340](https://github.com/bitcoin/bips/blob/master/bip-0340.mediawiki#design)
```
Tibo-lg, line 66: More for my understanding, but does this need to be wit implicit y-coordinate? I wanted to avoid exposing new types in my implementation so I just ended up using regular public key (33-bytes), so wondering if there could be an issue.

Tibo-lg (commented):
Looks good! I put some comments about things that could likely be optimized, but we can iterate over this in the future.

Messaging.md
benthecarman: Come taproot we'd have to change it back as the signature commits to amounts

https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki#L108

Messaging.md
benthecarman: AFIAK you should be able to just drop the first byte of the pubkey.

Messaging.md
Tibo-lg: Good point!

Messaging.md
Tibo-lg: Yeah that's what I do currently :) I guess it must be fine

Messaging.md
Tibo-lg: Mmm actually that's only true for the PKH types, I guess for SH types, we might want the spk as well as the witness script to compute fees (and check that the given witness script matches with the spk)

Messaging.md
nkohen: I totally agree that in the future there will be other versions, and perhaps this isn't made clear enough but I plan for multiple types of each of these TLVs to exist _at the same time_ in the future where users can use the type that best fits. I started with enumeration because that is the easiest kind to write simple test vectors for but in the future I do not expect this to be used for anything other than small enumerations like the election bet

Messaging.md
nkohen: yeah so I plan on doing a complete re-do of this one in the future as the `temp` suggests ... in the future what is really needed is much closer to what [dual funding proposal](https://github.com/niftynei/lightning-rfc/pull/1/files#diff-3369c5aa1774fef2ff1e246979f223eaR169) has where you give `prevtx` (to verify amount), `redeemscript` and `redeemscript_len`, as well as `max_witness_len` which is later verified during signing (and the process fails if either party tries to use a witness larger than `max_witness_len`, this way of doing things also simplifies fee computation as you can just use these numbers rather than computing from an enum with some extra data or something like this

Messaging.md
nkohen: I think if we want to match dual funding we will need to do something similar to their `serial_id`s to allow for more general ordering constructions, so for now I will leave any sorting complexity out

Messaging.md
nkohen: yeah ... the translation between x-only and normal is adding or subtracting a `02` byte on the front, but I figured it would be nice not to include the `02` on Schnorr keys ... open to changing this though

Messaging.md
Tibo-lg: Cool, I must confess I had missed the temp part :)

Messaging.md
Tibo-lg: I feel it depends if x-only pubkeys are going to be something common in the (near) future or if will still be mainly internal to libsecp (or other crypto libraries).

nkohen: I think either way if someone wants to support DLC spec stuff they are required to implement Schnorr interfaces which all use x-only so I don't think its a weird ask

Tibo-lg: I personally implemented the Schnorr module in rust using just regular public keys, because I didn't want to add extra types to the public interface and to keep things simple(r). If I add x-only pubkeys to the public interface, I also need to start adding util functions to get an x-only pubkey from private key, or from a regular public key... It's not infeasible, but it didn't seem super necessary either. And since they provide a `xonly_pubkey_from_pubkey` function in libsecp I thought it was maybe not meant to be exposed externally.

Messaging.md
```diff
@@ -44,15 +55,102 @@ The following convenience types are also defined:
...
+
+This message contains signatures of the funding transaction and any necessary information linking the signatures to their inputs.
+
+#### Version 0 `funding_signatures`
+
+1. type: 42776 (`funding_signatures_v0`)
+2. data:
+   * [`sha256`:`prev_txid_1`]
+   * [`u32`:`vout_1`]
+   * [`signatures`:`signature_1`]
```
Tibo-lg: nit: I don't think it needs to be plural?

Messaging.md
nkohen: tbh I don't know what the internal/external plan is for x-only but I know that on-chain there will be x-only keys and nonces once we have taproot so I think the types will be pretty prevalent in libraries. Furthermore I'd argue that since this is only the P2P part of the protocol you can just use whatever type you want internally and only do x-only conversions in your serialization/deserialization logic as it is pretty trivial.

Tibo-lg: Ok let's go with that for now then. I was also trying to understand if I should switch to x-only in my libraries but I guess I can figure that out later.

test/dlc_test.json
```diff
@@ -0,0 +1,329 @@
...
+    "params" : {
+      "oracleInfo" : {
+        "publicKey" : "a78bf5ed2841a93d4dcc1b719ebbcc90c7648746e5bd55fe7806494a93167c39",
+        "nonce" : "d2d6cb7049ef4edc2d98aa3e235e8d542573cf0c9a735516b9936aba52ae693e"
+      },
+      "contractInfo" : {
+        "8873e27fd5804a500792d3b197ec83e3d13311e161c431b223657b64ec78e355" : 0,
+        "9fc2710e301f8700a8a4c6933366c52045f23a59c92eaf17b0c3d6413861d068" : 200000000
+      },
+      "contractMaturityBound" : 100,
```
Tibo-lg: What is this used for?

test/dlc_test.json
```diff
@@ -0,0 +1,329 @@
...
+      "contractTimeout" : 200,
```
Tibo-lg: Does that correspond to the refund lock time?

test/dlc_test.json
nkohen: This is the locktime on the CET

test/dlc_test.json
nkohen: yes

test/dlc_test.json
Tibo-lg: nit but maybe better to just call it cetLockTime then (same for next one refundLockTime), it's more explicit IMHO

test/dlc_test.json
```diff
@@ -0,0 +1,329 @@
+[ {
+  "testInputs" : {
+    "params" : {
+      "oracleInfo" : {
+        "publicKey" : "a78bf5ed2841a93d4dcc1b719ebbcc90c7648746e5bd55fe7806494a93167c39",
+        "nonce" : "d2d6cb7049ef4edc2d98aa3e235e8d542573cf0c9a735516b9936aba52ae693e"
+      },
+      "contractInfo" : {
+        "8873e27fd5804a500792d3b197ec83e3d13311e161c431b223657b64ec78e355" : 0,
```
Tibo-lg: This is not really practical to parse. I can see in your code you use a Map, but usually maps don't guarantee ordering especially across implementations so I think it would be better to use an array here.

test/dlc_test.json
```diff
@@ -0,0 +1,329 @@
...
+        "9fc2710e301f8700a8a4c6933366c52045f23a59c92eaf17b0c3d6413861d068" : 200000000
+      },
+      "contractMaturityBound" : 100,
+      "contractTimeout" : 200,
+      "feeRate" : 5
+    },
+    "offerParams" : {
+      "collateral" : 100000000,
+      "fundingInputTxs" : [ {
```
Tibo-lg: Shouldn't you include the input value as well?

Tibo-lg: Just realized that `tx` is a transaction and not a txid (which I had assumed). Is there a specific reason not to just pass a txid and value since that is what is needed to build the messages and txs?

Tibo-lg: And rust-bitcoin seem not to be able to parse this transaction, AFAICT due to [how they resolve ambiguity of zero input transactions](https://github.com/rust-bitcoin/rust-bitcoin/blob/master/src/blockdata/transaction.rs#L243). So I would really recommend using txid + value to make things simpler.

Messaging.md
```diff
@@ -44,15 +55,109 @@ The following convenience types are also defined:
...
-* `oracle_info`: ???
-* `funding_input`: ???
-* `cet_signatures`: ???
-* `funding_signatures`: ???
+
+## DLC Specific Types
+
+The following DLC-specific types are used throughout the specification. All type numbers are placeholders subject to change both here and in [Protocol.md](Protocol.md).
+
+### The `contract_info` Message
```
Tibo-lg: nit: AFAIU these are not messages per say, as in they don't get exchanged directly but rather get included in messages, so I think it would be better to refer to them as `Type` rather than `Message`.

test/dlc_test.json
```diff
@@ -0,0 +1,329 @@
...
+        "idx" : 0,
+        "inputKey" : "e2d7b500267e5d1656f2f2fc0fd0580a48f525c0aa537d8b356c1793e9626131"
+      } ],
+      "changeAddress" : "bcrt1q4kw7nkyq42u6nnnhey3ylagpjscq7kwr4f2ufp",
+      "fundingPrivKey" : "5391f9340bdd7c194023d31176b6aab334593a9afe5d3356987d74b2833f2e59",
+      "payoutAddress" : "bcrt1qx9f936etaa43qxnu0ss7zppca296xglzt9z7wv"
+    }
+  },
+  "offer" : "fda71afd015d000f9188f13cb7b2c71f2a335e3a4fc328bf5beb436012afca590b1a11466e2206fda710508873e27fd5804a500792d3b197ec83e3d13311e161c431b223657b64ec78e35500000000000000009fc2710e301f8700a8a4c6933366c52045f23a59c92eaf17b0c3d6413861d068000000000bebc200fda71240a78bf5ed2841a93d4dcc1b719ebbcc90c7648746e5bd55fe7806494a93167c39d2d6cb7049ef4edc2d98aa3e235e8d542573cf0c9a735516b9936aba52ae693e03f12e6c13142cc488b4f3d31b45be115d9637f6ccf4b4e3f1223a1a71052889d11600143635941af39ada33de91fdc620242603d57ba89f0000000005f5e1000001fda71437002902000000000100c2eb0b000000001600145845070bc9e99b0db44039e0567e32408a1ff6360000000000000000ffffffff006c0000160014e79b5c55970d9276051d6a71f93bcaacb094be22000000000000000000000064000000c8",
```
Tibo-lg: I can't figure out why your message starts with `fd`. I think it's supposed to start with the message type which is the following `a71a`?

nkohen: Check the definition of BigSizeUInt here: https://github.com/lightningnetwork/lightning-rfc/blob/master/01-messaging.md#appendix-a-bigsize-test-vectors

test/dlc_test.json
nkohen: Sadly this is not possible: https://github.com/niftynei/lightning-rfc/pull/1#discussion_r471762248 (at least until we have Taproot)

nkohen: But I can add a dummy input if that would make your life easier

nkohen: added

test/dlc_test.json
nkohen: I'm going to make a separate very small PR for this once this gets in (to avoid merge conflicts as this will change all the files)

benthecarman (approved):

test/dlc_test.json
Tibo-lg: Thanks (haven't checked yet but hopefully it should work :) )

NicolasDorier:
trying to implement this on my side

Messaging.md
```diff
@@ -44,20 +55,114 @@ The following convenience types are also defined:
...
+2. data:
+   * [`u16`:`prevtx_len`]
+   * [`prevtx_len*byte`:`prevtx`]
+   * [`u32`:`prevtx_vout`]
+   * [`u32`:`sequence`]
+   * [`u16`:`max_witness_len`]
+   * [`u16`:`redeemscript_len`]
+   * [`redeemscript_len*byte`:`script`]
+
+`prevtx_tx` is the serialized transaction whose `prevtx_vout` output is being spent.
```
NicolasDorier, line 121: This is a deal breaker for me. 

In BTCPay Server, we don't have the prevtx, because if the user restored his wallet, it is through scantxoutsetinfo and we work on pruned node, so no way to get back this data.

NicolasDorier:
The test vector should also specify the pre images of the outcome's hash, and with some of them using accent to test string normalization.

Messaging.md
devrandom: I think this requirement can be dropped, because the signatures already cover the input value (it's a segwit feature).

NicolasDorier: Well it was not needed in the non-TLV version of the spec, why has it been added?

Messaging.md
```diff
@@ -44,20 +55,114 @@ The following convenience types are also defined:
...
+
+The following DLC-specific types are used throughout the specification. All type numbers are placeholders subject to change both here and in [Protocol.md](Protocol.md).
+
+### The `contract_info` Type
+
+This type contains information about a contracts outcomes and their corresponding payouts. To save space, only one side's POV is included in this message as the other can be derived using `remote_payout = total_collateral - local_payout`.
+
+#### Version 0 `contract_info`
+
+1. type: 42768 (`contract_info_v0`)
```
ariard, line 83: Even if it's temporary can you assign them in the LN's messages waiting rooms : https://github.com/lightningnetwork/lightning-rfc/issues/716 ?

Messaging.md
```diff
@@ -44,20 +55,114 @@ The following convenience types are also defined:
...
+This type contains information about the oracle(s) to be used in executing a DLC, and possibly the outcomes possible if these are not specified in the corresponding `contract_info`.
+
+#### Version 0 `oracle_info`
+
+1. type: 42770 (`oracle_info_v0`)
+2. data:
+   * [`x_point`:`oracle_public_key`]
+   * [`x_point`:`oracle_nonce`]
+
+This type of oracle info is for single-oracle, single signature (and hence single nonce) events.
```
ariard, line 104: A side-note, we should make mandatory in the Oracle spec-side to verify public key ownership to avoid someone impersonating the oracle and thus blocking the state of your contract.

Though I don't know the state of oracle discovery, if we rely on common public key infrastructure.

cc @LLFourn

Messaging.md
ariard: As a receiver you have to verify that announced inputs aren't malleable ones, how do you combine this with pruning mode ?

Messaging.md
```diff
@@ -44,20 +55,114 @@ The following convenience types are also defined:
...
+   * [`u16`:`prevtx_len`]
+   * [`prevtx_len*byte`:`prevtx`]
+   * [`u32`:`prevtx_vout`]
+   * [`u32`:`sequence`]
+   * [`u16`:`max_witness_len`]
+   * [`u16`:`redeemscript_len`]
+   * [`redeemscript_len*byte`:`script`]
+
+`prevtx_tx` is the serialized transaction whose `prevtx_vout` output is being spent.
+The transaction is used to validate this output's value and to validate that it is a SegWit output.
```
ariard: Add a `spent output` prefix for clarity.

Messaging.md
```diff
@@ -44,20 +55,114 @@ The following convenience types are also defined:
...
+   * ...
+   * [`u16`:`num_witness_elems_num_witnesses`]
+   * [`num_witness_elems_num_witnesses*witness_element`:`witness_elements_num_witnesses`]
+3. subtype: `witness_element`
+4. data:
+   * [`u16`:`len`]
+   * [`len*byte`:`witness`]
+
+`witness` is the data for a witness element in a witness stack. An empty `witness_stack` is an error,
+as every input must be Segwit. Witness elements should *not* include their length as part of the witness data.
```
ariard, line 166: @Christewart Can we tag as a V0.2 to reject malleable witness scripts ? I don't think Miniscript compilers have the interface yet but that something to consider in the future.

Messaging.md
nkohen: @devrandom it covers your own input value but not the others https://github.com/bitcoin/bips/blob/master/bip-0143.mediawiki#specification

If you don't include the full prev tx they can give you a valid txid pointing to a non-segwit address and convince you it is segwit (because you're trusting them by not requiring the full prevtx) once you sign the funding tx, they can include a script signature you weren't expecting, your signatures are still valid but the txid changes and all CETs and refund tx become invalid.

nkohen: It wasn't there before because I was lazy, they should be there, they also should be present in PSBTs as well as that they are in the LN dual funding proposal

benthecarman: #TaprootFixesThis because then the signatures will commit to the other inputs' spks

ariard (commented):
Sounds good to me.

Messaging.md
nkohen: Re ariard I agree, in bitcoin-s we had to add a transaction table to our db where we keep track of the prevtxs of our current UTXO set

Messaging.md
nkohen: Hmmm, I'm not sure this is a feasible constraint seeing as one party never sees the other's witnesses before they see them on-chain. Is there any reason witness malleability is an issue in this context as it doesn't invalidate future txs?

Messaging.md
NicolasDorier: @nkohen even with this, the previous transactions should not be required. Let me explain why it is problematic, and why it is not needed.

First the problem:
Users restoring wallet on BTCPay Server, are using `scantxoutset` RPC command. This allow them to restore the balance of a wallet, without having to redownload blocks.

The downside of it, is that our users don't have access to their previous transaction. And even if they do have access, this is really not user friendly as, when they create the [setup PSBT](https://github.com/dgarage/NDLC/blob/master/docs/Concepts.md#alice-wants-to-create-a-dlc-how-to-make-an-offer), they need to specify that they need to include the previous TX in the PSBT. (A normal user have no idea what it means, and would be confused if it does not work when he needs it)
On top of it, such option just does not exist for bitcoin core wallet for example.

So basically, as far as I am concerned, requiring the previous tx as part of the protocol either means:
1. I stop working on DLC until taproot fixes this
2. I use a variant of this protocol for NDLC
3. This protocol revert this change and does not requires the previous transaction.

Now, let me explain why `3.` does not represent a security risk.

In NDLC, the signature of the funding transaction is done at [this step](https://github.com/dgarage/NDLC/blob/master/docs/Concepts.md#alice-starts-the-contract), and it can be done with any wallet supporting PSBT.

Given the attack you explained, NDLC will not include the witness_utxo or non_witness_utxo in the funding PSBT that it asks the user to sign.

All wallet signing PSBTs, in such condition, would refuse to sign without updating the PSBT with data from their own UTXO Set, meaning that the witness_utxo is not controlled by the attacker.
On top of it, I will require, at NDLC level, that when the user submit the partially signed PSBT of the funding transaction, the witness_utxo is properly filled. This would detect any wallet blindly signing and thus being a danger for the user.
If the wallet blindly sign, it would be possible for the user to use another tool/block explorer to update the PSBT separately.

Such PSBT update can be trivially done by bitcoin core using only information from the UTXO Set (and thus, by BTCPay, which has access to it).
I believe it can also be done by Electrum. (Yes, trusting the electrum server)

nkohen: my attack has `witness_utxo` set but that field only has an output and that output isn't committed to by your signature (so the counter-party can use something else)

NicolasDorier: > my attack has witness_utxo set but that field only has an output and that output isn't committed to by your signature (so the counter-party can use something else)

The counter-party can't use another output, as the outpoint is committed to. (by `hashPrevouts`)

However, he can trick me by saying his input is segwit.
But my point is that somebody with the UTXO Set can protect against such attack without the need of the previous transaction.

However, for wallets without UTXO set, they can't verify it (Bitcoin-S and Wasabi wallet for example)

AFAIK, there is no UX available in bitcoin core to force the inclusion of previous transaction in PSBTs either.
So it means, you can't make DLC without having to develop a wallet from scratch for it. (and requiring full block scan rather than UTXO scan for restore)

I think this is a major hurdle, so on my side I will probably wait until taproot fix this before continuing my work on DLC, or use another protocol until then. I don't think there is a solution to solve this problem.

nkohen: So I agree with most of this but it isn't actually a problem if you don't have the UTXO set if they share the prevtx because worst case the prevtx is not actually real or the txo is already spent, in either case the funding tx will be invalid. This is ok. What is not okay is if you don't use prevtx, don't have the utxo set and they lie about there being a real txid, vout corresponding to an output (and they fake the output telling you it is segwit when it isn't) in this case they can lock up your funds.

Here is my proposal: The only safe way for light clients to do this is to use prevtx, otherwise they are trusting their coutner-party not to cheat them in this specific way which can lock up their funds. If a party has access to the UTXO set however, they really don't need or want prev-tx, so in the future I want there to be TLV versions of offer and accept that both have and don't have prev-tx and full nodes will receive non-prevtx messages while light clients will need prevtx. Because this is my plan, you will still be able to use the non-prevtx version (but it will be trusting) until you either have the infrastructure to support prevtx or taproot comes around after which this won't be necessary anymore. Thoughts?

nkohen: Also, the kind of good news is if your tool is somewhat manual, users can use external sources to validate prev_txid is in the utxo set so that it isn't counter-party trust

nkohen: oh and I think I misunderstood, if I now understand correctly, with this new versioned approach you should be able to be trustless if your counter-party accepts txid+vout instead of prevtx (if they are a full node for example) as the issue for you is getting your own prevtx, not validating the counter party's if you are given a prevtx right?

NicolasDorier: @nkohen so I searched a bit:

* Bitcoin core is including previous transaction when creating a new PSBT for the wallet since 0.20.1. (latest version)
* So I swtiched BTCPay Server to mirror the same behavior by default. Sadly, this won't work for restored wallet.

> as the issue for you is getting your own prevtx, not validating the counter party's if you are given a prevtx right?

Correct.

Stepping back a bit, I think the best reason for allowing witness_utxo in the protocol, is not about making it easier for wallets with the full utxo set. But to prepare taproot. Once we have taproot, the non_witness_utxo won't be needed in any case. So I think it makes sense to allow it in the protocol for this reason.
On my side, my implementation of this protocol I would allow witness_utxo even for segwit v0 as we have the utxo set, but still try to add non_witness_utxo, if possible, to offer/accept.

nkohen: Great news :) Lmk if there's anything I can do to help. And also to clarify, you're cool with moving forward with prevtx and adding a non-prevtx version in the near future (potentially the subsequent PR to this one)?

NicolasDorier: yes, seems good.

Messaging.md
nkohen: https://github.com/discreetlogcontracts/dlcspecs/issues/97

Messaging.md
nkohen: I think I'll wait a while longer on this until we are sure we will actually be using LN message format long-term (e.g. does #96 work as an approach) before adding to here but I will do so if it becomes clear we won't be using an incompatible variant.

Messaging.md
nkohen: https://github.com/discreetlogcontracts/dlcspecs/issues/98

Messaging.md
ariard: Yes witness malleability is a pinning vector and as such you can prevent confirmation of the funding transaction and escape the contract in case of non-favorable outcome. 

In theory you can parse a `witnessScript` and decide if a solving witness is malleable. But as pointed IIRC  Miniscript isn't supporting this.

ariard:
SGTM 3b88814

benthecarman (approved):

Messaging.md
LLFourn: > A side-note, we should make mandatory in the Oracle spec-side to verify public key ownership to avoid someone impersonating the oracle and thus blocking the state of your contract.

I didn't understand what this meant. How can you impersonate the oracle?
Note I made a comment here: https://github.com/discreetlogcontracts/dlcspecs/issues/61 about how I don't think it makes sense to send `oracle_nonce`.  If you do you the oracle must sign it along with the event data it is associated with. Perhaps this is what you are getting at.

NicolasDorier:
~@nkohen what if the funder is wrapped P2SH? He should gives the P2SH pre-image as part of the fundingInputs to prove it is segwit.~

Actually it is there, just not in the test vector.

Messaging.md
```diff
@@ -44,20 +61,114 @@ The following convenience types are also defined:
...
+
+#### Version 0  `funding_input`
+
+1. type: 42772 (`funding_input_v0`)
+2. data:
+   * [`u16`:`prevtx_len`]
+   * [`prevtx_len*byte`:`prevtx`]
+   * [`u32`:`prevtx_vout`]
+   * [`u32`:`sequence`]
+   * [`u16`:`max_witness_len`]
```
NicolasDorier, line 118: missing the `script` field.

NicolasDorier:
~`spk` is not defined, is it BigInt + bytes?~

NicolasDorier:
This does not seem to be valid TLV

```
A tlv_record represents a single field, encoded in the form:

    [bigsize: type]
    [bigsize: length]
    [length: value]
```

So the offer should start by `fda71a`, not by `a71a`

Tibo-lg:
@NicolasDorier it is super confusing but the outer messages are not TLVs, check this paragraph: https://github.com/nkohen/dlcspecs-1/blob/happy-path-test-vectors/Messaging.md#message-format

This is to ensure compatibility with LN.

test/dlc_test.json
```diff
@@ -0,0 +1,487 @@
+[ {
+  "testInputs" : {
+    "params" : {
+      "oracleInfo" : {
+        "publicKey" : "6b745e29cd1e8ec0cd3f3c817679689e2c55bf511a5032a46ac42123d68236b2",
+        "nonce" : "f336279279a95151a571ab441d9c5735aaa4c3080439a74f144b8a3005409dec"
+      },
+      "contractInfo" : [ {
```
Tibo-lg: Would it be possible to have this as an array of objects with proper fields? Like:
```json
[ {
   "outcome": "7014aeda818f943c1a45e3d90db1a73b7bd3e665a96d784d82afd84ee824cea7",
   "local_payout": 0
} ]
```

It is really a pain to parse in the current form.

NicolasDorier:
The test vector does not pass the spec.
The spec says:
```
[oracle_info:oracle_info]
[point:funding_pubkey]
```
But the test vector seems to have the payout scriptPubKey instead after oracle_info.

nkohen:
> The test vector does not pass the spec.
> The spec says:
> 
> ```
> [oracle_info:oracle_info]
> [point:funding_pubkey]
> ```
> 
> But the test vector seems to have the payout scriptPubKey instead after oracle_info.

I don't really know what you're referring to? That spec is for the `dlc_offer` message and I think you might be talking about the `testInputs` field which is very much not specified because it is only a test vector thing (where we have priv keys instead)?

Messaging.md
nkohen: not sure what you are referring to? If you're talking about p2sh that is the redeem script below and if not then you definitely must not include a witness script when giving your funding input as funding signatures must be delayed until after CET signatures are received.

NicolasDorier:
@nkohen I am talking in the TLV offer message of the test vector.

The `[ { .offer } ]` which is specified in `dlc_offer`. The byte array I am parsing does not seem to use the same order of field than what is specified here.

nkohen:
@NicolasDorier Here is my parsing of the first offer in the file:
```
ByteVector(2 bytes, 0xa71a)
ByteVector(1 bytes, 0x00)
ByteVector(32 bytes, 0x0f9188f13cb7b2c71f2a335e3a4fc328bf5beb436012afca590b1a11466e2206)
ByteVector(84 bytes, 0xfda7105064058b3477561188b9c89a0dbb8a843b797c4f4beb2a21dae4d3a1b54023b89e00000000000000004129cf1ed8463c29956c5edfe9e44de9f8da771f9f9b76f91897791a88bb2826000000000bebc200)
ByteVector(68 bytes, 0xfda71240ce266a66009b7e7b9f39c90ec77b6cc0cd7f58e087ed55c0b019545c44d38e2162fd055478041c159eb0b3b5ce3b9fbe28470522c6ac446873ed27f1c8dcf745)
ByteVector(33 bytes, 0x03efd58eb7b7c79fef1ff1edde8fb0b49c2358020097c016e05c17a129af01940f)
ByteVector(23 bytes, 0x160014c0bb4e0bfb0773843a7b4ff18a4a24e2312c3986)
ByteVector(8 bytes, 0x0000000005f5e100)
ByteVector(2 bytes, 0x0001)
ByteVector(100 bytes, 0xfda7146000520200000001b34e4345b6a07226f1aa91f2515096574fd6099eef1cd9885c8d069e5198cf760000000000000000000100c2eb0b000000001600146dd8f08240e4a3e25bfd3ebe43ae661b9cb3c02e0000000000000000ffffffff006c0000)
ByteVector(23 bytes, 0x160014b94dd1803e840d04b8f52332cff89ac6c42c4364)
ByteVector(8 bytes, 0x0000000000000005)
ByteVector(4 bytes, 0x00000064)
ByteVector(4 bytes, 0x000000c8)
```
You'll notice the 6th element is a pubkey as expected:
```
ByteVector(33 bytes, 0x03efd58eb7b7c79fef1ff1edde8fb0b49c2358020097c016e05c17a129af01940f)
```

NicolasDorier:
WTF so sorry @nkohen that was actually my implementation that mixed things up... I inverted in my test assertion actual and expected, so I got confused. Sorry for wasting your time :(

NicolasDorier:
however, @nkohen the parsed version above might be useful to include in the BIP.

Messaging.md
NicolasDorier: In that case, I don't understand the purpose of `max_witness_len`.
I can see what it is, but why is it needed?

Messaging.md
```diff
@@ -44,20 +61,114 @@ The following convenience types are also defined:
...
+
+This type of oracle info is for single-oracle, single signature (and hence single nonce) events.
+
+### The `funding_input` Type
+
+This type contains information about a specific input to be used in a funding transaction, as well as its corresponding on-chain UTXO.
+
+#### Version 0  `funding_input`
+
+1. type: 42772 (`funding_input_v0`)
```
NicolasDorier, line 112: This is quite confusing, in `contract_info_v0` we have a single TLV record for many outcomes. Here we have one TLV record for each input.

Messaging.md
```diff
@@ -44,20 +61,114 @@ The following convenience types are also defined:
...
+2. data:
+   * [`u16`:`prevtx_len`]
```
NicolasDorier, line 114: in theory not needed, as a parser knows how big is a transaction.

Messaging.md
```diff
@@ -44,20 +61,114 @@ The following convenience types are also defined:
...
+   * [`prevtx_len*byte`:`prevtx`]
+   * [`u32`:`prevtx_vout`]
+   * [`u32`:`sequence`]
+   * [`u16`:`max_witness_len`]
+   * [`u16`:`redeemscript_len`]
```
NicolasDorier: why, here the script is serialized as `u16` + `bytes` when everywhere else (payout script and change script), this is serialized as bitcoin script (Compact Size + bytes)

NicolasDorier:
Something wrong with vectors, I can't calculate the same `temporary_contract_id` as the Accept messages. I have the same byte array, tried all endianess possibilities.

The hash of the expected contract offer is different.

Messaging.md
nkohen: This is because contract information follows a uniform type (all outcomes fall under the same version, and if they don't a future version will have separate TLV records) whereas different inputs are independent (e.g. taproot input could be `funding_input_v1` and not have `prev_tx` at the same time as another input is `v0`)

Messaging.md
nkohen: I assume this is to allow clients which don't have transaction parser to make external calls to one after having parsed the transaction. Regardless of why I have made this record as close to this as possible: https://github.com/niftynei/lightning-rfc/pull/1/files#diff-3369c5aa1774fef2ff1e246979f223eaR169

Messaging.md
nkohen: I've done this to be consistent with https://github.com/niftynei/lightning-rfc/pull/1/files#diff-3369c5aa1774fef2ff1e246979f223eaR169 but perhaps I should rethink payout and change script to look similar to this.

NicolasDorier: I think either is fine, but mixing them is strange.

NicolasDorier:
If you don't have anymore that's OK, but it would be useful to get the oracle private key.

NicolasDorier:
Can you also provide the pre-images of the outcomes?

Messaging.md
nkohen: For deterministic fee computation

nkohen:
Going to merge spec changes only (test vectors now in #100) as these seem to be acked by basically everyone (and of course, we will iterate on the spec in future PRs)

PR #82, opened 2020-09-10
Use PSBT instead of custom format in offer/accept

NicolasDorier:
In retrospective, by far the hardest part of implementing DLCs spec was about calculating the fee rate of the funding transaction. (and this was even harder than implementing all the math stuff)

This PR is removing `num_funding_inputs`, `funding_inputs`, `change_spk`, `payout_spk`, `feerate_per_vb` from the protocol entirely, and replace it with `partial_funding_psbt` which convey the same information.

The advantage of it is that any wallet with basic support of PSBT can generate this `partial_funding_psbt`. So any wallet supporting PSBT can fund a DLC.

The feerate calculation is completely side stepped. Leaving it up to the end user to decide whether the resulting fund transaction has the appropriate feerate.

This allows both user to contibute what they want to the fee.

The CET/Refund transaction fee is half of the two partial funding psbt fee.

Example:
Sender has collateral of 40 000 sats and pay 4000 of fee. The receiver has collateral of 20 000 sats and pay 3000 of fee.

The funding transaction will have `(4000 + 3000) / 2 = 3500`  sats of fee. The funding output will be equals to `3500 + 40 000 + 20 000 = 63 500 sats`.
So the cet/refunds will also have 3500 sats of fee.

### DIFF
--- a/Protocol.md
+++ b/Protocol.md
@@ -133,6 +133,7 @@ The receiving node MUST reject the contract if:
   - `funding_pubkey` is not a valid secp256k1 pubkey in compressed format.
   - The inputs in `partial_funding_psbt` do not contribute at least `total_collateral_satohis`.
   - the fee rate of the `partial_funding_psbt` is below twice the minimum fee rate relay fee.
+  - the inputs in `partial_funding_psbt` does not contains `PSBT_IN_REDEEM_SCRIPT` and `PSBT_IN_WITNESS_SCRIPT` if inputs are P2SH or P2WSH.
 
 ### The `accept_dlc` Message
 
--- a/Protocol.md
+++ b/Protocol.md
@@ -70,12 +70,8 @@ the funding transaction and CETs.
    * [`contract_info`:`contract_info`]
    * [`oracle_info`:`oracle_info`]
    * [`point`:`funding_pubkey`]
-   * [`spk`:`payout_spk`]
+   * [`psbt`:`partial_funding_psbt`]
    * [`u64`:`total_collateral_satoshis`]
-   * [`u16`:`num_funding_inputs`]
-   * [`num_funding_inputs*funding_input`:`funding_inputs`]
-   * [`spk`:`change_spk`]
-   * [`u64`:`feerate_per_vb`]
    * [`u32`:`contract_maturity_bound`]
    * [`u32`:`contract_timeout`]
 
@@ -91,20 +87,15 @@ blockchains opened to the same peer (if it supports the target chains).
 specifies the oracle(s) to be used as well as their commitments to events.
 
 `funding_pubkey` is the public key in the 2-of-2 multisig script of
-the funding transaction output. `payout_spk` specifies the script
-pubkey that CETs and the refund transaction should use in the sender's output.
+the funding transaction output.
 
 `total_collateral_satoshis` is the amount the sender is putting into the
-contract. `num_funding_inputs` is the number of funding inputs contributed by
-the sender and `funding_inputs` contains outputs, outpoints, and expected weights
-of the sender's funding inputs. `change_spk` specifies the script pubkey that funding
-change should be sent to.
-
-`feerate_per_vb` indicates the fee rate in satoshi per virtual byte that both
-sides will use to compute fees in the funding transaction, as described in the
-[transaction specification](Transactions.md).
+contract.
 
 `contract_maturity_bound` is the nLockTime to be put on CETs. `contract_timeout` is the nLockTime to be put on the refund transaction.
+
+`funding_psbt` is a funded PSBT with one output paying exactly `total_collateral_satoshis` and optionally another change output.
+The script pubkey of the output paying `total_collateral_satoshis` is the script pubkey that CETs and the refund transaction should use in the sender's output.
 
 #### Requirements
 
@@ -119,11 +110,8 @@ The sending node MUST:
 
 The sending node SHOULD:
 
-  - set `feerate_per_vb` to at least the rate it estimates would cause the transaction to be immediately included in a block.
   - set `contract_maturity_bound` to no later than the earliest expected oracle signature time.
   - set `contract_timeout` sufficiently long after the latest possible oracle signature added to all other delays to closing the contract.
-  - set `payout_spk` to a previously unused script public key.
-  - set `change_spk` to a previously unused script public key.
 
 The receiving node MUST:
 
@@ -135,17 +123,16 @@ The receiving node MAY reject the contract if:
   - the `contract_info` is missing relevant events.
   - it does not want to use the oracle(s) specified in `oracle_info`.
   - `total_collateral_satoshis` is too small.
-  - `feerate_per_vb` is too small.
-  - `feerate_per_vb` is too large.
+  - the fee rate of the `partial_funding_psbt` is insufficient
 
 The receiving node MUST reject the contract if:
 
   - the `chain_hash` value is set to a hash of a chain that is unknown to the receiver.
   - the `contract_info` refers to events unknown to the receiver.
   - the `oracle_info` refers to an oracle unknown or inaccessible to the receiver.
-  - it considers `feerate_per_vb` too small for timely processing or unreasonably large.
   - `funding_pubkey` is not a valid secp256k1 pubkey in compressed format.
-  - `funding_inputs` do not contribute at least `total_collateral_satohis` plus full [fee payment](Transactions.md#fee-payment).
+  - The inputs in `partial_funding_psbt` do not contribute at least `total_collateral_satohis`.
+  - the fee rate of the `partial_funding_psbt` is below twice the minimum fee rate relay fee.
 
 ### The `accept_dlc` Message
 
@@ -159,10 +146,7 @@ and closing transactions.
    * [`32*byte`:`temporary_contract_id`]
    * [`u64`:`total_collateral_satoshis`]
    * [`point`:`funding_pubkey`]
-   * [`spk`:`payout_spk`]
-   * [`u16`:`num_funding_inputs`]
-   * [`num_funding_inputs*funding_input`:`funding_inputs`]
-   * [`spk`:`change_spk`]
+   * [`psbt`:`partial_funding_psbt`]
    * [`cet_signatures`:`cet_signatures`]
    * [`signature`:`refund_signature`]
 
@@ -188,8 +172,6 @@ The receiver:
     - MAY reject the contract.
   - if `cet_signatures` or `refund_signature` fail validation:
     - MUST reject the contract.
-- if `funding_inputs` do not contribute at least `total_collateral_satohis` plus [fee payment](Transactions.md#fee-payment)
-  - MUST reject the contract.
 
 Other fields have the same requirements as their counterparts in `offer_dlc`.
 
--- a/Transactions.md
+++ b/Transactions.md
@@ -14,9 +14,6 @@ This details the exact format of all on-chain transactions, which both sides nee
     * [Offerer Output](#offerer-output)
     * [Accepter Output](#accepter-output)
 * [Refund Transaction](#refund-transaction)
-* [Fees](#fees)
-  * [Expected Weight of the Funding Transaction](#expected-weight-of-the-funding-transaction)
-  * [Expected Weight of the Contract Execution or Refund Transaction](#expected-weight-of-the-contract-execution-or-refund-transaction)
 * [Test Vectors](#test-vectors)
 * [References](#references)
 * [Authors](#authors)
@@ -30,9 +27,7 @@ The funding inputs and change output script public keys are negotiated in the of
 
 ## Funding Transaction Input and Output Ordering
 
-The Offerer/Initiator's inputs come first and in the order that they were listed in the Offer message, the Accepter's inputs come after this in the order they are listed in the Accept message. The funding output comes first, followed by the Offerer's change output if applicable, and then the Accepter's change output if applicable.
-
-Note that this will likely change in the future.
+The Offerer/Initiator's inputs come first and in the order that they were listed in the Offer message, the Accepter's inputs come after this in the order they are listed in the Accept message. The funding outputs comes first, followed by the Offerer's outputs if applicable, and then the Accepter's change output if applicable.
 
 ## Funding Inputs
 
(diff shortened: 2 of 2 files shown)


NicolasDorier:
if you don't want to use psbt for some reason, at least you should let `feerate_per_vb` go and replace it by `contributed_fee` in both offer and accept message. I can make another PR if you prefer than route.

benthecarman:
Concept ACK

I think this will make things a lot easier. Like you note the fee calculations should no longer be an issue. This should also make it easier to fund from a something like a hardware wallet or any separate wallet (ie bitcoind).

One thing we will need to be mindful of in the spec is detailing how the implementation should be verifying the output values/fees.

Protocol.md
```diff
@@ -135,17 +123,16 @@ The receiving node MAY reject the contract if:
   - the `contract_info` is missing relevant events.
   - it does not want to use the oracle(s) specified in `oracle_info`.
   - `total_collateral_satoshis` is too small.
-  - `feerate_per_vb` is too small.
-  - `feerate_per_vb` is too large.
+  - the fee rate of the `partial_funding_psbt` is insufficient
```
benthecarman, line 127: There should probably be a requirement that the `PSBT_IN_WITNESS_UTXO ` field is filled for each input

and the `PSBT_IN_REDEEM_SCRIPT ` and `PSBT_IN_WITNESS_SCRIPT ` are filled for inputs that have redeem/witness scripts.

Without these you cannot correctly calculate the fee.

Protocol.md
```diff
@@ -159,10 +146,7 @@ and closing transactions.
    * [`32*byte`:`temporary_contract_id`]
    * [`u64`:`total_collateral_satoshis`]
    * [`point`:`funding_pubkey`]
-   * [`spk`:`payout_spk`]
-   * [`u16`:`num_funding_inputs`]
-   * [`num_funding_inputs*funding_input`:`funding_inputs`]
-   * [`spk`:`change_spk`]
+   * [`psbt`:`partial_funding_psbt`]
```
benthecarman, line 150: Should denote if this `partial_funding_psbt` should have the offerer's inputs and outputs or not.

(I think it should).

benthecarman: Maybe a way to solve #18 is here and we can just have the accepter due the shuffling

benthecarman: If we are having the offerer's inputs and outputs in this psbt it should be noted the they must verify they are all included correctly.

NicolasDorier: > Should denote if this partial_funding_psbt should have the offerer's inputs and outputs or not.

I think it should not, it should be done exactly the same way as the offerer. The goal is that such partial funding PSBT can be created out of any wallet with PSBT support just by creating a transaction sending the collateral to yourself. If we need to combine with other inputs, the process will be more complicated. (that said, a DLC tool could easily translate a PSBT from a wallet in the first form, to the later form, it is just a bit more complicated)

There is also another reason: We don't want the offerer to have to validate the PSBT of the acceptor beyond what is strictly necessary. It is way better that both ther acceptor and offerer build the funding transaction by themselves, rather than the offerer not building the funding transaction, but have to verify that the acceptor did it right.

And if the offerer knows how to do it right, it could just do it himself in the first place.

The only reason it would make sense for the acceptor to add the offerer inputs, is if we wanted to let discretion on how to build the funding, up to the acceptor, giving him flexibility. So, instead of both agreeing on how to build the funding transaction, the acceptor decides how to build it, and the offerer enforces some rules upon it.
I think this is fine as well. No strong opinion on one way or the other. But I favor a bit on having the protocol "symmetrical" (where acceptor don't put the inputs of the offerer in the partial psbt)

Protocol.md
NicolasDorier: https://github.com/discreetlogcontracts/dlcspecs/pull/82/commits/6589f5e2ef4543ef3ff67bf14b8526a67cfb9c23

Protocol.md
benthecarman: Thank you for the thorough explanation. I agree with you now, we should have the accepter only include their own inputs and outputs

Tibo-lg:
NACK for me. Passing PSBT around is going to bloat messages. I think PSBTs are really nice and I don't think there is anything preventing using them outside of the protocol, but it doesn't make sense to me to send them over the wire, especially if we start using LN stack and have some tight packet constrains. 

Also something I am not sure I fully understand. Even if you pass a PSBT, you still somehow need to decide how much fee to pay right? How do you do this computation? Don't you anyway need to compute the size of you inputs and potential change output to match a given fee-rate?

Similarly on the receiver side, don't you also need to check that the provided fee is proper?

> if you don't want to use psbt for some reason, at least you should let feerate_per_vb go and replace it by contributed_fee in both offer and accept message. I can make another PR if you prefer than route.

In that case I guess you would still need to compute the size of the contributed data in order to determine whether it is an appropriate amount or not right? I'm sorry if I'm missing something here, but it looks to me like PSBT doesn't really solve the problem, unless what you are proposing is to just accept whatever fee the counter party has chosen without checking anything.

Finally I agree that fee computation is a big pain and source of error. But I think that is mainly because how to compute them is not specified at the moment. If we just add some pseudo code in the specs and some test vectors, I don't think it will be that difficult.

NicolasDorier:
Why do you feel PSBT to be bloated? there is no superfluous information in it.
For example, the current protocol does not support funding from P2SH segwit wallets.

PSBT already specified all of it, so there is no need to re-implement the wheel.

> In that case I guess you would still need to compute the size of the contributed data in order to determine whether it is an appropriate amount or not right? I'm sorry if I'm missing something here, but it looks to me like PSBT doesn't really solve the problem, unless what you are proposing is to just accept whatever fee the counter party has chosen without checking anything.

It solves the problem. The pain point is that currently both parties need to be in agreement down to the satoshi on how much is spent on fee. This sidestep the problem completely by not requiring it. If both party have small differences in how they calculate the estimated fee rate once the transaction is completely signed is not a problem anymore.

There is way to solve this problem without PSBT

> if you don't want to use psbt for some reason, at least you should let feerate_per_vb go and replace it by contributed_fee in both offer and accept message. I can make another PR if you prefer than route.

But I don't see why reinventing the wheel. PSBT already specify how the information needed is formatted, and most wallet can generate it. There is no superfluous data.

> If we just add some pseudo code in the specs and some test vectors, I don't think it will be that difficult.

It is difficult. The spec has this pseudo code, but really relevant details like how you round down odds number or whether you use weight or virtual byte, or whether you add before dividing or you divide before adding affect the result.

As an example:
Imagine that we both manage to agree on the estimated weight of funding is X, and the max CET size is Y, there are at least 8 ways to calculate the output values.

Should the fund output be `TotalCollateral + (Y/4) * fee_rate`: What if Y is odd? Round up? Round down?

Now about the change of the offerer: Should it be 

* `TotalInput - Collateral - (Y + X)/2/4 * fee_rate`
* `TotalInput - Collateral - (Y + X)/8 * fee_rate` 
* `TotalInput - Collateral - (Y/4/2 + X/4/2) * fee_rate` 
* `TotalInput - Collateral - (Y/8 + X/8) * fee_rate` 

If we keep the remainder satoshi, should we just add those satoshi to the second output or not?

Or maybe the fee_rate should be converted into s/wu instead of s/vb first so we only divide once? Maybe not? I don't know that's not specified.

Note that on real number all are the same, but not when you work with integers.

There is basically 8 ways (actually maybe more, since it goes up combitionally) to calculate output values even if we come into agreement on the weight of the transaction. (and I did not even managed to reach this step with the specification)

This is useless bloat to the specification. Look this PR, it removes 220 lines.
Those 220 lines were not specific enough to properly calculate the weight properly and were not even talking about how to handle integers divisions. I think if it was fully specified, it could take up to 500 lines.

Tibo-lg:
So correct me if I'm wrong, but I feel you are saying hat PSBT solves two separate issues:
1. Formatting of data
2. Fee computation

> But I don't see why reinventing the wheel. PSBT already specify how the information needed is formatted, and most wallet can generate it. There is no superfluous data.

That's 1. I agree with you on that point, but the issue is that I'm afraid that a single PSBT might not fit into the size limit of LN messages. Then we will need to start dividing it into smaller chunks, and I'm not sure we will end up in a better place than having custom messages.

> It solves the problem. The pain point is that currently both parties need to be in agreement down to the satoshi on how much is spent on fee. This sidestep the problem completely by not requiring it. If both party have small differences in how they calculate the estimated fee rate once the transaction is completely signed is not a problem anymore.

> There is way to solve this problem without PSBT

That's 2. And as you're saying, there is way to solve it without PSBT, and I would actually say that PSBT does not help for this. For compatibility, we could add a `fund_output` value given by the offerer that the receiver should accept if it's within a reasonable range of his own computation. But you still need to compute how much fee you need to pay and verify how much fee your counter party is paying.

In you change you put:
> the fee rate of the `partial_funding_psbt` is insufficient

But how do I know it is insufficient? I need to first compute it right? And we need to agree to some extent on how should this be computed, so someone wanting to make this computation will have to go looking for the 220 lines you deleted somewhere else, and maybe end up with a computation that is way off what others have implemented. Personally I much prefer having something written down in the specs that I can follow to implement and some tests vector so that I'm sure that things are working as expected.

> Those 220 lines were not specific enough to properly calculate the weight properly and were not even talking about how to handle integers divisions.

I agree with you on that. I don't think what is currently in the specs is enough, or even useful. And this is not what I was referring to when I talked about having pseudo code. Actually maybe the best would be to have a reference implementation so that there could be no room for confusion (and I again I think this would be useful regardless of whether we use PSBT or no). Then we could just say "The fee computation has to match this implementation (link) and pass this test vectors" and be done with it (and then maybe remove the 220 lines of explanation as I think in that case maybe code is better than words).

Anyway it's not like I will fight to death over that :) . If everybody is convinced that's the way to go I'll implement it, but I think it is trying to hide issues that will come and bite us later.

devrandom:
NACK regarding PSBT concept.  I believe that PSBT is useful for interoperability between wallets and signers that don't otherwise have a common protocol.  I don't think it makes sense to nest the PSBT format inside another format - it increases complexity and message sizes.  Therefore I would stick to the general format described in the DLC specs.

PSBT can still be used for local signing purposes.

I haven't looked at the fee calculation issues, so I'm not commenting about those.

NicolasDorier:
> That's 1. I agree with you on that point, but the issue is that I'm afraid that a single PSBT might not fit into the size limit of LN messages. Then we will need to start dividing it into smaller chunks, and I'm not sure we will end up in a better place than having custom messages.

You will have to do that anyway, since the size depends on the number of funding inputs, which is controlled by the peers what if the user has 10 inputs?

> But how do I know it is insufficient? I need to first compute it right?

"You received this offer, the offerer is paying for 10 sat/bytes, do you accept?".

The value "10 sat/bytes" is computed, but it does not need to be the exact same way across implementation, this is a UI concern, in practice if the fee rate is 10.01 sat/bytes instead of 9.997 sat/bytes, it should not cause the whole exchange to fail, as from user perspective there is no difference.

Whether a fee is sufficient is a subjective concern, it can't be computed, but it need to be ACK by the user, it should not be decided by the protocol. (Except that it is above the minimum relay fee)

Tibo-lg:
> You will have to do that anyway, since the size depends on the number of funding inputs, which is controlled by the peers what if the user has 10 inputs?

Of course, it's actually because we will likely need to have multiple messages that I think PSBT don't offer a big advantage.

> The value "10 sat/bytes" is computed, but it does not need to be the exact same way across implementation, this is a UI concern, in practice if the fee rate is 10.01 sat/bytes instead of 9.997 sat/bytes, it should not cause the whole exchange to fail, as from user perspective there is no difference.

So first do we agree that the fee computation issue is dissociated from the use of PSBT?

Otherwise I agree that we could allow some error margin in the fee computation as I mentioned above. But that doesn't really change the fact that anybody wanting to implement the protocol will have to implement something to compute fees, and so I think that it has its place in the specifications, unless there is another specification for this that we could simply refer to (that would actually be the best).

NicolasDorier:
> So first do we agree that the fee computation issue is dissociated from the use of PSBT?

Yes. That's why I proposed another way to fix the fee issue in the event you don't want PSBT (https://github.com/discreetlogcontracts/dlcspecs/pull/82#issuecomment-689928845)

>  But that doesn't really change the fact that anybody wanting to implement the protocol will have to implement something to compute fees, 

What I am trying to do is to make sure that people not calculating it the same way don't have frictions. Most Bitcoin libraries already do that out of the box, and people can just reuse this. I am quite sure none of them have same results.

ariard:
NACK

I think this PR have the following issues :
* free-rider behavior, "my feerate is lower than yours but my weight is higher thanks to aggregate my inputs"
* protocol complexity, know you parse have to support both PSBT format AND still the format used to announce DLC non-transaction data
* pinning issue, by adopting a low feerate it's easier to stuck the funding in the mempool, opening room for exploitation
* future-compatibility, when we move off-chain, we will rely on channel funding transaction, and it's really likely it won't support PSBTs, so DLC wallet will have to understand TLV format
* transaction malleability, not committing to the feerate (`hashPrevouts`, `hashOutputs`) and only using SIGHASH_SINGLE | SIGHASH_ANYONECANPAY means I can still your fee before broadcast

Multi-party _distrusted_ transactions don't have the same requirement than normal transactions and thus sanitizing the feerate is a necessity, not only a matter of UX.

> What I am trying to do is to make sure that people not calculating it the same way don't have frictions

The issue about calculating is a real one, but more a ecosystem-wise one and it would be better to have a BIP once for all, reusable by any protocol. Telling you to use weight, how to share common fields cost and a way to round up.

nkohen:
NACK-with-caveat

I think for many reasons people have listed and my opinion that PSBTs only really make any sense in manual, non-algorithmic contexts this isn't a good way to go, especially seeing as we're hoping to make DLCs entirely algorithmic without too many manual steps (especially during construction).

That said, I would like to re-iterate that if in the future we come across a use case where PSBT makes sense (say there's a lot of tech-savy OTC traders who want more freedom and are doing validation somewhat manually), then we can always introduce a new TLV type for an offer that just wraps a PSBT or alternatively a new contract flags byte could signal that you want to do the PSBT version of contract negotiation instead of the algorithmic one.

Either way I don't think the deleted lines in this PR should be deleted but I'm open to the idea that someday a use case is found for PSBTs and a new version/contract flag is deployed to add support for this.

That said, I definitely think this is out-of-scope for v0.1

nkohen:
> if you don't want to use psbt for some reason, at least you should let `feerate_per_vb` go and replace it by `contributed_fee` in both offer and accept message. I can make another PR if you prefer than route.

In response to this comment, you should know that the plan for fee computation (which I am beginning to work on) is to be similar to the [dual funding proposal](https://github.com/niftynei/lightning-rfc/pull/1/files#diff-3369c5aa1774fef2ff1e246979f223eaR169) where users give max witness size when communicating their funding inputs and later when inputs are added during signing messages, these max's are verified. Fee computation then becomes quite simple in a manner similar to what you are suggesting where rather than doing actual size computations one just adds up all the max sizes and uses that as the computed size to be used in fee calculation.

NicolasDorier:
@nkohen awesome, I think this field would solve the main pain, closing this.

PR #87, opened 2020-09-14
Add Non-Interactive Protocol 

### LINKED ISSUE #78: On-Chain Handling
@ariard has volunteered to write a specification draft for handling and watching of on-chain events.

This includes things like double-spending funding inputs, mempool pinning issues, and more!

ariard:
This spec draft should cover :
* offchain to onchain reactions
* onchain to onchain reactions
* security bumping strategy
* reorgs
* out-of-protocol outputs tracking
* diverse recommendations w.r.t base layer

Fix #78

### DIFF
--- a/Non-Interactive-Protocol.md
+++ b/Non-Interactive-Protocol.md
@@ -1,0 +1,250 @@
+# Non-Interactive Onchain Protocol
+
+# Table of Contents
+
+* [Abstract](#abstract)
+* [General Requirements](#general-requirements)
+* [Fee-Bumping Policy](#fee-bumping-policy)
+* [Funding Phase](#funding-phase)
+* [Execution Phase](#execution-phase)
+* [Reorgs](#reorgs)
+
+## Abstract
+
+DLC allows two parties to conduct stateful bitcoin contracts by giving each
+of the parties *cross-signed range of CET transactions*. These CETs transactions
+are all spending the same output from a funding transaction and thus only one of
+CET can confirm. A CET is finalized according to the outcome signed by the previously
+selected oracle.
+
+As the contract is established once funding transaction signatures have been exchanged,
+from then a party should assume counterparty non-cooperation and ensure either
+confirmation of the funding or double-spend of its collatral inputs to avoid exploitation.
+
+Once funding transaction signatures are exchanged, DLC parties aren't required
+to interact anymore. They are three scenarios to consider:
+
+1. The funding transaction doesn't confirm, a party must either bump its feerate
+to confirm before DLC maturing or double-spend its funding collateral, thus cancelling the
+DLC.
+
+2.  The funding transaction confirms, the oracle releases the outcome signature, a
+participant must broadcast the DLC transaction paired with the outcome.
+
+3. The funding transaction confirms, oracle doesn't release the outcome signature, independently
+of counterparty behavior, the party must broadcast the refund transaction after its timelock
+expiration.
+
+# General Requirements
+
+A DLC client should have censorship-resistant access to the blockchain, use a
+local fee-estimator and ensure it's well-connected to the tx-relay network.
+
+We recommend to not react on any mempool event as otherwise it would be a source of
+cross-layer mapping. <sup>[Cross-layer Mapping](https://arxiv.org/pdf/2007.00764.pdf)</sup>
+
+We recommend to not produce local non-adaptor signatures for transaction until their broadcast
+is required to avoid their misuage in case of site compromise, assuming an external signer.
+
+Further, we recommend duplicate backups of counterparty signatures to avoid relying on
+counterparty honest behavior in case of primary storage failure.
+
+Transaction finalization is not implying a mandatory key material access.[ FIXME: Client Key Management. ]
+
+# Fee bumping policy
+
+An onchain state machine must take actions based on different timers parameterized by three
+values: `congestion_bump_frequency`/`security_bump_frequency`/`security_point`/`refund_security_point`.
+Their definition and setting recommendations are laid out in this section, their usage is precised
+in following sections.
+
+A `security_point` is a client heuristic after which the confirmation of the funding transaction
+is a matter of contract safety, otherwise the user is exposed to a bad-player double-spend if either
+the outcome odds are modified or a signature oracle is released. This point is function of event
+signing time, which may be static or dynamic, we defer point qualification to client.
+
+If the signing time is static (block height, epoch) or predictable we recommend setting the 
+`security_point` with a confirmation buffer of 20 blocks compared to signing time.
+
+The `refund_security_point` is a client heuristic after which the confirmation of the refund
+transaction is a matter of contract safety, otherwise the user is exposed to a malicious CET spend
+if the oracle key material is compromised. This point is function of the level of trust in the
+oracle, we defer point qualification to client.
+
+Client's `congestion_bump_frequency`/`security_bump_frequency` can be scheduled at least on three different basis:
+- a block height
+- a local clock
+- an observed feerate fluctuation
+
+Note, these criterias might be observable by an adversary and we recommend to implementations
+to pad any criteria with a random delay to diminish risks of tx-relay jamming.
+
+We don't provide recommendations for the `congestion delay` as its settings depends of user
+liquidity preferences and available local data to predict the feerate. Confirmation of non
+time-sensitive transactions should happen but a user may prefer to delay them in case of
+mempool-congestion.
+
+We provide recommendations for the `security_bump_frequency` as any transaction under this fee-bumping
+policy has a time-sensitive confirmation. A delay of 1 block and a fast-confirmation feerate
+as reported by fee-estimator should be preferred.
+
+Dividing the fee-bumping occurence between different frequencies avoid a client overpaying in
+fees in case of mempool spikes, while anticipating that at some point during DLC execution, the
+confirmation of the funding is a security concern.
+
+# Funding Phase
+
+A party always starts the protocol in the funding phase. The funding phase can
+terminate by either a transition to the execution phase or a protocol abort.
+
+## Requirements
+
+A node:
+
+  - if a counterparty's funding signature is received:
+    - MUST finalize the funding transaction with a local `SIGHASH_ALL` signatures
+    - MUST broadcast the funding transaction
+
+A node:
+
+  - if a funding signature has been sent or a funding transaction broadcast and there is no confirmation after 6 blocks:
+    - MUST spend the change output with a CPFP increasing package feerate to shorten confirmation time
+    - MUST start a fee-bumping timer of length `congestion_bump_frequency`
+    - if `congestion_bump_frequency` expires:
+      - MUST spend the change output with a CPFP increasing package feerate to shorten confirmation time
+      - MUST replace the CPFP transaction according to BIP 125
+      - MUST reschedule timer for next bump
+    - if `security_point` is reached:
+      - MUST spend a funding input with a higher-feerate transaction/absolute fee than any previously propagated package feerate
+      - MUST start a fee-bumping timer of length `security_bump_frequency`
+      - MUST disable `congestion_bump_frequency` timer
+    - if `security_bump_frequency` expires:
+      - MUST spend a funding input with a higher-feerate transaction/absolute fee than any previously propagated package feerate
+      - MUST replace the CPFP transaction according to BIP 125
+      - MUST reschedule timer for next bump
+    - MUST NOT spend more in bumping fees than its collateral input value
+    - if outcome odds are unfavorable:
+      - MAY disable funding fee-bumping
+
+A node:
+
+  - if the funding transaction confirms:
+    - MUST track spend of the funding output
+    - MUST advances its off-chain state to the execution phase
+    - MUST note its change output for later spend
+
+A node:
+
+  - if the funding input double-spend confirms:
+    - MUST abort the DLC and prevent further interactions with counterparty
+
+## Rationale
+
+Funding transaction must be confirmed to consider the off-chain state initialized. As mempools
+congestion may fluctuate between fees negotiation and transaction propagation, party must adjust
+unilaterally the package feerate (CPFP) according to their confirmation preferences. A party may 
+choose to wait for mempools draining before to adjust the feerate to avoid a fee waste as block 
+height of transition to execution phase doesn't matter, assuming it's before `contract_maturity_bound`.
+
+Reaching near `contract_maturity_bound` requires to confirm the funding transaction quickly as it's
+now a security matter. After oracle signature release, a malicious party could try to double-spend
+an unconfirmed funding transaction to avoid losing its collateral, and thus without assuming deceitful
+cooperation with the oracle.
+
+In case of counterparty funding input double-spend, cleaning offchain state reduce further exploitation
+at this level. E.g a DLC implementation feeded with a _correct_ signature whereas the contract is
+invalid but still displaying an incorrect information to the user. A double-spending counterparty
+is also "breaking" the previously negotiated DLC and agreed by its signature exchange. Thus further
+interactions should be warranted with this buggy/malicious peer.
+
+# Execution Phase
+
+The execution phase transitions to the terminal phase either by a CET or a refund
+transaction confirmations.
+
+## Requirements
+
+A node:
+
+  - if an oracle signature is received:
+    - MUST finalize the CET with a local `SIGHASH_ALL` signature
+    - MUST broadcast the CET
+    - MAY spend its outcome output with a CPFP increasing package feerate to shorten confirmation time
+    - if the CET is not confirmed and `contract_timeout` is 20 blocks ahead of local tip:
+      - MUST rebroadcast the CET
+      - MUST spend its outcome output with a CPFP increasing package feerate to shorten confirmation time
+      - MUST start a fee-bumping timer of length `security_bump_frequency`
+    - if `security_bump_frequency` expires:
+      - MUST spend the change output with a replacing CPFP increasing package feerate to shorten confirmation time
+      - MUST replace the CPFP transaction according to BIP 125
+      - MUST reschedule timer for next bump
+    - MUST NOT spend more in bumping fees than its outcome value
+    - MAY NOT spend more in bumping fees than its outcome value - refund value if oracle is trusted
+    - if the outcome realization is unfavorable:
+      - MAY disable CET fee-bumping
+
+A node:
+
+  - if a CET transaction confirms:
+    - MUST note its outcome output for later spend
+    - MUST advances its off-chain state to terminal phase
+
+A node:
+
+  - if no oracle signature has been received and `contract_timeout` is reached:
+    - MUST finalize the refund transaction with a local `SIGHASH_ALL` signature
+    - MUST broadcast the refund transaction
+    - if the refund is not confirmed after `congestion_bump_frequency`:
+      - MUST spend the change output with a CPFP increasing package feerate to shorten confirmation time
+      - MUST start a fee-bumping timer of length `refund_congestion_bump_frequency`
+    - if `refund_congestion_bump_frequency` expires:
+      - MUST spend the change output with a replacing CPFP increasing package feerate to shorten confirmation time
+      - MUST replace the CPFP transaction according to BIP 125
+      - MUST reschedule timer for next bump
+    - if `refund_security_point` is reached:
+      - MUST spend a funding input with a higher-feerate transaction/absolute fee than any previously propagated package feerate
+      - MUST start a fee-bumping timer of length `security_bump_frequency`
+      - MUST disable `refund_congestion_bump_frequency` timer
+    - if `security_bump_frequency` expires:
+      - MUST spend a funding input with a higher-feerate transaction/absolute fee than any previously propagated package feerate
+      - MUST replace the CPFP transaction according to BIP 125
+      - MUST reschedule timer for next bump
+
+A node:
+
+  - if a refund transaction confirms:
+    - MUST note its outcome output for later spend
+    - MUST advances its off-chain state to terminal phase
+
+
+## Rationale
+
+Reaching near `contract_timeout` requires to confirm the valid CET quickly as it's now a security 
+matter. A malicious party could try to cancel a lost DLC by confirming the refund transaction.
+
+A refund transaction confirmation is less sensitive but it should noted that in the meanwhile
+the oracle key material is toxic, as in case of compromise, a CET could be signed and confirmed
+by a malicious counterparty. This confirmation should be managed in function of the user trust
+towards the oracle.
+
+# Reorgs
+
+Reorgs must be handled by clients as they are a source of funds loss in case of mishandling.
+
+## Requirements
+
+A node:
+
+  - if a funding transaction is disconnected
+    - MUST roll back its state to funding phase starting
+    - if a CET or refund transaction is disconnected
+      - MUST roll back its state to execution phase starting
+
+
+# Authors
+
+Antoine Riard <antoine.riard@gmail.com>
+
+![Creative Commons License](https://i.creativecommons.org/l/by/4.0/88x31.png "License CC-BY")
+<br>
+This work is licensed under a [Creative Commons Attribution 4.0 International License](http://creativecommons.org/licenses/by/4.0/).


Non-Interactive-Protocol.md
```diff
@@ -0,0 +1,173 @@
...
+transaction confirmations.
+
+## Requirements
+
+A node:
+	- if an oracle signature is received:
+		- MUST finalize the CET with a local `SIGHASH_ALL` signature
+		- MUST broadcast the CET
+		- MAY spend its outcome output with a CPFP increasing package feerate to confirm soonly
+		- if the CET is not confirmed and `contract_timeout is 20 blocks ahead of local tip:
```
benthecarman: missing closing `

Non-Interactive-Protocol.md
```diff
@@ -0,0 +1,173 @@
...
+
+2. The oracle doesn't release the outcome signature, independently of counterparty
+behavior, the party must broadcast the refund transaction after its timelock expiration.
+
+# Table of Contents
+	* [General Requirements](#general-requirements)
+	* [Funding Phase](#funding-phase)
+	* [Execution Phase](#execution-phase)
+	* [Fee-Bumping Policy](#fee-bumping-policy)
+	* [Reorgs] (#reorgs)
```
benthecarman: Need to remove the space here to create the link

Non-Interactive-Protocol.md
```diff
@@ -0,0 +1,173 @@
...
+
+Client's `congestion_delay`/`security_delay` can be scheduled on three different basis:
+- a block height
+- a local clock
+- an observed feerate fluctuation
+
+We don't provide recommendations for the congestion delay as its settings depends of user
+liquidity preferences and available local data to predict the feerate.
+
+As `security_delay` must be offensive, we recommend to set it to 1 block.
```
benthecarman: I think this section should be right after `General Requirements`, that way you have a better idea what these terms mean while reading.

Non-Interactive-Protocol.md
```diff
@@ -0,0 +1,173 @@
...
+
+Funding transaction must be confirmed to consider the off-chain state initialized. As mempools
+congestion may fluctuate between fees negotiation and transaction propagation, party must adjust
+unilaterally feerate according to their confirmation preferences. A party may choose to wait
+for mempools draining before to adjust the feerate to avoid a fee waste as block height of 
+transition to execution phase doesn't matter, assuming it's before `contract_maturity_bound`.
+
+Reaching near `contract_maturiy_bound` requires to confirm the funding transaction quickly as it's
+now a security matter. After oracle signature release, a malicious party could try to double-spend
+an unconfirmed funding transaction to avoid loosing its collateral.
```
benthecarman: > loosing

losing

Non-Interactive-Protocol.md
```diff
@@ -0,0 +1,173 @@
...
+# Funding Phase
+
+A party always starts the protocol in the funding phase. The funding phase can
+terminate by either a transaction to the execution phase or a protocol abort.
+
+## Requirements
+
+A node:
+	- if a counterparty's funding signature is received:
+		- MUST finalize the funding transaction with a local `SIGHASH_ALL` signature
```
benthecarman: Should a node do something if they see their counter party broadcast the tx with a different sig hash type?

Non-Interactive-Protocol.md
```diff
@@ -0,0 +1,173 @@
...
+		- MUST NOT spend more in bumping fees than its outcome value
+
+A node:
+	- if a CET transaction confirms:
+		- MUST note its outcome output for latter spend
+		- MUST advances its off-chain state to terminal phase
+
+A node:
+	- if `contract_timeout` is reached:
+		- MUST finalize the CET with a local `SIGHASH_ALL` signature
```
benthecarman: > MUST finalize the CET

should this be `MUST finalize the refund transaction`

benthecarman (changes requested):
Looks good overall

nit: You used soonly in quite a few places, but I don't think it is a word

Non-Interactive-Protocol.md
```diff
@@ -0,0 +1,173 @@
...
+In the optimistic way, once the oracle signature is released, parties interact
+off-chain to settle the DLC. They generate a *closing_tranaction* and publish it on the
+blockchain.
+
+In the pessimitic way, a party interact with the chain to unilaterally settle the DLC.
+They are two scenarios to consider:
+
+1.  The oracle releases the outcome signature but the DLC counterparty goes wrong,
+either going offline or maliciously behaving. The party must broadcast the DLC
+transaction paired with the outcome.
```
Tibo-lg: With adaptor version, that's not the case anymore. There is no need for interaction and there is no closing transaction. They just use the oracle signature to decrypt one of the adaptor signature from their counter party and any of them can broadcast the corresponding CET.

So AFAIK as soon as the fund transaction is confirmed, there is no way for any party to misbehave (apart from colluding with the oracle but that's another story).

Non-Interactive-Protocol.md
```diff
@@ -0,0 +1,173 @@
...
+# General Requirements
+
+A DLC client should have censorship-resistant access to the blockchain, use a
+local fee-estimator and ensure it's well-connected to the tx-relay network.
+
+We recommend to not react on any mempool event as otherwise it would be a source of
+cross-layer mapping.
+
+We recommend to not sign local transactions until their broadcast is required
+to avoid their misuage in case of site compromise, assuming an external signer.
```
Tibo-lg: Again with adaptor version there is no really notion of local transaction. You need to generate the adaptor signature for the CETs which are the same for both parties. But maybe you can mention that they should not store them?

Non-Interactive-Protocol.md
```diff
@@ -0,0 +1,173 @@
...
+		- MUST finalize the funding transaction with a local `SIGHASH_ALL` signature
+		- MUST broadcast the funding transaction
+
+A node:
+	- if a funding signature has been sent or a funding transaction broadcast and there is no confirmation after 6 blocks:
+		- MUST spend the change output with a CPFP increasing package feerate to confirm soonly
+		- MUST start a bumping timer of length `congestion_delay`
+		- if `congestion_delay` expires:
+			- MUST spend the change output with a CPFP increasing package feerate to confirm soonly
+			- MUST replace according to BIP 125
```
Tibo-lg: I understand but it would be clearer if you wrote "... replace the funding transaction ..."

Non-Interactive-Protocol.md
```diff
@@ -0,0 +1,173 @@
...
+			- MUST reschedule timer for next bump
+		- if `contract_maturity_bound` is 20 blocks ahead from local tip:
+			- MUST spend the funding input with a higher-feerate transaction/absolute fee than any previously propagated package feerate
+			- MUST start a bumping timer of length `security_delay`
+			- MUST disable `congestion_delay` timer
+		- if `security_delay` expires:
+			- MUST spend the funding input with a higher-feerate transaction/absolute fee than any previously propagated package feerate
+			- MUST replace according to BIP 125
+			- MUST reschedule timer for next bump
+		- MUST NOT spend more in bumping fees than its collateral input value
```
Tibo-lg: I would love to see a timed automata to describe your protocol. But I can add it later (also that will help me confirm I understood correctly :) )

Non-Interactive-Protocol.md
```diff
@@ -0,0 +1,173 @@
...
+
+A node:
+	- if the funding transaction confirms:
+		* MUST track spend of the CET output
```
Tibo-lg: Not sure I get why you need to track spend of the CET output, do you mean the fund tx output?

Non-Interactive-Protocol.md
```diff
@@ -0,0 +1,173 @@
...
+		* MUST advances its off-chain state to the execution phase
+		* MUST note its change output for latter spend
```
Tibo-lg: nit: "latter" -> "later"

Non-Interactive-Protocol.md
```diff
@@ -0,0 +1,173 @@
...
+	- if the collateral double-spend confirms:
+		* MUST abort the DLC and prevent further interactions with counterparty
+
+## Rationale
+
+Funding transaction must be confirmed to consider the off-chain state initialized. As mempools
+congestion may fluctuate between fees negotiation and transaction propagation, party must adjust
+unilaterally feerate according to their confirmation preferences. A party may choose to wait
+for mempools draining before to adjust the feerate to avoid a fee waste as block height of 
+transition to execution phase doesn't matter, assuming it's before `contract_maturity_bound`.
```
Tibo-lg: I'm not sure how can they unilaterally adjust fee rate?

Non-Interactive-Protocol.md
Tibo-lg: I'm not sure I fully agree with that. In my view, as long as you don't have a confirmed fund transaction, you shouldn't consider yourself as being in the contract. Your counter-party double spending the fund tx might be a malicious behavior, but from what I understand it might also be genuine. So I'm not sure why you label that as a "security" issue.

Non-Interactive-Protocol.md
```diff
@@ -0,0 +1,173 @@
...
+		- MUST finalize the CET with a local `SIGHASH_ALL` signature
+		- MUST broadcast the CET
+		- MAY spend its outcome output with a CPFP increasing package feerate to confirm soonly
+		- if the CET is not confirmed and `contract_timeout is 20 blocks ahead of local tip:
+			- MUST rebroadcast the CET
+			- MUST spend its outcome output with a CPFP increasing package feerate to confirm soonly
+			- MUST start a bumping timer of length `security_delay`
+		- if `security_delay` expires:
+			- MUST spend the change output with a replacing CPFP increasing package feerate to confirm soonly
+			- MUST replace according to BIP 125
```
Tibo-lg: Replace what with what? Are you talking about the transaction spending the CET output?

Non-Interactive-Protocol.md
```diff
@@ -0,0 +1,173 @@
...
+	- if `contract_timeout` is reached:
+		- MUST finalize the CET with a local `SIGHASH_ALL` signature
+		- MUST broadcast the refund transaction
+		- if the refund is not confirmed after 6 blocks:
+			- MUST spend the change output with a CPFP increasing package feerate to confirm soonly
+			- MUST start a bumping timer of length `congestion_delay`
+		- if `congestion_delay` expires:
+			- MUST spend the change output with a replacing CPFP increasing package feerate to confirm soonly
+			- MUST replace according to BIP 125
+			- MUST reschedule timer for next bump
```
Tibo-lg: What's the issue with the refund not being included in 6 blocks? AFAICT if you reached the refund state you probably have time no?

Non-Interactive-Protocol.md
```diff
@@ -0,0 +1,173 @@
...
+
+A node:
+	- if a CET transaction confirms:
```
Tibo-lg: CET->refund

Tibo-lg (commented):
Looks good, apart from some minor comments. Also your formatting is weird if you could fix it it'd be great.

Non-Interactive-Protocol.md
LLFourn: Also somehow the markdown rendering of these lists if failing.

Non-Interactive-Protocol.md
LLFourn: Well, they can attempt to delay the CET and broadcast the refund transaction.

Non-Interactive-Protocol.md
```diff
@@ -0,0 +1,173 @@
...
+	- if an oracle signature is received:
+		- MUST finalize the CET with a local `SIGHASH_ALL` signature
+		- MUST broadcast the CET
+		- MAY spend its outcome output with a CPFP increasing package feerate to confirm soonly
+		- if the CET is not confirmed and `contract_timeout is 20 blocks ahead of local tip:
+			- MUST rebroadcast the CET
+			- MUST spend its outcome output with a CPFP increasing package feerate to confirm soonly
+			- MUST start a bumping timer of length `security_delay`
+		- if `security_delay` expires:
+			- MUST spend the change output with a replacing CPFP increasing package feerate to confirm soonly
```
LLFourn: soonly isn't a word I hear used often. "soon" is usually sufficient.

```suggestion
			- MUST spend the change output with a replacing CPFP increasing package feerate to shorten confirmation time
```

Non-Interactive-Protocol.md
ariard: Right we don't have a closing transaction anymore. Deleted. 

Note, a party can misbehave by a bunch of way before or after oracle signature release like time-dilating the honest party or DoSing its node until refund timeout.

Non-Interactive-Protocol.md
ariard: Right, I remove the local transaction mention. They can either not store the signature or avoid to produce them at all until it's needed. That's a recommendation not mandatory behavior as it depends on your signing setup.

I use the word local few times to mean this holder-of-the-CET-and-not-the-counterparty e.g in the context of signature. Should I upgrade local -> holder ?

Non-Interactive-Protocol.md
ariard: Talked off-line. A nice to have at some point.

Non-Interactive-Protocol.md
ariard: Actually that the CPFP which must be replaced so precised even it's verbose.

Fixed some markdown issues, lmk if that was the ones you spotted.

Non-Interactive-Protocol.md
ariard: If the funding transaction confirms and at CET maturing your counterparty broadcast and gets confirmed the CET you must detect it and mark for spend your balance output.

You can parse onchain CETs by their TXIDs by if you have a thousand of them that's more efficient to track the funding output spent.

Note, you want your onchain parser to be efficient if protocol participants are delegating their CET monitoring to some third-party entity like a watchtower which may have tens of thousands of CET to watch.

Non-Interactive-Protocol.md
ariard: Added mention of `package feerate` and CPFP. By attaching a high-fee CPFP to your funding transaction you increase feerate without cooperation of your protocol participant.

Let me know if it's clearer or need a deeper rewording.

Non-Interactive-Protocol.md
ariard: Made it clearer replacing the CPFP transaction. There is non-null risk that an ulterior rebroadcasting of higher-feerate CPFP might be rejected by the mempool if you add unconfirmed inputs.

Non-Interactive-Protocol.md
ariard: I think we agree on "you shouldn't consider yourself as being in the contract" as I made mandatory : "if the funding transaction confirms MUST advances its off-chain state to the execution phase". 

> Your counter-party double spending the fund tx might be a malicious behavior, but from what I understand it might also be genuine.

Without interaction with your counterparty how do you know if it's a malicious or genuine behavior :) ? In case of doubt, assume the worst-case scenario.

Non-Interactive-Protocol.md
ariard: Let's say an oracle is buggy or under attack. If it's key material is compromised after the CET maturation, a malicious CET can be finalized by your counterparty as long as the refund isn't confirmed.

Say differently a lazy oracle key material is toxic until the refund is confirmed.

Non-Interactive-Protocol.md
ariard: Fixed.

ariard:
Thanks @benthecarman @Tibo-lg @LLFourn for the reviews. Fixed most of your comments.

Still open discussions :

*  [holder-or-local terminology](https://github.com/discreetlogcontracts/dlcspecs/pull/87#discussion_r488308392)
*  [markdown issues](https://github.com/discreetlogcontracts/dlcspecs/pull/87#discussion_r488309139)
* [unilateral fee bumping wording](https://github.com/discreetlogcontracts/dlcspecs/pull/87#discussion_r488310808)
* [labelling bumping as a security issue](https://github.com/discreetlogcontracts/dlcspecs/pull/87#discussion_r488311483)
* [mandatory bumping of the refund](https://github.com/discreetlogcontracts/dlcspecs/pull/87#discussion_r488312319)

Added work under Creative Commons.

Non-Interactive-Protocol.md
```diff
@@ -0,0 +1,175 @@
+# Non-interactive Onchain Protocol
+
+## Abstract
+
+DLC allows two parties to conduct stateful bitcoin contracts by giving each
+of the parties *cross-signed range of CET transactions*. These CET transactions
```
nkohen: nit: `CET transactions` -> `CETs`

Non-Interactive-Protocol.md
```diff
@@ -0,0 +1,175 @@
...
+are all spending the same output from a funding transaction and thus only one of
+the alternative can confirm. A CET is finalized according to the outcome signed
```
nkohen: nit: only one CET can confirm

Non-Interactive-Protocol.md
```diff
@@ -0,0 +1,175 @@
...
+by the previously selected oracle.
+
+As the contract is established once funding transaction signatures have been exchanged,
+from then a party should assume counterparty non-cooperation and ensure either
+confirmation of the funding or double-spend of its collatral inputs to avoid exploitation.
+
+In the optimistic way, once the oracle signature is released, parties interact
+off-chain to settle the DLC.
```
nkohen: I thought that now that we have adaptor sigs there's no more difference between co-operative and unilateral?

Non-Interactive-Protocol.md
```diff
@@ -0,0 +1,175 @@
...
+
+In the pessimitic way, a party interact with the chain to unilaterally settle the DLC.
+They are two scenarios to consider:
+
+1.  The oracle releases the outcome signature but the DLC counterparty goes wrong,
+either going offline or maliciously behaving. The party must broadcast the DLC
```
nkohen: I don't think we require counter-party for closing?

Non-Interactive-Protocol.md
```diff
@@ -0,0 +1,175 @@
...
+
+2. The oracle doesn't release the outcome signature, independently of counterparty
+behavior, the party must broadcast the refund transaction after its timelock expiration.
+
+# Table of Contents
+	* [General Requirements](#general-requirements)
+	* [Fee-Bumping Policy](#fee-bumping-policy)
+	* [Funding Phase](#funding-phase)
+	* [Execution Phase](#execution-phase)
+	* [Reorgs](#reorgs)
```
nkohen: nit: i would personally prefer this be above the abstract section and include the abstract as the first point. Up to you though

Non-Interactive-Protocol.md
```diff
@@ -0,0 +1,175 @@
...
+
+# General Requirements
+
+A DLC client should have censorship-resistant access to the blockchain, use a
+local fee-estimator and ensure it's well-connected to the tx-relay network.
+
+We recommend to not react on any mempool event as otherwise it would be a source of
+cross-layer mapping.
```
nkohen: Could you define or link out to what cross-layer mapping means?

Non-Interactive-Protocol.md
```diff
@@ -0,0 +1,175 @@
...
+
+We recommend to not sign transactions until their broadcast is required
+to avoid their misuage in case of site compromise, assuming an external signer.
+
+# Fee bumping policy
+
+Client's `congestion_delay`/`security_delay` can be scheduled on three different basis:
```
nkohen: It would be nice to start with what these delays are and how they are used before defining values and potential values for them

Non-Interactive-Protocol.md
```diff
@@ -0,0 +1,175 @@
...
+- a block height
+- a local clock
+- an observed feerate fluctuation
+
+We don't provide recommendations for the congestion delay as its settings depends of user
+liquidity preferences and available local data to predict the feerate.
+
+As `security_delay` must be offensive, we recommend to set it to 1 block.
```
nkohen: Not sure what you mean by offensive but this might be cleared up by a description of what `security_delay` is from above ^

Non-Interactive-Protocol.md
```diff
@@ -0,0 +1,175 @@
...
+# Funding Phase
+
+A party always starts the protocol in the funding phase. The funding phase can
+terminate by either a transition to the execution phase or a protocol abort.
+
+## Requirements
+
+A node:
+	- if a counterparty's funding signature is received:
+		- MUST finalize the funding transaction with a local `SIGHASH_ALL` signature
```
nkohen: nit: MUST finalized the funding transaction with local signatures using `SIGHASH_ALL`
(the key thing being plural as they may have more than one signature)

Non-Interactive-Protocol.md
```diff
@@ -0,0 +1,175 @@
...
+		- MUST broadcast the funding transaction
+
+A node:
+	- if a funding signature has been sent or a funding transaction broadcast and there is no confirmation after 6 blocks:
+		- MUST spend the change output with a CPFP increasing package feerate to shorten confirmation time
+		- MUST start a bumping timer of length `congestion_delay`
```
nkohen: I'm not sure what is meant by the word "bumping" but I think that this sentence is more clear without it, also it should be made clear above when defining `congestion_delay` that it is a relative delay (as opposed to an absolute timelock)

Non-Interactive-Protocol.md
```diff
@@ -0,0 +1,175 @@
...
+		- if `congestion_delay` expires:
+			- MUST spend the change output with a CPFP increasing package feerate to shorten confirmation time
+			- MUST replace the CPFP transaction according to BIP 125
+			- MUST reschedule timer for next bump
+		- if `contract_maturity_bound` is 20 blocks ahead from local tip:
+			- MUST spend the funding input with a higher-feerate transaction/absolute fee than any previously propagated package feerate
+			- MUST start a bumping timer of length `security_delay`
```
nkohen: Same comment here

Non-Interactive-Protocol.md
```diff
@@ -0,0 +1,175 @@
...
+
+A node:
+	- if a funding signature has been sent or a funding transaction broadcast and there is no confirmation after 6 blocks:
+		- MUST spend the change output with a CPFP increasing package feerate to shorten confirmation time
+		- MUST start a bumping timer of length `congestion_delay`
+		- if `congestion_delay` expires:
+			- MUST spend the change output with a CPFP increasing package feerate to shorten confirmation time
+			- MUST replace the CPFP transaction according to BIP 125
+			- MUST reschedule timer for next bump
+		- if `contract_maturity_bound` is 20 blocks ahead from local tip:
```
nkohen: 20 blocks seems quite short considering you could have year-long contracts which aren't getting confirmed. Perhaps we should also have something like
```
if there is no confirmation of the funding transaction after at least ??? blocks:
    - MAY spend a funding input with a higher feerate ...
```

Non-Interactive-Protocol.md
```diff
@@ -0,0 +1,175 @@
...
+			- MUST spend the funding input with a higher-feerate transaction/absolute fee than any previously propagated package feerate
+			- MUST start a bumping timer of length `security_delay`
+			- MUST disable `congestion_delay` timer
+		- if `security_delay` expires:
+			- MUST spend the funding input with a higher-feerate transaction/absolute fee than any previously propagated package feerate
```
nkohen: nit: `spend a funding input` (there may be more than one)

Non-Interactive-Protocol.md
```diff
@@ -0,0 +1,175 @@
...
+			- MUST replace the CPFP transaction according to BIP 125
+			- MUST reschedule timer for next bump
+		- MUST NOT spend more in bumping fees than its collateral input value
+
+A node:
+	- if the funding transaction confirms:
+		- MUST track spend of the CET output
```
nkohen: isn't it the funding output they need to track? Perhaps that's what you mean and are using different terminology? (CET output would be the final payout whereas funding output is the output spent by CETs)

Non-Interactive-Protocol.md
```diff
@@ -0,0 +1,175 @@
...
+		- MUST advances its off-chain state to the execution phase
+		- MUST note its change output for later spend
+
+A node:
+	- if the collateral double-spend confirms:
```
nkohen: nit: `collateral -> funding input` just so that we use consistent terminology

Non-Interactive-Protocol.md
```diff
@@ -0,0 +1,175 @@
...
+# Execution Phase
+
+The execution phase transitions to the terminal phase either by a CET or a refund
+transaction confirmations.
+
+## Requirements
+
+A node:
+	- if an oracle signature is received:
+		- MUST finalize the CET with a local `SIGHASH_ALL` signature
```
nkohen: and a completed remote adaptor signature

Non-Interactive-Protocol.md
```diff
@@ -0,0 +1,175 @@
...
+		- MUST broadcast the CET
+		- MAY spend its outcome output with a CPFP increasing package feerate to shorten confirmation time
+		- if the CET is not confirmed and `contract_timeout` is 20 blocks ahead of local tip:
+			- MUST rebroadcast the CET
+			- MUST spend its outcome output with a CPFP increasing package feerate to shorten confirmation time
+			- MUST start a bumping timer of length `security_delay`
```
nkohen: nit: bumping

Non-Interactive-Protocol.md
```diff
@@ -0,0 +1,175 @@
...
+			- MUST reschedule timer for next bump
+		- MUST NOT spend more in bumping fees than its outcome value
+
+A node:
+	- if a CET transaction confirms:
+		- MUST note its outcome output for later spend
+		- MUST advances its off-chain state to terminal phase
+
+A node:
+	- if `contract_timeout` is reached:
```
nkohen: if `contract_timeout` is reached and no oracle signature has been received

I think this is an important change because we want the above section to supersede this one I think?

nkohen: Especially if you're in the process of applying BIP 125 to your CPFP tx spending the CET

Non-Interactive-Protocol.md
```diff
@@ -0,0 +1,175 @@
...
+		- MUST finalize the refund transaction with a local `SIGHASH_ALL` signature
+		- MUST broadcast the refund transaction
+		- if the refund is not confirmed after 6 blocks:
+			- MUST spend the change output with a CPFP increasing package feerate to shorten confirmation time
+			- MUST start a bumping timer of length `congestion_delay`
```
nkohen: nit: bumping

nkohen (commented):
Looking good :) Just lots of nits and a couple suggestions and concerns

Non-Interactive-Protocol.md
nkohen: Perhaps it should go from `MUST` to `SHOULD` as it is up to the client how much they trust the oracle? I guess I don't have strong opinions on this though, also it does seem like this is at least a little less high-priority than uses of `congestion_delay` above, perhaps this should be allowed to be a different less strict delay to save on fees?

Non-Interactive-Protocol.md
```diff
@@ -0,0 +1,175 @@
+# Non-interactive Onchain Protocol
```
nkohen: nit: `Non-Interactive` (capital I)

Non-Interactive-Protocol.md
Tibo-lg: Also it doesn't seem to display properly (maybe it's just me?)

Christewart:
It doesn't display right for me either

Non-Interactive-Protocol.md
Tibo-lg: No I think local is fine if you're talking about signature.

Non-Interactive-Protocol.md
Tibo-lg: Display is still weird (maybe even worse than before): https://github.com/ariard/dlcspecs/blob/2020-09-non-interactive-protocol/Non-Interactive-Protocol.md

Non-Interactive-Protocol.md
ariard: Good call, reworded this part a bit, lmk.

Non-Interactive-Protocol.md
ariard: See above, fixed.

Non-Interactive-Protocol.md
ariard: Took same indentation that other files. And included abstract.

Non-Interactive-Protocol.md
ariard: Sure, see this nice paper research : https://arxiv.org/pdf/2007.00764.pdf. Added a footnote.

Non-Interactive-Protocol.md
ariard: As of initial commit 7977904, this section was at the end after delays have been defined and used in previous section. IIRC Ben or TIbo made me the remark it would be clearer first but can't find comment in Github mess anymore. Reverted.

Non-Interactive-Protocol.md
ariard: In fact I forget a Rationale for the Execution phase added one.

Note, it's a MUST because ultimately you should manage refund transaction confirmation but `congestion_delay` and feerate are user settings so if you want to spend on fees just pick up values which suits you like `congestion_delay`=100. This was the intention of laying out a Fee bumping policy, let me know if I can make it clearer.

Non-Interactive-Protocol.md
ariard: Rewrote this part.

Non-Interactive-Protocol.md
ariard: How you can pick up a `congestion_delay` is advised in fee bumping policy. Added the prefix "fee-" to bumping. I think the expression fee-bumping is used by the LN ecosystem to designate "automatic increase of the feerate of time-sensitive txn after a delay" though likely not consistently beyond RL :)

Non-Interactive-Protocol.md
ariard: Just to be sure `contract_maturity_bound` is more or less the absolute height at which the oracle secret will be released ? This is the reference point which matters.

Non-Interactive-Protocol.md
ariard: Ah, we need to agree on the terminology. It's a funding tx output which commits to a CET redeemscript spends by a CET transaction. Do you have a suggestion ?

Non-Interactive-Protocol.md
ariard: Only one of the party is tweaking its key right ? Maybe it would be better to do a returns to the part of the spec detailing the witness ?

Non-Interactive-Protocol.md
ariard: Added the "no oracle signature..." requirement. How can you apply a CPFP to your CET if you don't have an oracle signature ?

ariard:
Thanks @nkohen for review, updated at 8b4ad92

Added a Rationale for the Execution phase and extended the Fee-Bumping Policy section, also updated with more minor comments.

I think we still need to agree around some wording and terminology.

See non-solved:
* [order of refund/CET reactions](https://github.com/discreetlogcontracts/dlcspecs/pull/87#discussion_r500371803)
* [CET finalization wording](https://github.com/discreetlogcontracts/dlcspecs/pull/87#discussion_r500369127)
* [CET output vs CET tx](https://github.com/discreetlogcontracts/dlcspecs/pull/87#discussion_r500366164)
* [20-blocks security timer starting point](https://github.com/discreetlogcontracts/dlcspecs/pull/87#discussion_r500364672)
* [`congestion_delay` semantic](https://github.com/discreetlogcontracts/dlcspecs/pull/87#discussion_r500379919)

And still markdown ?

Tibo-lg:
> And still markdown ?

Yeah you can just look at how it renders: https://github.com/ariard/dlcspecs/blob/2020-09-non-interactive-protocol/Non-Interactive-Protocol.md

Non-Interactive-Protocol.md
```diff
@@ -0,0 +1,195 @@
...
+# General Requirements
+
+A DLC client should have censorship-resistant access to the blockchain, use a
+local fee-estimator and ensure it's well-connected to the tx-relay network.
+
+We recommend to not react on any mempool event as otherwise it would be a source of
+cross-layer mapping. <sup>[Cross-layer Mapping](https://arxiv.org/pdf/2007.00764.pdf)</sup>
+
+We recommend to not sign transactions until their broadcast is required
+to avoid their misuage in case of site compromise, assuming an external signer.
```
nkohen: nit: Perhaps clarify that this is for non-adaptor signatures

Non-Interactive-Protocol.md
```diff
@@ -0,0 +1,195 @@
...
+# Funding Phase
+
+A party always starts the protocol in the funding phase. The funding phase can
+terminate by either a transition to the execution phase or a protocol abort.
+
+## Requirements
+
+A node:
+	- if a counterparty's funding signature is received:
+		- MUST finalize the funding transaction with a local `SIGHASH_ALL` signature
```
nkohen: Here and below funding signature -> funding signatures (as there can be many inputs with many sigs)

Non-Interactive-Protocol.md
nkohen: but the problem here (especially with long-term events) is that `contract_maturity_bound` is not strictly speaking enforceable, that is to say nothing is stopping the oracle from broadcasting a signature earlier than this, and in fact this can be expected if hash rate goes in such a way that the expected block height of a future date is way off.

Non-Interactive-Protocol.md
nkohen: I would say there is a funding tx output spent by a CET, and I wouldn't attach ownership of the redeem script to CETs as they all share so that seems to be in the domain of the funding transaction/funding output still. But I think regardless of this minute difference the wording in the spec is a little confusing as the thing a client must track to see if it is spent is the funding output not the CET (just like in LN you track the funding output)

Non-Interactive-Protocol.md
```diff
@@ -0,0 +1,195 @@
...
+# Execution Phase
+
+The execution phase transitions to the terminal phase either by a CET or a refund
+transaction confirmations.
+
+## Requirements
+
+A node:
+	- if an oracle signature is received:
+		- MUST finalize the CET with a local `SIGHASH_ALL` signatures
```
nkohen: What are your thoughts on using local and remote adaptor signatures for this purpose as an alternative to using private keys to sign at claim time? I think it might be nice to mention that this is an option

Non-Interactive-Protocol.md
nkohen: also to be completely clear, unlike contract_timeout, contract_maturity_bound is not guaranteed to be enforced by CET timelocks (along with oracles having the option of signing before then)

Non-Interactive-Protocol.md
```diff
@@ -0,0 +1,195 @@
...
+		- MUST spend the change output with a CPFP increasing package feerate to shorten confirmation time
+		- MUST start a fee-bumping timer of length `congestion_delay`
+		- if `congestion_delay` expires:
+			- MUST spend the change output with a CPFP increasing package feerate to shorten confirmation time
+			- MUST replace the CPFP transaction according to BIP 125
+			- MUST reschedule timer for next bump
+		- if `contract_maturity_bound` is 20 blocks ahead from local tip:
+			- MUST spend a funding input with a higher-feerate transaction/absolute fee than any previously propagated package feerate
+			- MUST start a fee-bumping timer of length `security_delay`
+			- MUST disable `congestion_delay` timer
```
nkohen: security_delay and congestion_delay should probably be defined above with guidance given on their values (e.g. security_delay has to be really short as you only have 20 blocks)

Non-Interactive-Protocol.md
nkohen: And also this isn't true for the refund tx which I believe is signed during setup

Non-Interactive-Protocol.md
nkohen: Ah I see now that this happens below, fee bumping policy section should probably move above as these are undefined terms at this point in the spec

Non-Interactive-Protocol.md
ariard: Updated to precise that local signature production, which is different from the remote signature exchanged at contract initialization. 

Also, added a new recommendation on duplicate backups based on an offline conv with Chris.

Non-Interactive-Protocol.md
ariard: Note that initial design goal of specifying a `contract_maturity_bound` is to have a clear schelling point at which to start an implementation must start to broadcast and fee-bump double-spend of its broadcast. Otherwise I fear implementor to pickup arbitrary, too late delays to achieve security. The "grace delay" of 20 blocks is likely a bit arbitrary and maybe should be higher but I think it's worthy to have this emergency mechanism well-implemented.

Modified wording to "at least 20 blocks", giving more room for implementations.

I agree that oracle secret releasing at `contract_maturity_bound` is purely a gentleman agreement but it would provide a blockchain-enforced protection in case of oracle compromise. An attacker signing arbitrary states and accessing counterparty key material can't confirm malicious CET before `contract_maturity_bound`. Thus letting the oracle accessing the CET cancellation key (#94).

It may also be used in the future as some kind of oracle misbehavior "proof". You combine a CET with a `contract_maturity_bound` + an oracle secret released that you can timestamp in the blockchain before reaching `contract_maturity_bound`.

> and in fact this can be expected if hash rate goes in such a way that the expected block height of a future date is way off.

I think we can extend #94-style of protection with some native block delta aware script (playing with nLocktime double-semantic of height/epoch) to cancel CET if the block height is beyond any expectation.

Non-Interactive-Protocol.md
ariard: Good to me, updated.

Non-Interactive-Protocol.md
ariard: Do you a formalization somewhere of what you're proposing ? Sounds good to me, we can still modify this later when we have something more fleshed out.

Non-Interactive-Protocol.md
ariard: Yes I know but Tibo made the other point when the section come first, they were defined but their usage wasn't explained . Added a line about text organization to avoid user confusion.

ariard:
Updated, thanks @nkohen!

Ongoing discussion points :
* [presence and semantic of `contract_maturity_bound`](https://github.com/discreetlogcontracts/dlcspecs/pull/87#discussion_r506957146)

I think markdown is fixed but let me know.

Non-Interactive-Protocol.md
nkohen: Essentially the idea is to store the adaptor sigs you give your counter party along with the ones they give you so that you don't need to access keys at claim time

Non-Interactive-Protocol.md
```diff
@@ -0,0 +1,212 @@
...
+A node:
+
+  - if a funding signature has been sent or a funding transaction broadcast and there is no confirmation after 6 blocks:
+    - MUST spend the change output with a CPFP increasing package feerate to shorten confirmation time
+    - MUST start a fee-bumping timer of length `congestion_delay`
+    - if `congestion_delay` expires:
+      - MUST spend the change output with a CPFP increasing package feerate to shorten confirmation time
+      - MUST replace the CPFP transaction according to BIP 125
+      - MUST reschedule timer for next bump
+    - if `contract_maturity_bound` is at least 20 blocks ahead from local tip:
```
nkohen: If you want a suggestion I would be more comfortable with something like `(contract_maturity_bound + signing_time)/2`. I really think that using something like 20 provides absolutely no value here because I would expect most `contract_maturity_bound`s in the world to have a margin of error well over 20 blocks.

nkohen: And on a separate note, I am also concerned about what this could mean for events which an oracle will either sign at any time present to contract maturity (such as an event like "will btc price exceed 20k before end of year" which could generate a "yes" signature at any time but will only generate a "no" signature at new year). In this case it seems like some number of blocks after signing (or worse, if a price movement is happening) then parties will want to enter this state

nkohen: And also, another issue with using `contract_maturity_bound` as closely as 20 is that I see virtually no difference in the incentive to double-spend a long-standing DLC 20 blocks before contract maturity and at contract maturity, either way the counter-party likely has near certainty about what the outcome is going to be which they did not have when they agreed to the DLC

Non-Interactive-Protocol.md
```diff
@@ -0,0 +1,212 @@
...
+
+  - if the funding transaction confirms:
+    - MUST track spend of the funding output
+    - MUST advances its off-chain state to the execution phase
+    - MUST note its change output for later spend
+
+A node:
+
+  - if the funding input double-spend confirms:
+    - MUST abort the DLC and prevent further interactions with counterparty
```
nkohen, line 139: I think this deserves a rationale, as looking back at this I don't remember why this is necessary

Non-Interactive-Protocol.md
```diff
@@ -0,0 +1,212 @@
...
+
+The execution phase transitions to the terminal phase either by a CET or a refund
+transaction confirmations.
+
+## Requirements
+
+A node:
+
+  - if an oracle signature is received:
+    - MUST finalize the CET with a local `SIGHASH_ALL` signature
```
nkohen, line 170: See https://github.com/discreetlogcontracts/dlcspecs/pull/87#discussion_r514905208

Non-Interactive-Protocol.md
```diff
@@ -0,0 +1,212 @@
...
+    - MAY spend its outcome output with a CPFP increasing package feerate to shorten confirmation time
+    - if the CET is not confirmed and `contract_timeout` is 20 blocks ahead of local tip:
+      - MUST rebroadcast the CET
+      - MUST spend its outcome output with a CPFP increasing package feerate to shorten confirmation time
+      - MUST start a fee-bumping timer of length `security_delay`
+    - if `security_delay` expires:
+      - MUST spend the change output with a replacing CPFP increasing package feerate to shorten confirmation time
+      - MUST replace the CPFP transaction according to BIP 125
+      - MUST reschedule timer for next bump
+    - MUST NOT spend more in bumping fees than its outcome value
```
nkohen, line 181: Perhaps this requirement should be made more tight to say that you shouldn't spend more than `outcome_value - refund_value`?

Non-Interactive-Protocol.md
```diff
@@ -0,0 +1,212 @@
...
+A node:
+
+  - if a CET transaction confirms:
+    - MUST note its outcome output for later spend
+    - MUST advances its off-chain state to terminal phase
+
+A node:
+
+  - if no oracle signature has been received and `contract_timeout` is reached:
+    - MUST finalize the refund transaction with a local `SIGHASH_ALL` signature
```
nkohen, line 195: same comment here as above where they could be told to store this sig when they send it to the coutner-party alongside the one they got from counter-party so as to make key access not needed

Non-Interactive-Protocol.md
```diff
@@ -0,0 +1,212 @@
...
+    - MUST broadcast the refund transaction
+    - if the refund is not confirmed after `congestion_delay`:
+      - MUST spend the change output with a CPFP increasing package feerate to shorten confirmation time
+      - MUST start a fee-bumping timer of length `congestion_delay`
+    - if `congestion_delay` expires:
```
nkohen: I'm not in favor of re-using the `congestion_delay` that is used for the funding transaction, I think that this should be explicitly called `refund_congestion_delay` as it will likely be a much larger number than the funding's `congestion_delay` because in the funding case you are worried about information around odds of outcomes changing incentivizing counter-party to double-spend inputs, whereas here in the refund case the only worry is that the oracle could be or become compromised which is likely less probable than information changing.

nkohen (commented):
I think we're nearly there. I just have a couple remaining concerns :)

benthecarman (approved):

Non-Interactive-Protocol.md
ariard: Left a rationale. Lmk what you think. IMO best practice to clean off-chain state machine early to avoid further miscs/exploitations (CVE-2020-26896 class of bugs)

Non-Interactive-Protocol.md
ariard: Replied there.

Non-Interactive-Protocol.md
ariard: Ah I see. I think it's a trade-off storing sigs is less constrainin, but they might be maliciously used in case of compromise. Instead of accessing key material twice. 

May we defer this to a future Recommendations for Key Management, which should be elsewhere than a primary onchain-stuff specification, IMO ?

Non-Interactive-Protocol.md
ariard: Because you assume worst-case scenario your counterparty confirms the refund and your lose only the `outcome_value - refund_value` diff ? I assume the worst-case is a oracle key compromise signing any state thus the value at risk is the full outcome value.

Non-Interactive-Protocol.md
ariard: Resolved, replied there.

Non-Interactive-Protocol.md
ariard: > And also, another issue with using contract_maturity_bound as closely as 20 is that I see virtually no difference in the incentive to double-spend a long-standing DLC 20 blocks before contract maturity

In this case, it's not about counterparty incentive, but your block delay expectation to confirm your double-spend. I was thinking 3 blocks (fast-confirmation) plus 17 of security margin, based on roughly what we're using to accelerate bumping timer in RL.

But your point about anytime-signing event is a really good one, and added with the general observation that we don't have a way to make signing time a block-or-epoch consensus-enforced fixed point, we should defer more to clients.

What about introducing a `security_point` to which the evaluation is left to client, with recommendations on how to define it according to event types (anytime-signing, long-standing, ...) but still with a block confirmation safety buffer (the current 20 blocks) and a fee-bumping timer requirement ?

I still aim to separate cleanly congestion from security confirmation strategy, as you don't want a really bad mempool spike suddenly turning DLC client fee-bumpers in mad cows, and overpaying in fees.

Non-Interactive-Protocol.md
ariard: See my reply on `contract_maturity_bound` first. I think a congestion fee-bumping strategy must be different from a security fee-bumping one to prevent mempool spikes provoking massive and useless fees burning for DLC currently in funding phase. I might be to much cautious here, but that's something we have to care about. At least, until we move off-chain where funding and its congestion management isn't blurred with DLC setting up (caveat with zero-conf channels).

So minding `congestion_delay` isn't for odds of outcomes incentivizing counter-party to double-spend inputs, but still conceding you the oracle compromise risk for refund is a different risk, we may a) introduce  a `refund_security_delay` to specifically address the risk of oracle compromise or b) replace directly `congestion_delay` by a security one to bound implementation complexity.

ariard:
Thanks for the review, still few important points to correct, see my proposals for either:

* [introducing a `security_points` to avoid signing time uncertainty](https://github.com/discreetlogcontracts/dlcspecs/pull/87#discussion_r515443672)
* [rational for cleaning offchain state machine early](https://github.com/discreetlogcontracts/dlcspecs/pull/87#discussion_r515440655)
* [deference for future key management recommendations](https://github.com/discreetlogcontracts/dlcspecs/pull/87#discussion_r515441562)
* [fee-bumping upper bound](https://github.com/discreetlogcontracts/dlcspecs/pull/87#discussion_r515442071)
* [dissociate security/congestion delays for refund](https://github.com/discreetlogcontracts/dlcspecs/pull/87#discussion_r515445122)

ariard:
Updated at b7c35d5, see #dlc LDK slack 11/03 for resolution of pending discussion points.

PR #105, opened 2020-10-07
Oracle Announcement TLVs

benthecarman:
Initial proposal for oracle announcement TLVs.

The idea here is that the oracle signs a hash of an `oracle_event` TLV.
After doing so, peers that already trust the oracle's key should be able to pass around `oracle_announcement` messages to learn of events.

I imagine the spec for if the oracle signs the sha256 vs tagged sha256 of the `oracle_event` TLV will go in the actual `Oracle.md` spec

### DIFF
--- a/Messaging.md
+++ b/Messaging.md
@@ -23,6 +23,13 @@ All data fields are unsigned big-endian unless otherwise specified.
       * [Version 0 `cet_adaptor_signatures`](#version-0-cet_adaptor_signatures)
     * [The `funding_signatures` Type](#the-funding_signatures-type)
       * [Version 0 `funding_signatures`](#version-0-funding_signatures)
+    * [The `event_descriptor` Type](#the-event_descriptor-type)
+      * [Version 0 `external_event_descriptor`](#version-0-external_event_descriptor)
+      * [Version 0 `enum_event_descriptor`](#version-0-enum_event_descriptor)
+    * [The `oracle_event` Type](#the-oracle_event-type)
+      * [Version 0 `oracle_event`](#version-0-oracle_event)
+    * [The `oracle_announcement` Type](#the-oracle_announcement-type)
+      * [Version 0 `oracle_announcement`](#version-0-oracle_announcement)
 * [Authors](#authors)
 
 ## Connection Handling and Multiplexing
@@ -69,6 +76,7 @@ The following convenience types are also defined:
 * `script_sig`: A bitcoin script signature encoded as ASM prefixed a `u16` value indicating its length.
 * `short_contract_id`: an 8 byte value identifying a contract funding transaction on-chain (see [BOLT #7](https://github.com/lightningnetwork/lightning-rfc/blob/master/07-routing-gossip.md#definition-of-short-channel-id))
 * `bigsize`: a variable-length, unsigned integer similar to Bitcoin's CompactSize encoding, but big-endian.  Described in [BigSize](https://github.com/lightningnetwork/lightning-rfc/blob/master/01-messaging.md#appendix-a-bigsize-test-vectors).
+* `utf8_string`: a UTF-8 encoded string
 
 ## DLC Specific Types
 
@@ -165,9 +173,68 @@ This type contains signatures of the funding transaction and any necessary infor
 `witness` is the data for a witness element in a witness stack. An empty `witness_stack` is an error,
 as every input must be Segwit. Witness elements should *not* include their length as part of the witness data.
 
+### The `event_descriptor` Type
+
+This type contains information about the outcomes in an event for which an oracle plans on releasing a signature over.
+
+#### Version 0 `external_event_descriptor`
+
+1. type: 55300 (`external_event_descriptor_v0`)
+2. data:
+   * [`utf8_string`:`string`]
+
+`string` can refer to anything here and it is up to the oracle and user to agree on how to interpret it.
+
+#### Version 0 `enum_event_descriptor`
+
+1. type: 55302 (`enum_event_descriptor_v0`)
+2. data:
+   * [`u16`:`outcome_1_len`]
+   * [`utf8_string`:`outcome_1`]
+   * ...
+   * [`u16`:`outcome_n_len`]
+   * [`utf8_string`:`outcome_n`]
+
+This type of event descriptor is a simple enumeration where the value `n` is omitted from being explicitly included as it can be derived from the length field of the TLV.
+
+Each `outcome_n` corresponds to the pre-image of a possible outcome that the oracle could sign.
+
+### The `oracle_event` Type
+
+This type contains information about an event for which an oracle plans on releasing a signature over.
+
+#### Version 0 `oracle_event`
+
+1. type: 55330 (`oracle_event_v0`)
+2. data:
+   * [`x_point`:`oracle_public_key`]
+   * [`x_point`:`oracle_nonce`]
+   * [`u32`:`event_maturity_epoch`]
+   * [`event_descriptor`:`event_descriptor`]
+   * [`utf8_string`:`event_uri`]
+
+`event_maturity_epoch` refers to the earliest time this event is expected to be signed, in epoch seconds.
+
+`event_uri` is a name and/or categorization of this event given by the oracle.
+
+### The `oracle_announcement` Type
+
+This type contains information about an announcement of an oracle to attest to an event in the future.
+
+#### Version 0 `oracle_announcement`
+
+1. type: 55332 (`oracle_announcement`)
+2. data:
+   * [`signature`:`annoucement_signature`]
+   * [`oracle_event`:`oracle_event`]
+
+The `annoucement_signature` is a signature of the hash of the serialized `oracle_event` using the `oracle_public_key`.
+
 ## Authors
 
 Nadav Kohen <nadavk25@gmail.com>
+
+Ben Carman <benthecarman@live.com>
 
 ![Creative Commons License](https://i.creativecommons.org/l/by/4.0/88x31.png "License CC-BY")
 <br>
--- a/Messaging.md
+++ b/Messaging.md
@@ -191,13 +191,14 @@ This type contains information about the *exact* and fully specified outcomes  i
 
 1. type: 55302 (`enum_event_descriptor_v0`)
 2. data:
+   * [`u16`:`num_outcomes`]
    * [`u16`:`outcome_1_len`]
    * [`string`:`outcome_1`]
    * ...
    * [`u16`:`outcome_n_len`]
    * [`string`:`outcome_n`]
 
-This type of event descriptor is a simple enumeration where the value `n` is the number of outcomes in the event. `n` is omitted because it can be deduced by parsing each outcome until reaching the end of the TLV.
+This type of event descriptor is a simple enumeration where the value `n` is the number of outcomes in the event.
 
 Each `outcome_i` corresponds to the pre-image of a possible outcome that the oracle could sign.
 
--- a/Messaging.md
+++ b/Messaging.md
@@ -190,12 +190,12 @@ This type contains information about the *exact* and fully specified outcomes  i
 1. type: 55302 (`enum_event_descriptor_v0`)
 2. data:
    * [`u16`:`outcome_1_len`]
-   * [`utf8_string`:`outcome_1`]
+   * [`string`:`outcome_1`]
    * ...
    * [`u16`:`outcome_n_len`]
-   * [`utf8_string`:`outcome_n`]
+   * [`string`:`outcome_n`]
 
-This type of event descriptor is a simple enumeration where the value `n` is the number of outcomes in the event. `n` is not given because it can be found by parsing each outcome's length followed by the actual outcome until you has parsed the same number of bytes as the length field of the TLV.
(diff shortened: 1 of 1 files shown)


Messaging.md
```diff
@@ -69,6 +76,7 @@ The following convenience types are also defined:
 * `script_sig`: A bitcoin script signature encoded as ASM prefixed a `u16` value indicating its length.
 * `short_contract_id`: an 8 byte value identifying a contract funding transaction on-chain (see [BOLT #7](https://github.com/lightningnetwork/lightning-rfc/blob/master/07-routing-gossip.md#definition-of-short-channel-id))
 * `bigsize`: a variable-length, unsigned integer similar to Bitcoin's CompactSize encoding, but big-endian.  Described in [BigSize](https://github.com/lightningnetwork/lightning-rfc/blob/master/01-messaging.md#appendix-a-bigsize-test-vectors).
+* `utf8_string`: a UTF-8 encoded string
```
nkohen: using NFC for normalization (with a link)

nkohen: also I would drop `utf8_` and just call it `string`

Messaging.md
```diff
@@ -165,10 +173,69 @@ This type contains signatures of the funding transaction and any necessary infor
...
+
+This type contains information about the outcomes in an event for which an oracle plans on releasing a signature over.
+
+#### Version 0 `external_event_descriptor`
+
+1. type: 55300 (`external_event_descriptor_v0`)
+2. data:
+   * [`utf8_string`:`string`]
+
+`string` can refer to anything here and it is up to the oracle and user to agree on how to interpret it.
```
nkohen: string -> external_name

Messaging.md
```diff
@@ -165,10 +173,69 @@ This type contains signatures of the funding transaction and any necessary infor
 `witness` is the data for a witness element in a witness stack. An empty `witness_stack` is an error,
 as every input must be Segwit. Witness elements should *not* include their length as part of the witness data.
 
+### The `event_descriptor` Type
+
+This type contains information about the outcomes in an event for which an oracle plans on releasing a signature over.
```
nkohen: about the _exact_ and fully specified outcomes

Messaging.md
```diff
@@ -165,10 +173,69 @@ This type contains signatures of the funding transaction and any necessary infor
...
+
+1. type: 55302 (`enum_event_descriptor_v0`)
+2. data:
+   * [`u16`:`outcome_1_len`]
+   * [`utf8_string`:`outcome_1`]
+   * ...
+   * [`u16`:`outcome_n_len`]
+   * [`utf8_string`:`outcome_n`]
+
+This type of event descriptor is a simple enumeration where the value `n` is omitted from being explicitly included as it can be derived from the length field of the TLV.
```
nkohen: Actually on this occasion `n`  should probably be included because length can't be derived extrinsically due to variable internal length. (It was included in funding_signatures above, for example)

Messaging.md
```diff
@@ -165,10 +173,69 @@ This type contains signatures of the funding transaction and any necessary infor
...
+
+Each `outcome_n` corresponds to the pre-image of a possible outcome that the oracle could sign.
```
nkohen: don't use `n` here as the meaning is taken

Messaging.md
```diff
@@ -165,10 +173,69 @@ This type contains signatures of the funding transaction and any necessary infor
...
+
+1. type: 55330 (`oracle_event_v0`)
+2. data:
+   * [`x_point`:`oracle_public_key`]
+   * [`x_point`:`oracle_nonce`]
+   * [`u32`:`event_maturity_epoch`]
+   * [`event_descriptor`:`event_descriptor`]
+   * [`utf8_string`:`event_uri`]
+
+`event_maturity_epoch` refers to the earliest time this event is expected to be signed, in epoch seconds.
```
nkohen: UTC time zone

Messaging.md
```diff
@@ -165,10 +173,69 @@ This type contains signatures of the funding transaction and any necessary infor
...
+This type contains information about an announcement of an oracle to attest to an event in the future.
+
+#### Version 0 `oracle_announcement`
+
+1. type: 55332 (`oracle_announcement`)
+2. data:
+   * [`signature`:`annoucement_signature`]
+   * [`oracle_event`:`oracle_event`]
+
+The `annoucement_signature` is a signature of the hash of the serialized `oracle_event` using the `oracle_public_key`.
```
nkohen: Should probably mention that this is the blob to copy/paste

Messaging.md
benthecarman: We still shouldn't need `n` because of the `Length` in the TLV. However I will clarify that in this line

nkohen: I don't understand how length in the TLV is helpful in determining `n`?

nkohen: and re your clarification you cannot just parse the individual lengths because that requires going through the whole blob (you don't know where the lengths are going to be a priori)

Messaging.md
```diff
@@ -165,10 +173,69 @@ This type contains signatures of the funding transaction and any necessary infor
...
+   * [`string`:`external_name`]
+
+`external_name` can refer to anything here and it is up to the oracle and user to agree on how to interpret it.
+
+#### Version 0 `enum_event_descriptor`
+
+1. type: 55302 (`enum_event_descriptor_v0`)
+2. data:
+   * [`u16`:`outcome_1_len`]
+   * [`utf8_string`:`outcome_1`]
```
Tibo-lg: Shouldn't that be the `string` type you defined?

Messaging.md
```diff
@@ -165,10 +173,69 @@ This type contains signatures of the funding transaction and any necessary infor
...
+   * ...
+   * [`u16`:`outcome_n_len`]
+   * [`utf8_string`:`outcome_n`]
+
+This type of event descriptor is a simple enumeration where the value `n` is the number of outcomes in the event. `n` is not given because it can be found by parsing each outcome's length followed by the actual outcome until you has parsed the same number of bytes as the length field of the TLV.
```
Tibo-lg: > `n` is not given because it can be found by parsing each outcome's length followed by the actual outcome until you has parsed the same number of bytes as the length field of the TLV.

`n` is omitted because it can be deduced by parsing each outcome until reaching the end of the TLV.

(or at least fix the "you has")

Messaging.md
```diff
@@ -165,10 +173,69 @@ This type contains signatures of the funding transaction and any necessary infor
...
+
+Each `outcome_i` corresponds to the pre-image of a possible outcome that the oracle could sign.
```
Tibo-lg, line 203: I feel this sentence is useless, and "pre-image of a possible outcome" doesn't really make sense to me.

Messaging.md
```diff
@@ -165,10 +173,69 @@ This type contains signatures of the funding transaction and any necessary infor
...
+This type contains information about an announcement of an oracle to attest to an event in the future.
+
+#### Version 0 `oracle_announcement`
+
+1. type: 55332 (`oracle_announcement`)
+2. data:
+   * [`signature`:`annoucement_signature`]
+   * [`oracle_event`:`oracle_event`]
+
+The `annoucement_signature` is a signature of the hash of the serialized `oracle_event` using the `oracle_public_key`. This can be shared with peers that have already verified this oracle's public key.
```
Tibo-lg: > using the `oracle_public_key`

that is valid with respect to `oracle_public_key`.

Might be picky but sounds weird to say you make a signature using a public key.

Tibo-lg (commented):
So should all event descriptor TLVs be described here? I was in the process of defining them in my oracle proposal but I guess it will be redundant to have them in both places.

Messaging.md
benthecarman: Wanted to be clear that it isn't the hash like in the `contractInfo`

Messaging.md
benthecarman: I like this a lot better. I agree it sounded like you signed the pub key but didn't know how to say it better.

Messaging.md
benthecarman: This is my implementation of it, this should be doable for everyone:

```
    val iter = ValueIterator(value)
    val builder = Vector.newBuilder[String]

    while (iter.index < value.length) {
      val len = UInt16(iter.takeBits(16))
      val outcomeBytes = iter.take(len.toInt)
      val str = new String(outcomeBytes.toArray, StandardCharsets.UTF_8)
      builder.+=(str)
    }

    EnumEventDescriptorV0TLV(builder.result())
```

You don't know `n` until after you've parsed all of them

Tibo-lg: I would like to check that this is feasible with ldk as is, please remind me if I don't update.

nkohen: I agree that it isn't absolutely vital to include `n` but I think it is good practice to allow length to be known without requiring linear time computation, the trade-off of adding a UInt16 to the front seems worth it and is good practice

benthecarman: I guess it's just 2 bytes can just add it.

benthecarman: Added.

Messaging.md
```diff
@@ -165,10 +175,84 @@ This type contains signatures of the funding transaction and any necessary infor
...
+2. data:
+   * [`int32`:`start`]
+   * [`int32`:`stop`]
+   * [`u16`:`step`]
+
+`start` refers to the first possible outcome number
+
+`end` refers to the last possible outcome number
+
+`step` refers to the increment between each outcome
```
LLFourn, line 217: this should be `start: i32`, `step: u32`, `count: u32` to avoid invalid states like `end` not coinciding with a step.

Messaging.md
```diff
@@ -165,10 +175,84 @@ This type contains signatures of the funding transaction and any necessary infor
...
+
+### The `oracle_event` Type
+
+This type contains information about an event for which an oracle plans on releasing a signature over.
+
+#### Version 0 `oracle_event`
+
+1. type: 55330 (`oracle_event_v0`)
+2. data:
+   * [`x_point`:`oracle_public_key`]
```
LLFourn, line 227: I think public key should not be included in the TLV. You should verify the signature on the data before decoding it which by definition means you already know the public key. This can cause a bug where someone verifies the signature the public key included rather than the one they have stored.

LLFourn (commented):
I think this is not viable for me on the oracle side.
Writing custom serializes and deserializers is very error prone and tedious.
Another concern is that it's hard to add custom fields.
What do I do if I want to add a nested structure to inform the clients about the teams long names, logos etc? Adding a new fields is really hard and makes it difficult to experiment with different metadata. Note that all of this metadata should be signed and I *guess* transferred via the P2P layer.

I realize custom encoding is what they did in lightning and you are trying to follow their lead but this is far outside from the domain of lightning and even Bitcoin in general.
This is the way in which a trusted party should announce an event it will attest to.
I would rather use a standard serialization format like CBOR. Even JSON would be more appropriate.
This makes implementations much easier to write. For example, I could implement the entire spec with just this code in rust and have it be correct by construction:

```rust
#[derive(Deserialize, Serialize)
#[serde(rename_all = "snake_case", tag = "type")]
enum EventDescriptor {
       Enumerated { outcomes: Vec<String> },
       Range { start: i32, step: u32, count: u32 }
}

#[derive(Deserialize, Serialize)]
struct Event {
   oracle_nonce: XOnly,
   event_maturity_epoch: u32,
   event_descriptor: EventDescriptor,
   event_uri: String
}
```

Implementing this suggestion would take quite a lot of time and effort both now and in the future to maintain.

Does this make sense?

Messaging.md
benthecarman: If the public key is not included how do you know which oracle to check against?

Messaging.md
benthecarman: I like this better

Messaging.md
LLFourn: Good point. I think a string `oracle_id` along with with the TLV where you send the signature. It is not signed but lets the client know which key to use (and if they don't have it, potentially where to get it). 

e.g.
- theoracle.com
- twitter.com/OutcomeObserver

LLFourn:
Perhaps a way to reconcile my comments on encoding is to simply have a field `encoding` in the TLV which specifies the encoding of the announcement. Implementations can choose to support whichever encodings they want. I specifically want to be able to encode the announcement in JSON and just sign a JSON string for now.

Messaging.md
LLFourn: Another thing you could do for now is just move the public key into the `oracle_announcement` section which makes it a bit less awkward.

Messaging.md
```diff
@@ -165,10 +175,84 @@ This type contains signatures of the funding transaction and any necessary infor
...
+   * [`x_point`:`oracle_nonce`]
+   * [`u32`:`event_maturity_epoch`]
+   * [`event_descriptor`:`event_descriptor`]
+   * [`string`:`event_uri`]
+
+`event_maturity_epoch` refers to the earliest time this event (UTC) is expected to be signed, in epoch seconds.
```
Christewart, line 233: Just so everyone is aware, this hits the `UInt32` max value (`4294967295`) at 

>02/07/2106 @ 6:28am (UTC) 

So ~80 years.

benthecarman: Bitcoin has the same problem with its block header, we could instead use UInt64 to avoid this

PR #108, opened 2020-10-14
Use oracle_announcement in place of distinct oracle_info

nkohen:


Messaging.md
```diff
@@ -88,30 +86,18 @@ The following DLC-specific types are used throughout the specification. All type
...
-1. type: 42770 (`oracle_info_v0`)
-2. data:
-   * [`x_point`:`oracle_public_key`]
-   * [`x_point`:`oracle_nonce`]
-
-This type of oracle info is for single-oracle, single signature (and hence single nonce) events.
+This type of contract info is a simple enumeration where the value `n` is omitted from being
+explicitly included as it can be derived from the length field of the TLV and should already be
+known from the corresponding `oracle_announcement`.
+The order of payouts matches the order of the outcomes in the `oracle_announcement`.
```
benthecarman, line 100: The `oracle_announcement` doesn't have the outcomes, it is actually `oracle_announcement` -> `oracle_event` -> `event_descriptor`

benthecarman: Maybe should exclude multi nonce descriptors here because I would have no idea what the ordering would be here

benthecarman (changes requested):
Concept ACK

Messaging.md
nkohen: not sure what you mean? If you find an enum_contract_info you better have gotten an enum_event_descriptor

nkohen:
Replaced by #130

PR #110, opened 2020-10-28
Numeric Outcome DLCs

### LINKED ISSUE #65: Decomposing quantities into multiple signatures
In my URI format proposal a point came up about "exponent + mantissa" vs decimal decomposition [came up](https://github.com/discreetlogcontracts/dlcspecs/pull/63#discussion_r474335179). This deserves it's own discussion I think.

WRT to what @Tibo-lg said:

> I considered this as well, but I think the mantissa exponent as described in the original paper is actually better because you only need 2 R-values for any case and you don't have issues of going out of bounds.

I have not done a rigorous investigation into this topic and perhaps it deserves one but in my mind exponent + mantissa is in practice much less efficient for users than decimal decomposition (which is less efficient for Oracles but I think that is ok).

How many possible mantissas should there be for BTC/USDT for example? 1,024? Then every contract has to make a transaction covering all 1,024 mantissas. Covering all possible relevant exponents then may double or triple this number.

Decimal decomposition is having one signature for each digit and having an upper limit on the value e.g. $999,999 for BTC/USDT requires 6 oracle signatures. I claim that this means users have to create far fewer transactions (see the referenced text for an example) and they are simpler from an implementation perspective. Furthermore, they are not lossy (unless the price goes over the threshold) which means you can determine what the exact price was from the signatures up to the digit resolution (rather than an approximation of it).

There is another reason you want full resolution: you want people making a contract a year ago on the price tomorrow and people making a contract today on the price tomorrow to use the same event/oracle signatures. If the price has changed a lot e.g. 10x in the last year then the mantissa will not be covering the lower digits of the actual outcome which may be very relevant to people who are betting on short daily price swings. 

I don't think the bounds really differentiates the two schemes except that in one the bounds move exponentially and the other they move linearly. In theory there should be a cap on the exponents too in case the USD or (USDT) goes full Zimbabwe and you need the entire supply of USDT to buy one satoshi!

More than happy to be proven wrong about this.
I think the goal of this issue is for someone to "do the math" and plot this scheme against exponent/mantissa for number of signatures needed by oracles vs number of outcomes users need to cover.

nkohen:
This introduces a new document which specifies how to construct and execute DLCs based on (multi-nonce) numeric outcomes.

This includes CET compression optimizations as well as an implementation of general payout curve support, which are combined to describe a (nearly) deterministic CET computation algorithm which is approximately minimal. (I am open to splitting this up into many files if people want).

Remaining TODOs:

- [ ] Add support for signed numeric outcomes (I'm currently leaning towards just running the algorithm on positive and negative numbers separately).
- [ ] Add integration with other spec files (specifically introducing a new contract info version and integrating precision ranges into the accept message somehow).
- [ ] Pretty explanatory diagrams and pictures
- [ ] Add explicit numbers around how much better base 2 is than any other base (specifically 10) in every way and make a recommendation in the spec that oracles support base 2 signing

And lastly I intend to create test vectors as I have almost all of this implemented in a branch on bitcoin-s but these tests shouldn't block this PR being discussed and merged as they will be added in a subsequent PR.

Fixes #65

### DIFF
--- a/CETCompression.md
+++ b/CETCompression.md
@@ -1,0 +1,379 @@
+# Contract Execution Transaction Compression
+
+## Introduction
+
+When constructing a DLC for a [numeric outcome](NumericOutcome.md), there are often an unreasonably large number of
+possible outcomes to construct a unique CET for every outcome.
+We remedy this fact with a CET compression mechanism specified in this document which allows
+any flat portions of the DLC's [payout curve](PayoutCurve.md) to be covered with only a logarithmic number of CETs.
+
+It is common for payout curves to have constant extremal payouts for a large number of cases
+representing all outcomes considered sufficiently unlikely.
+These intervals with constant extremal payouts are often called "collars" and these collars
+can be compressed to negligible size making the remaining number of CETs proportional
+to the number of sufficiently likely outcomes.
+Furthermore, through the use of [rounding intervals](NumericOutcome.md#rounding-intervals), even portions of the payout curve which are not
+completely flat can be compressed to some extent, normally causing the total number of CETs to be
+divided by some power of two.
+
+This is accomplished through the use of digit decomposition where oracles attesting to
+numeric outcomes sign each digit of the outcome individually.
+There are as many nonces as there are possible digits required and CETs are claimed using
+only some of these signatures, not necessarily all of them.
+
+When not all of the signatures are used, then that corresponding CET represents all events
+which agree on the digits for which signatures were used and may have any value at all other
+digits where signatures were ignored.
+
+## Table of Contents
+
+* [Adaptor Points with Multiple Signatures](#adaptor-points-with-multiple-signatures)
+* [CET Compression](#cet-compression)
+  * [Concrete Example](#concrete-example)
+  * [Abstract Example](#abstract-example)
+  * [Analysis of CET Compression](#analysis-of-cet-compression)
+    * [Counting CETs](#counting-cets)
+    * [Optimizations](#optimizations)
+  * [Algorithms](#algorithms)
+  * [Reference Implementations](#reference-implementations)
+* [Authors](#authors)
+
+## Adaptor Points with Multiple Signatures
+
+Given public key `P` and nonces `R1, ..., Rn` we can compute `n` individual signature points for
+a given event `(d1, ..., dn)` in the usual way: `si * G = Ri + H(P, Ri, di)*P`.
+To compute a composite adaptor point for all events which agree on the first `m` digits, where
+`m` is any positive number less than or equal to `n`, the sum of the corresponding signature
+points is used: `s(1..m) * G = (s1 + s2 + ... + sm) * G = s1 * G + s2 * G + ... + sm * G`.
+
+When the oracle broadcasts its `n` signatures `s1, ..., sn`, the corresponding adaptor secret can be
+computed as `s(1..m) = s1 + s2 + ... + sm` which can be used to broadcast the CET.
+
+#### Rationale
+
+This design allows implementations to re-use all [transaction construction code](Transactions.md) without modification
+because every CET needs as input exactly one adaptor point just like in the single-nonce setting.
+
+Another design that was considered was adding keys to the funding output so that parties could collaboratively
+construct `m` adaptor signatures and where `n` signatures are put on-chain in every CET which would reveal
+all oracle signatures to both parties when a CET is published.
+This design's major drawbacks is that it creates a very distinct fingerprint and makes CET fees significantly worse.
+Additionally it leads to extra complexity in contract construction.
+This design's only benefit is that it results in simpler (although larger) fraud proofs.
+
+The large multi-signature design was abandoned because the above proposal is sufficient to generate fraud proofs.
+If an oracle incorrectly signs for an event, then only the sum of the digit signatures `s(1..m)`
+is recoverable on-chain using the adaptor signature which was given to one's counter-party.
+This sum is sufficient information to determine what was signed however as one can iterate through
+all possible composite adaptor points until they find one whose pre-image is the signature sum found on-chain.
+This will determine what digits `(d1, ..., dm)` were signed and these values along with the oracle
+announcement and `s(1..m)` is sufficient information to generate a fraud proof in the multi-nonce setting. 
+
+## CET Compression
+
+Anytime there is a range of numeric outcomes `[start, end]` which result in the same payouts for all parties,
+then a compression function described in this section can be run to reduce the number of CETs to be `O(log(L))`
+instead of `L` where  `L = end - start + 1` is the length of the interval being compressed.
+
+Because this compression of CETs only works for intervals which result in the same payout, the [CET calculation algorithm](NumericOutcome.md#contract-execution-transaction-calculation)
+first splits the domain into buckets of equal payout, and then applies the compression algorithm from this
+document to individual intervals, `[start, end]` where all values in this interval have some fixed payout.
+
+Most contracts are expected to be concerned with some subset of the total possible domain and every
+outcome before or after that range will result in some constant maximal or minimal payout.
+This means that compression will drastically reduce the number of CETs to be of the order of the size
+of the probable domain, with further optimizations available when parties are willing to do some [rounding](NumericOutcome.md#rounding-intervals).
+
+The compression algorithm takes as input a range `[start, end]`, a base `B`, and the number of digits
+`n` (being signed by the oracle) and returns an array of arrays of integers (which will all be in the range `[0, B-1]`).
+An array of integers corresponds to a single event equal to the concatenation of these integers (interpreted in base `B`).
+
+### Concrete Example
+
+Before generalizing or specifying the algorithm, let us run through a concrete example.
+
+We will consider the range `[135677, 138621]`, in base `10`.
+Note that they both begin with the prefix `13` which must be included in every CET, for this purpose I omit these digits for
+the remainder of this example as we can simply examine the range `[5677, 8621]` and prepend a `13` to all results.
+
+To cover all cases while looking at as few digits as possible in this range we need only consider
+`5677`, `8621` and the following cases:
+
+```
+5678, 5679,
+568_, 569_,
+57__, 58__, 59__,
+
+6_, 7_,
+
+80__, 81__, 82__, 83__, 84__, 85__,
+860_, 861_,
+8620
+```
+
+where `_` refers to an ignored digit (an omission from the array of integers).
+(Recall that all of these are prefixed by `13`).
+Thus, we are able to cover the entire interval of `2944` outcomes using only `20` CETs!
+
+Here it is again in binary (specifically the range `[5677, 8621]`, not the original range with the `13` prefix in base 10):
+Outliers are `5677 = 01011000101101` and `8621 = 10000110101101` with cases:
+
+```
+0101100010111_,
+0101100011____,
+01011001______,
+0101101_______,
+010111________,
+011___________,
+
+100000________,
+1000010_______,
+100001100_____,
+10000110100___,
+100001101010__,
+10000110101100
+```
+
+And so again we are able to cover the entire interval of `2944` outcomes using only `14` CETs this time.
+
+### Abstract Example
+
+Before specifying the algorithm, let us run through a general example and do some analysis.
+
+Consider the range `[(prefix)wxyz, (prefix)WXYZ]` where `prefix` is some string of digits in base `B` which
+`start` and `end` share and `w, x, y, and z` are the unique digits of `start` in base `B` while `W, X, Y, and Z`
+are the unique digits of `end` in base `B`.
+
+To cover all cases while looking at as few digits as possible in this (general) range we need only consider
+`(prefix)wxyz`, `(prefix)WXYZ` and the following cases:
+
+```
+wxy(z+1), wxy(z+2), ..., wxy(B-1),
+wx(y+1)_, wx(y+2)_, ..., wx(B-1)_,
+w(x+1)__, w(x+2)__, ..., w(B-1)__,
+
+(w+1)___, (w+2)___, ..., (W-1)___,
+
+W0__, W1__, ..., W(X-1)__,
+WX0_, WX1_, ..., WX(Y-1)_,
+WXY0, WXY1, ..., WXY(Z-1)
+```
+
+where `_` refers to an ignored digit (an omission from the array of integers) and all of these cases have the `prefix`.
+
+### Analysis of CET Compression
+
+This specification refers to the first three rows of the abstract example above as the **front groupings** the fourth row
+in the example as the **middle grouping** and the last three rows in the example as the **back groupings**.
+
+Notice that the patterns for the front and back groupings are nearly identical.
+
+#### Counting CETs
+
+Also note that in total the number of elements in each row of the front groupings is equal to `B-1` minus the corresponding digit.
+That is to say, `B-1` minus the last digit is the number of elements in the first row and then the second to last digit and so on.
+Likewise the number of elements in each row of the back groupings is equal to the corresponding digit.
+That is to say, the last digit corresponds to the last row, second to last digit is the second to last row and so on.
+This covers all but the first digit of both `start` and `end` (as well as the two outliers `wxyz` and `WXYZ`).
+Thus the total number of CETs required to cover the range will be equal to the sum of the unique digits of `end` except the first, 
+plus the sum of the unique digits of `start` except for the first subtracted from `B-1` plus the difference of the first digits plus one.
+
+A corollary of this is that the number of CETs required to cover a range of length `L` will be `O(B*log_B(L))` because `log_B(L)`
+corresponds to the number of unique digits between the start and end of the range and for each unique digit a row is
+generated in both the front and back groupings of length at most `B-1 ` which corresponds to the coefficient in the order bound.
+
+This counting also shows us that base 2 is the optimal base to be using in general cases as it will, in general, outperform all larger bases
+in both large and small intervals.
+Note that the concrete example above was chosen to be easy to write down in base 10 (large digits in `start`, small digits in `end`) and so it should not
+be thought of as a general candidate for this particular consideration.
+
+To help with intuition on this matter, consider an arbitrary range of three digit numbers in base 10.
+To capture the same range in base 2 we need 10 digit binary numbers.
+However, a random three digit number in base 10 is expected to have a digit sum of 15, while a random ten digit binary number expects a digit sum of only 5!
+Thus we should expect base 2 to outperform base 10 by around 3x on average.
+This is because using binary results in a compression where each row in the diagram above has only a single element, which corresponds
+to binary compression's ability to efficiently reach the largest possible number of digits ignored which itself covers the largest number of cases.
+Meanwhile in a base like 10, each row can take up to 9 CETs before moving to a larger number of digits ignored (and cases covered).
+Another way to put this is that the inefficiency of base 10 which seems intuitive at small scales is actually equally present at *all scales*!
+
+One final abstract way of intuiting that base 2 is optimal is the following:
+We wish to maximize the amount of information that we may ignore when constructing CETs, because abstractly every bit of information ignored
+in a CET doubles the number of cases covered with a single transaction and signature.
+Thus, if we use any base other than 2, say 10, then we will almost always run into situations where redundant information is needed because we can
+only ignore a decimal digit at a time where a decimal digit has 3.3 bits of information.
+Meanwhile in binary where every digit encodes exactly a single bit of information, we are able to perfectly ignore all redundant bits of information
+resulting in some number near 3.3 times fewer CETs on average.
+
+#### Optimizations
+
+Because `start` and `end` are outliers to the general grouping pattern, there are optimizations that could potentially be made when they are added.
+
+Consider the example in base 10 of the range `[2200, 4999]` which has the endpoints `2200` and `4999` along with the groupings
+
+```
+2201, 2202, 2203, 2204, 2205, 2206, 2207, 2208, 2209,
+221_, 222_, 223_, 224_, 225_, 226_, 227_, 228_, 229_,
+23__, 24__, 25__, 26__, 27__, 28__, 29__,
+
+3___,
+
+40__, 41__, 42__, 43__, 44__, 45__, 46__, 47__, 48__,
+490_, 491_, 492_, 493_, 494_, 495_, 496_, 497_, 498_,
+4990, 4991, 4992, 4993, 4994, 4995, 4996, 4997, 4998
+```
+
+This grouping pattern captures the exclusive range `(2200, 4999)` and then adds the endpoints in ad-hoc to get the inclusive range `[2200, 4999]`.
+But this method misses out on a good amount of compression as re-introducing the endpoints allows us to replace the first two rows with
+a single `22__` and the last 3 rows with just `4___`.
+
+This optimization is called the **endpoint optimization**.
+
+More generally, let the unique digits of `start` be `start = (x1)(x2)...(xn)` then if `x(n-i+1) = x(n-i+2) = ... = xn = 0`, then the
+first `i` rows of the front groupings can be replaced by `(x1)(x2)...(x(n-i))_..._`.
+In the example above, `start` ends with two zeros so that the first two rows are replaced by `22__`.
+
+Likewise, let the unique digits of `end` be `end = (y1)(y2)...(yn)` then if `y(n-j+1) = y(n-j+2) = ... = yn = B-1`, then the last `j` rows
+of the back groupings can be replaced by `(y1)(y2)...(y(n-j))_..._`.
+In the example above, `end` ends with three nines so that the last three rows can are replaced by `4___`.
+
+There is one more optimization that can potentially be made.
+If the unique digits of `start` are all `0` and the unique digits of `end` are all `B-1` then we will have no need for a middle grouping as we can cover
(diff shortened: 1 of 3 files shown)


NumericOutcome.md
```diff
@@ -0,0 +1,437 @@
...
+
+#### Rationale
+
+This design allows implementations to re-use all [transaction construction code](Transactions.md) without modification
+because every CET needs as input exactly one adaptor point just like in the single-nonce setting.
+
+Another design that was considered was adding keys to the funding output so that parties could collaboratively
+construct `m` adaptor signatures and where `n` signatures are put on-chain in every CET which would reveal
+all oracle signatures to both parties when a CET is published.
+This design's major drawbacks is that it creates a very distinct fingerprint and makes CET fees significantly worse.
```
benthecarman: Makes tx building and negotiation more complicated too

NumericOutcome.md
```diff
@@ -0,0 +1,437 @@
...
+
+(w+1)___, (w+2)___, ..., (W-1)___,
+
+W0__, W1__, ..., W(X-1)__,
+WX0_, WX1_, ..., WX(Y-1)_,
+WXY0, WXY1, ..., WXY(Z-1)
+```
+
+where `_` refers to an ignored digit (an omission from the array of integers) and all of these cases have the `prefix`.
+I refer to the first three rows as the **front groupings** the fourth row as the **middle grouping** and the last three rows
```
benthecarman: I'd try to stay away from first person in the spec.

NumericOutcome.md
```diff
@@ -0,0 +1,437 @@
...
+Note that there are two more possible optimizations to be made, which I call the **row optimization**, using the outliers `wxyz` and `WXYZ`.
+If `z=0` then the entire first row can be replaced with `wxy_` and if `Z=B-1` then the entire last row can be replaced with `WXY_`.
+There are another two possible optimizations in the case where the front or back groupings are not needed, which
+I call **grouping optimization**, that again use the outliers to the above pattern `wxyz` and `WXYZ`.
+For the front this is the case when `x=y=z=0` so that the front groupings can be replaced with `w___`.
+Likewise if `X = Y = Z = B-1` then the back groupings can be replace with `W___`.
+Lastly, if both grouping optimizations can be made, `w = 0` and `W = B-1` then a **total optimization** can be made and the
+whole range can be represented using only a single CET corresponding to `(prefix)____`.
+
+### Algorithms
```
benthecarman: The functions in these should probably be written in a more universal language like python or just psdueo code

benthecarman (commented):
Some initial review

NumericOutcome.md
```diff
@@ -0,0 +1,437 @@
...
+all possible composite adaptor points until they find one whose pre-image is the signature sum found on-chain.
+This will determine what digits `(d1, ..., dm)` were signed and these values along with the oracle
+announcement and `s(1..m)` is sufficient information to generate a fraud proof in the multi-nonce setting. 
+
+## Contract Execution Transaction Compression
+
+Anytime there is a range of numeric outcomes `[start, end]` which result in the same payouts for all parties,
+then a compression function can be run to reduce the number of CETs to be `O(log(L))` instead of `L` where 
+`L = end - start + 1` is the length of the interval being compressed.
+
```
Ichiro0219, line 71: Does O mean base?
If so, is this formula O * log O (L)?

NumericOutcome.md
```diff
@@ -0,0 +1,437 @@
...
+WXY0, WXY1, ..., WXY(Z-1)
+```
+
+where `_` refers to an ignored digit (an omission from the array of integers) and all of these cases have the `prefix`.
+I refer to the first three rows as the **front groupings** the fourth row as the **middle grouping** and the last three rows
+as the **back groupings**.
+
+Notice that the patterns for the front and back groupings are nearly identical, and that in total the number of CETs that
+will be required to cover the range will be equal to the sum of the unique digits of `end` plus the sum of `B-1` minus the
+unique digits of `start`.
```
Ichiro0219: Is it `B-1` minus sum of the unique digits of `start`?

nkohen: no, if `start = 123`, and `B = 10` then I am referring to `(9-1) + (9-2) + (9-3)` as this is the number of columns in the front grouping row corresponding to that digit

NumericOutcome.md
nkohen: Curious to see what people who code in non-scala think about the readability, please leave comments here!

NumericOutcome.md
nkohen: I will try to make this more clear though

nkohen: There, lmk what you think!

NicolasDorier:
I must say I have a hard time understanding it from the Example part, but this is probably not your fault, I am bad at understanding when things are too abstract.

I would suggest, instead of using abstraction, just assuming B=10 and take a real, concrete example of range.
Or maybe as a more "concrete example" new part.

NumericOutcome.md
```diff
@@ -0,0 +1,437 @@
...
+This means that the number of CETs required to cover a range of length `L` will be `O(B*log_B(L))` because `log_B(L)`
+corresponds to the number of unique digits between the start and end of the range and for each unique digit a row is
+generated in both the front and back groupings of length at most `B-1 ` which corresponds to the coefficient in the order bound.
+This counting shows us that base 2 is the optimal base to be using in general cases as it will outperform all larger bases
+in both large and small ranges in general.
+
```
Ichiro0219: The number of CETs can be further reduced by selecting  (B^ n-1) or  (B ^ n) for `start` and `end`.

nkohen: So while this is true, I'm not sure that it is particularly practical?

NumericOutcome.md
```diff
@@ -0,0 +1,453 @@
...
+only feasible if the number of possible outcomes is of manageable size. 
+
+When an outcome can be any of a large range of numbers, then using a simple enumeration of
+all possible numbers in this range is unwieldy.
+We optimize by using numeric decomposition in which the oracle signs each digit of the outcome
+individually so that many possible outcomes can be compressed into a single CET by ignoring certain signatures.
+There will be as many nonces as there are possible digits required and CETs are claimed using
+some number of these signatures, not necessarily all of them.
+If not all signatures are used, then this CET corresponds to all events which agree on the digits for which
+signatures are used and may have any value at all other digits for which signatures are ignored.
```
Tibo-lg: IMHO this paragraph is very hard to read and the concept is already explained in the oracle spec that is under PR (and for which @Christewart is even adding more detailed example), so I would just remove that and reference the oracle spec.

NumericOutcome.md
```diff
@@ -0,0 +1,453 @@
...
+Most contracts are expected to be concerned with some subset of the total possible domain and every
+outcome before or after that range will result in some constant maximal or minimal payout.
+This means that compression will drastically reduce the number of CETs to be of the order of the size
+of the probable domain, with further optimizations available when parties are willing to do some rounding.
+
+The compression algorithm takes as input a range `[start, end]`, a base `B`, and the number of digits
+`n` (being signed by the oracle) and returns an array of arrays of integers (which will all be in the range `[0, B-1]`).
+An array of integers corresponds to a single event equal to the concatenation of these integers (interpreted in base `B`).
+
+### Example
```
Tibo-lg: Your example is too complex and you add too much details (which should be included somewhere though). Better to take a simple example in base 10 so that it reads easily, and then convince that actually base 2 is better.

NumericOutcome.md
```diff
@@ -0,0 +1,453 @@
...
+To help with intuition on this matter, consider an arbitrary range of three digit numbers in base 10.
+To capture the same range in base 2 we need 10 digit binary numbers.
+However, a random three digit number in base 10 is expected to have a digit sum of 15, while a random ten digit binary number expects a digit sum of only 5!
+Thus we should expect base 2 to outperform base 10 by around 3x on average.
+This is because using binary results in a compression where each row in the diagram above has only a single element, which corresponds
+to binary compression's ability to efficiently reach the largest possible number of digits ignored which itself covers the largest number of cases.
+Meanwhile in a base like 10, each row can take up to 9 CETs before moving to a larger number of digits ignored (and cases covered).
+Another way to put this is that the inefficiency of base 10 which seems intuitive at small scales is actually equally present at *all scales*!
+
+Note that there are two more possible optimizations to be made, which this specification calls the **row optimization**, using the outliers `wxyz` and `WXYZ`.
```
Tibo-lg: That doesn't belong in the example at all, maybe an "optimizations" sub-section wouldn't be a bad idea.

NumericOutcome.md
Tibo-lg: I agree with @benthecarman. Tbh it's fine by me since I got a little bit used to read Scala by now, but it definitely adds to the complexity of something that is already not that easy. I think Python would probably be better. I'm not sure for pseudo code, personally I often find it harder to read than actual code. Also still good to link to your implementation.

NumericOutcome.md
```diff
@@ -0,0 +1,453 @@
...
+    currentNum = currentNum / base // Note that this is integer division
+
+    digit.toInt
+  }
+  
+  backwardsDigits.reverse
+}
+```
+
+I will use this function for the purposes of this specification but note that when iterating through a range of sequential numbers,
```
Tibo-lg: Nit: `We` (to be consistent)

nkohen:
> I would suggest, instead of using abstraction, just assuming B=10 and take a real, concrete example of range.
> Or maybe as a more "concrete example" new part.

Great idea, I've added a concrete example before the general abstract one

NumericOutcome.md
```diff
@@ -0,0 +1,453 @@
...
+      val groupings = front ++ middle ++ back
+
+      groupings.map { digits =>
+        prefixDigits ++ digits
+      }
+    }
+}
+```
+
+## General Payout Curves
```
Tibo-lg: If you want to have something merge in the not so distant future, I would take that in a different PR. It's going to be difficult to give you feedback on that without actually trying to implement it myself I feel.

Tibo-lg (commented):
Overall looks great! Maybe just need to be made a bit easier to read.

NicolasDorier:
@nkohen  thank you for the concrete example, it is quite good, though I still have trouble understanding the optimization.

Trying to reformulate with my own word to know if I understood correctly:

Take an example of this range in b10: `[010000-012999]`
I thought you could compress all digits, not only the back and the front, in term of CET you would need:

```
n0=0
n1=1
n2=[0,2]
n3=[0,9]
n4=[0,9]
n5=[0,9]
```
Normally it would require: `1 * 1 * 3 * 9 * 9 * 9 = 2187 CET` right?
However, for any digit which can be any digit in the base, you can ignore them, and thus you need only
`1 * 1 * 3 = 3 CET`.

Is it the essence of the compression step?

NumericOutcome.md
nkohen: To be completely honest I think that it doesn't make much sense for these kinds of concerns to be included in the oracle spec for anything other than motivation whereas here they are necessary so I think it should be the other way around.

That said, I am more than happy to re-do the paragraph to make it more readable!

NicolasDorier:
^ I think it would be nice to have a concrete example of all compressions you are talking about.

NicolasDorier:
For `[010000-012989]` however, you can't just ignore the last one. So you would need `1 * 1 * 3 * 8 * 9` CET ?

NicolasDorier:
Removed my comments.
I think I figured out how to compute all the CET, but I fail to understand the three optimization, it would be helpful to have a concrete example.

nkohen:
> Removed my comments.
> I think I figured out how to compute all the CET, but I fail to understand the three optimization, it would be helpful to have a concrete example.

Just pushed an attempt to make the optimizations more clear.

nkohen:
Ugh, and as I did that I realized that there are actually some optimizations I missed and that all of the optimizations including the ones I've missed can be stated cleverly as a single optimization, I will be pushing this change up shortly

NumericOutcome.md
NicolasDorier: As C# guy I can understand the code, we have very similar constructs.
But I think if Scala is chosen, at the very least it should not use fancy constructions.

NumericOutcome.md
```diff
@@ -0,0 +1,542 @@
...
+6_, 7_,
+
+80__, 81__, 82__, 83__, 84__, 85__,
+860_, 861_,
+8620
+```
+
+where `_` refers to an ignored digit (an omission from the array of integers).
+(Recall that all of these are prefixed by `13`).
+Thus, we are able to cover the entire interval of `2944` outcomes using only `20` CETs!
```
LLFourn: I seem to have completely missed what you are talking about. You care about each of the outcomes which all lead to a different distribution of funds. How can you express 2944 different payouts in only 20 CETs!?

NicolasDorier: Consider the range `[135677, 138621]`.
Let's take a random number in the interval `136412`.

If you encrypted the signature of a CET with a pubkey being the sum of the sigpoints of the three first digit being `136`, then ANY attestation signing for those three digits can be used to decrypt the CET.
And this is fine because `136***` is always in the wanted interval, regardless of the other `***` signatures.

Basically he disregards the less significant digits when he knows for sure that whichever they are, the attestation would still fall into the range.

Got hard time to understand, tried to do the same with my own method and found the same 20 CET.

nkohen: @LLFourn The only ranges being considered for compression have a constant payout value, perhaps this should be made more clear?

nkohen: From the beginning of this section:
```
Anytime there is a range of numeric outcomes `[start, end]` which result in the same payouts for all parties,
then a compression function can be run to reduce the number of CETs to be `O(log(L))` instead of `L` where 
`L = end - start + 1` is the length of the interval being compressed.
```

LLFourn: Ok thanks. My bad. I didn't get the `same payout` bit! I see now where you are going with this.

Perhaps try and explain that applying the compression algorithm involves first splitting the the range of interest into buckets with the same payout. You then use your trick to compress the number of CETs in a bucket.

NumericOutcome.md
Ichiro0219: This idea is valid in both decimal and binary numbers, so I think it is practical. As a more practical description, it may be better to add the description like below to the #optimizations chapter.

Choosing a value closer to (B ^ n) for `start` and `end` will reduce the number of CETs.

NumericOutcome.md
Ichiro0219: LGTM

NumericOutcome.md
```diff
@@ -0,0 +1,542 @@
...
+W0__, W1__, ..., W(X-1)__,
+WX0_, WX1_, ..., WX(Y-1)_,
+WXY0, WXY1, ..., WXY(Z-1)
+```
+
+where `_` refers to an ignored digit (an omission from the array of integers) and all of these cases have the `prefix`.
+
+### Analysis of CET Compression
+
+This specification refers to the first three rows as the **front groupings** the fourth row as the **middle grouping** and the last three rows
```
Tibo-lg: It is not super clear what the rows are referring to

NumericOutcome.md
```diff
@@ -0,0 +1,542 @@
...
+40__, 41__, 42__, 43__, 44__, 45__, 46__, 47__, 48__,
+490_, 491_, 492_, 493_, 494_, 495_, 496_, 497_, 498_,
+4990, 4991, 4992, 4993, 4994, 4995, 4996, 4997, 4998
+```
+
+This grouping pattern captures the exclusive range `(2200, 4999)` and then adds the endpoints in ad-hoc to get the inclusive range `[2200, 4999]`.
+But this method misses out on a good amount of compression as re-introducing the endpoints allows us to replace the first two rows with
+a single `22__` and the last 3 rows with just `4___`.
+
+More generally, if the unique digits of `start = (x1)(x2)...(xn)` then if `x(n-i+1) = x(n-i+2) = ... = xn = 0`, then the first `i` rows
```
Tibo-lg: I don't get this sentence. Actually tbh I didn't get this optimization until looking at the code example.
Maybe:
More generally, let the unique digits of `start` be `start = (x1)(x2)...(xn)`. Then if `x(n-i+1) = x(n-i+2) = ... = xn = 0`, then the first `i` rows...

Also I think it could be useful to add an example for this (or just extend the previous one)

Tibo-lg: Just noticed you actually have the example just above... But the sentence could still use some rewording IMHO.

NumericOutcome.md
```diff
@@ -0,0 +1,542 @@
...
+  } else {
+    val fromFront = nonZeroDigits.init.flatMap { // Note the flatMap collapses the rows of the grouping 
+      case (lastImportantDigit, unimportantDigits) =>
+        val fixedDigits = digits.dropRight(unimportantDigits + 1)
+        (lastImportantDigit + 1).until(base).map { lastDigit => // Note that this range excludes lastImportantDigit and base
+          fixedDigits :+ lastDigit
+        }
+    }
+
+    nonZeroStartDigits.map(_._1).reverse +: fromFront // Add Endpoint
```
Tibo-lg: I think this should be `nonZeroDigits`?

NumericOutcome.md
```diff
@@ -0,0 +1,542 @@
...
+
+```scala
+def groupByIgnoringDigits(start: Long, end: Long, base: Int, numDigits: Int): Vector[Vector[Int]] = {
+    val (prefixDigits, startDigits, endDigits) = separatePrefix(start, end, base, numDigits)
+    
+    if (start == end) { // Special Case: Range Length 1
+        Vector(prefixDigits)
+    } else if (startDigits.forall(_ == 0) && endDigits.forall(_ == base - 1)) { // Total Optimization
+        Vector(prefixDigits)
+    } else if (prefix.length == numDigits - 1) { // Special Case: Front Grouping = Back Grouping
```
Tibo-lg: prefixDigits?

NumericOutcome.md
```diff
@@ -0,0 +1,542 @@
...
+Finally we are able to use all of these pieces to compress a range to an approximately minimal number of outcomes (by ignoring digits).
+
+```scala
+def groupByIgnoringDigits(start: Long, end: Long, base: Int, numDigits: Int): Vector[Vector[Int]] = {
+    val (prefixDigits, startDigits, endDigits) = separatePrefix(start, end, base, numDigits)
+    
+    if (start == end) { // Special Case: Range Length 1
+        Vector(prefixDigits)
+    } else if (startDigits.forall(_ == 0) && endDigits.forall(_ == base - 1)) { // Total Optimization
+        Vector(prefixDigits)
```
Tibo-lg: I think this doesn't work if there is no common prefix.

nkohen: If there is no common prefix and start digits are all `0` and end digits are all `base - 1` then that is just the entire range in which case there's no point in having a DLC

Tibo-lg: Good point, but if you go with computing positive and negative outcomes separately it might start to make sense.

NumericOutcome.md
```diff
@@ -0,0 +1,561 @@
...
+
+The large multi-signature design was abandoned because the above proposal is sufficient to generate fraud proofs.
+If an oracle incorrectly signs for an event, then only the sum of the digit signatures `s(1..m)`
+is recoverable on-chain using the adaptor signature which was given to one's counter-party.
+This sum is sufficient information to determine what was signed however as one can iterate through
+all possible composite adaptor points until they find one whose pre-image is the signature sum found on-chain.
+This will determine what digits `(d1, ..., dm)` were signed and these values along with the oracle
+announcement and `s(1..m)` is sufficient information to generate a fraud proof in the multi-nonce setting. 
+
+## Contract Execution Transaction Compression
```
Christewart: It might be useful to add a motivation section here? 

basically why do we need tx compression? Why can't we just do the dumb thing and have a CET for every payout.

NumericOutcome.md
```diff
@@ -0,0 +1,561 @@
...
+between the previous and next endpoints.
+The first integer is called `event_outcome` and contains the actual number that could be signed by the oracle
+(note: not in the serialization used by the oracle) which corresponds to an x-coordinate on the payout curve.
+The second integer is called `outcome_payout` and is set equal to the local party's payout should
+`event_outcome` be signed which corresponds to a y-coordinate on the payout curve.
+
+Note that this `payout_function` is from the offerer's point of view.
+To evaluate the accepter's `payout_function`, you must evaluate the offerer's `payout_function` at a given
+`event_outcome` and subtract the resulting payout from `total_collateral`.
+It is important that you do NOT construct the accepter's `payout_function` by replacing all `outcome_payout`s in the
```
Christewart: It's probably worth bolding this text and maybe making it it's own paragraph.

NumericOutcome.md
```diff
@@ -0,0 +1,561 @@
...
+In this section we detail the TLV serialization for a general `payout_function`.
+
+#### Version 0 payout_function
+
+1. type: ??? (`payout_function_v0`)
+2. data:
+   * [`u16`:`num_pts`]
+   * [`boolean`:`is_endpoint_1`]
+   * [`bigsize`:`event_outcome_1`]
+   * [`bigsize`:`outcome_payout_1`]
```
nkohen: I've been thinking about this more and I think I want to change this to allow payouts specified to have more precision. Obviously you can't have sub-satoshi payouts and rounding would still happen eventually but pre-maturely rounding here can reduce fidelity on other outcomes being interpolated and also breaks reproducibility when trying to use certain "on-paper equivalent" derivations of the same curve

CETCompression.md
```diff
@@ -0,0 +1,379 @@
...
+can be compressed to negligible size making the remaining number of CETs proportional
+to the number of sufficiently likely outcomes.
+Furthermore, through the use of [rounding intervals](NumericOutcome.md#rounding-intervals), even portions of the payout curve which are not
+completely flat can be compressed to some extent, normally causing the total number of CETs to be
+divided by some power of two.
+
+This is accomplished through the use of digit decomposition where oracles attesting to
+numeric outcomes sign each digit of the outcome individually.
+There are as many nonces as there are possible digits required and CETs are claimed using
+only some of these signatures, not necessarily all of them.
```
benthecarman, line 22: Maybe mention primairly using base 2 here

CETCompression.md
```diff
@@ -0,0 +1,379 @@
...
+
+This design allows implementations to re-use all [transaction construction code](Transactions.md) without modification
+because every CET needs as input exactly one adaptor point just like in the single-nonce setting.
+
+Another design that was considered was adding keys to the funding output so that parties could collaboratively
+construct `m` adaptor signatures and where `n` signatures are put on-chain in every CET which would reveal
+all oracle signatures to both parties when a CET is published.
+This design's major drawbacks is that it creates a very distinct fingerprint and makes CET fees significantly worse.
+Additionally it leads to extra complexity in contract construction.
+This design's only benefit is that it results in simpler (although larger) fraud proofs.
```
benthecarman: Maybe also mention the fraud proof is more precise

CETCompression.md
```diff
@@ -0,0 +1,379 @@
...
+This sum is sufficient information to determine what was signed however as one can iterate through
+all possible composite adaptor points until they find one whose pre-image is the signature sum found on-chain.
+This will determine what digits `(d1, ..., dm)` were signed and these values along with the oracle
+announcement and `s(1..m)` is sufficient information to generate a fraud proof in the multi-nonce setting. 
+
+## CET Compression
+
+Anytime there is a range of numeric outcomes `[start, end]` which result in the same payouts for all parties,
+then a compression function described in this section can be run to reduce the number of CETs to be `O(log(L))`
+instead of `L` where  `L = end - start + 1` is the length of the interval being compressed.
```
benthecarman: I think this would read better saying 

> can be run to reduce the number of CETs from `L` to `O(log(L))`

NumericOutcome.md
```diff
@@ -0,0 +1,143 @@
...
+then the amount to be used in the CET output for this party will be the closer of `value - (value % R)` or `value - (value % R) + R`, rounding
+up in the case of a tie.
+
+#### Reference Implementations
+
+* [bitcoin-s](https://github.com/bitcoin-s/bitcoin-s/blob/adaptor-dlc/core/src/main/scala/org/bitcoins/core/protocol/dlc/RoundingIntervals.scala)
+
+#### Rounding Interval Serialization
+
+1. type: ??? (`rounding_intervals_v0`)
```
benthecarman: pick number

PayoutCurve.md
```diff
@@ -0,0 +1,166 @@
...
+but instead can usually use only the user-provided parameters to directly compute a small number of relevant
+interpolation points.
+
+### Curve Serialization
+
+In this section we detail the TLV serialization for a general `payout_function`.
+
+#### Version 0 payout_function
+
+1. type: ??? (`payout_function_v0`)
```
benthecarman: pick number

CETCompression.md
```diff
@@ -0,0 +1,379 @@
...
+There is one more optimization that can potentially be made.
+If the unique digits of `start` are all `0` and the unique digits of `end` are all `B-1` then we will have no need for a middle grouping as we can cover
+this whole interval with just a single CET of `(prefix)_..._`.
+This optimization is called the **total optimization**.
+
+### Algorithms
+
+We will first need a function to decompose any number into its digits in a given base.
+
+```scala
```
benthecarman, line 248: These still should be written in something like python

benthecarman (changes requested):

CETCompression.md
nkohen: I'm not sure this is the best place for this

Tibo-lg (approved):
LGTM we can iterate over that

PR #113, opened 2020-11-05
Specify signing algo for Oracle

### LINKED ISSUE #111: (Tagged) Hashing and Signing by Oracle of Message
As @LLFourn mentions [here](https://github.com/discreetlogcontracts/dlcspecs/pull/55#discussion_r515442902), we need to specify a signing algorithm which includes a tagged hash for oracles signing messages.

We also need to decide what goes into this signed message other than just the outcome (e.g. the event id as well?)

benthecarman:
Closes #111

### DIFF
--- a/Oracle.md
+++ b/Oracle.md
@@ -24,6 +24,7 @@ This necessary information is committed to in a so-called [_event descriptor_](#
       - [Version 0 `oracle_event`](#version-0-oracle_event)
 - [Oracle announcements](#oracle-announcements)
       - [Version 0 `oracle_announcement`](#version-0-oracle_announcement)
+- [Signing Algorithm](#signing-algorithm)
 
 ## Event descriptor
 
@@ -118,6 +119,8 @@ UTF-8 is chosen as being a widely supported and easy to implement encoding forma
 For numerical outcomes represented in bases greater than 10, each digit should be converted to base 10 before being encoded (note that base 10 numbers in UTF-8 take values in the 0x0030-0x0039 range).
 This helps preventing any confusion about the capitalization of letters or the introduction of non-standard characters.
 
+Signing should be done using the [signing algorithm](#Signing-Algorithm) using the tag `attestation/v0`.
+
 ### Serialization of event descriptors
 
 Event descriptors should be serialized using [TLV format](https://github.com/lightningnetwork/lightning-rfc/blob/master/01-messaging.md#type-length-value-format) as described bellow.
@@ -193,7 +196,20 @@ The TLV serialization of oracle announcements is as follow.
    * [`x_point`:`oracle_public_key`]
    * [`oracle_event`:`oracle_event`]
 
-where `signature` is a Schnorr signature over a sha256 hash of the serialized `oracle_event`.
+where `signature` is a Schnorr signature over a sha256 hash of the serialized `oracle_event`, using the tag `announcement/v0`.
+
+## Signing Algorithm
+
+Signatures should be generated following the algorithm specified in [BIP 340](https://github.com/bitcoin/bips/blob/master/bip-0340.mediawiki).
+Tagged hashes should also be used as defined in the [design section of BIP 340](https://github.com/bitcoin/bips/blob/master/bip-0340.mediawiki#Design).
+
+The secret key (`sk`) should be the private key corresponding to the oracle's public key.
+
+The algorithm `Sign(sk, message, tag)` is defined as:
+
+* Let H = `tag_hash("DLC/oracle/" || tag)`
+* Let `m` = H(`message`)
+* Return `BIP340_sign(sk, m)`
 
 ## Footnotes
 


Oracle.md
```diff
@@ -113,12 +113,30 @@ If the rate is greater than $999999, the oracle signs the digits `9`, `9`, `9`,
 
 ### Serialization and signing of outcome values
 
-Every outcome value should be encoded as [a UTF-8 string with NFC normalization](./Messaging.md#Fundamental-types), before being passed to the signing algorithm.
+Every outcome value (`outcome_value`) should be encoded as [a UTF-8 string with NFC normalization](./Messaging.md#Fundamental-types), before being passed to the signing algorithm.
```
Tibo-lg: nit: I'm not sure why you add `outcome_value` here since you define it below

Oracle.md
```diff
@@ -113,12 +113,30 @@ If the rate is greater than $999999, the oracle signs the digits `9`, `9`, `9`,
...
 UTF-8 is chosen as being a widely supported and easy to implement encoding format.
 
 For numerical outcomes represented in bases greater than 10, each digit should be converted to base 10 before being encoded (note that base 10 numbers in UTF-8 take values in the 0x0030-0x0039 range).
 This helps preventing any confusion about the capitalization of letters or the introduction of non-standard characters.
 
+#### Signing Algorithm
+
+When signing an outcome [BIP 340](https://github.com/bitcoin/bips/blob/master/bip-0340.mediawiki) must be used.
```
Tibo-lg: nit: this sentence reads weird to me (ignore if it's just me). What about: "Signatures should be generated following the algorithm specified in BIP340".

Oracle.md
```diff
@@ -113,12 +113,30 @@ If the rate is greater than $999999, the oracle signs the digits `9`, `9`, `9`,
...
+Tagged hashes will also be used which is defined in the [design section of BIP 340](https://github.com/bitcoin/bips/blob/master/bip-0340.mediawiki#Design).
```
Tibo-lg: nit: "will also be used which is defined" -> "should also be used as defined"

Oracle.md
```diff
@@ -113,12 +113,30 @@ If the rate is greater than $999999, the oracle signs the digits `9`, `9`, `9`,
...
+
+The secret key (`sk`) should be the private key corresponding to the oracle's public key.
+
+The nonce (`R`) should be the nonce the oracle pre-committed to for this event.
```
Tibo-lg: Don't you need to mention `k` as well?

Tibo-lg (commented):
mainly nits

Oracle.md
benthecarman: Fixed.

Oracle.md
benthecarman: added

Oracle.md
benthecarman: Figured it is more clear

Oracle.md
```diff
@@ -113,12 +113,30 @@ If the rate is greater than $999999, the oracle signs the digits `9`, `9`, `9`,
...
+
+The message (`m`) should be calculated as follows:
+
+ * Let `tag` = Event descriptor name (eg a Version 0 enum_event_descriptor is `enum_event_descriptor_v0`)
+ * Let `outcome_value` = the serialized event outcome value being attested to by the oracle
+
+Message (`m`) will be: hash <sub>tag</sub> (R || outcome_value)
+
+Finally, the signature should satisfy the [BIP 340 Verification algo](https://github.com/bitcoin/bips/blob/master/bip-0340.mediawiki#verification).
+
```
LLFourn, line 123: Some questions:

- Is this only for signing the attestation? What about the announcement?
- Why is there R in the attestation message?

What do you thing about the following idea:

tags should separate your signatures from other *applications* not from different uses of signatures in the same application. So we could just use `oracle` (it would be nice to have spec numbers so we could be self-referential like BIP340) and then differentiate between announcement and attestation within the message itself.

I would also prefer to be explicit about the signing and verification algorithms. e.g

The signing algorithm `sign(sk, message)` takes as input:

- *sk*: a secp256k1 secret key
- *message*: the message to be signed as an array of bytes

and is defined as:

- Set *m* to hash<sub>oracle</sub>(*message*)
- Return BIP340_sign(*sk*, *m*)

[EDIT] discussion of this can continue in issue: ~https://github.com/discreetlogcontracts/dlcspecs/pull/113~ #111 (thanks @nkohen).

nkohen: in case anyone is confused, I think #111 is the issue

Oracle.md
```diff
@@ -195,3 +196,15 @@ The TLV serialization of oracle announcements is as follow.
...
+## Signing Algorithm
+
+Signatures should be generated following the algorithm specified in [BIP 340](https://github.com/bitcoin/bips/blob/master/bip-0340.mediawiki).
+Tagged hashes should also be used as defined in the [design section of BIP 340](https://github.com/bitcoin/bips/blob/master/bip-0340.mediawiki#Design).
+
+The secret key (`sk`) should be the private key corresponding to the oracle's public key.
+
+The algorithm `Sign(sk, message)` is defined as:
+
+* Let `m` = hash<sub>oracle</sub>(message)
```
LLFourn: Thanks @benthecarman. We had a short discussion about this last meeting.
There are at least two but potentially more things an oracle can sign for:

1. announcements
2. attestations

So we want to make sure to separate these things. I would have liked to avoid having different tagged hashes these *but* it looks like announcements are necessarily just encoded differently to attestations because:

1. announcements are structured data (so TLVs, JSON etc)
2. attestations are just single strings (I want this to make using oracle signatures easy and useful on their own and not require wonky encoding implementations).

If they were both encoded the same we could just have different markers for each type of message to separate and then a single "oracle" in the tagged hash. But given the different encoding requirements we might do best to push up the application separation into the tag.

I suggest two tags:

1. `oracle/attestation`
2. `oracle/announcement`

and therefore changing the spec here to Sign(sk, message, tag):

* Let H = tag_hash("oracle/" || tag);
* Let m = H(message)
* Return BIP340_sign(sk,m)

and then just invoking `Sign(sk,message,"announcement")` in the announcement part of the spec etc.
Also as we've discussed before it would be better to have an acronym for the specification documents so we can just put `BET04` in there rather than `oracle`.

benthecarman: Updated to use `oracle/attestation` & `oracle/announcement`.

> Also as we've discussed before it would be better to have an acronym for the specification documents so we can just put BET04 in there rather than oracle.

Since we don't have a nice acronym for the DLC spec we just do something like `DLCOraclev0/attestation`

LLFourn: Good idea. Maybe `DLC/oracle/attestation/v0` since it's a bit clearer?

benthecarman: Updated :)

LLFourn (approved):
LGTM :)

PR #114, opened 2020-11-10
Introduce first version of ECDSA adaptor signature spec

### LINKED ISSUE #50: Adaptor signature documentation / explaination
I think we should add some documentation on the repo about our choice to use adaptor signatures and what variant we are using with links to more extensive documentation on what they are. They are a core piece of how the DLC protocol works and aren't widely adopted in the ecosystem (yet), so it's a good idea to make sure we explain these. A 3rd party reading the DLC spec for the first time may not even be aware of what they are.

This is probably a subtask of #46

LLFourn:
This is a draft of the ECDSA adaptor signature spec.

Feel free to post an early review on structure, format and design decisions.
See if you can find any mistakes in the algorithms.

TODO:
- [x] test vectors
- [x] review Jonas' implementation to see if there are any meaningful differences (@jesseposner) 
- [ ] reference table

Fixes #50


ecdsa_adaptor.md
```diff
@@ -0,0 +1,196 @@
...
+## Adaptor Signatures
+
+*Adaptor signatures* are an efficient form of verifiable signature encryption that can achieve our required structure.
+Given a anticipated signature `Y` we may encrypt a signature (for the purposes of this document, an ECDSA signature) under `Y` such that once a party learns the discrete logarithm of `Y` they can decrypt it and get a valid ECDSA signature.
+
+An interesting aspect of adaptor signatures is that they leak the decryption key (the discrete logarithm of `Y`) when they are decrypted.
+In our case the leaked decryption key is is the scalar *s* value of a [[BIP340]] signature which turns out to be quite useful.
+The ability to recover the *s* value from the on-chain transaction actually improves the security of the protocol compared to the original DLC description in [[Dry]].
+To see why, imagine an oracle who engages in bets based on their own signatures.
+In order to not maintain their credibility they must publish the correct signature scalar `s_valid` publicly, but what is to stop them from privately generating a favorable and wrong `s_cheat` to settle their own bet?
```
Tibo-lg: to not -> to

ecdsa_adaptor.md
```diff
@@ -0,0 +1,196 @@
...
+The algorithms presented here resemble the scheme from [[Fou]] but the security proof of the scheme in [[Aum]] can be applied to it which requires a proof of knowledge of the discrete logarithm to be attached to each encryption key.
+Although we do not explicitly attach proofs of knowledge, in the DLC application the encryption key is the anticipated signature of an oracle which by definition the oracle must "know" and therefore we can safely omit the proof of knowledge.
+
+## Notation and Conventions
+
+- secp256k1 operations: We refer to *points* and *scalars* of the secp256k1 group and the operations between them.
+  - *n* denotes the curve order of secp256k1 `0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFEBAAEDCE6AF48A03BBFD25E8CD0364141`
+  - Scalars are integers between *0..(n-1)* inclusive.
+  - Scalars have three infix operations: multiplication (`*`) and addition (`+`) and subtraction (`-`) which are done modulo *n*.
+  - Scalars have a postfix operation (`^1`) which denotes inversion modulo *n*.
```
nkohen: I assume that is meant to be a `-1`

ecdsa_adaptor.md
```diff
@@ -0,0 +1,196 @@
...
+## Proof of Discrete Logarithm Equality
+
+As part of the ECDSA adaptor signature a proof of discrete logarithm equality must be provided.
+In our case this is a proof that the discrete logarithm of some `X` to the standard base `G` is the same as the discrete logarithm of some `Z` to the base `Y`.
+This proof can be constructed by using *equality composition* on two Sigma protocols proving knowledge of the discrete logarithm between both pairs of points.
+In other words the prover proves knowledge of `a` such that `X = a * G` and `b` such that `Z = b * Y` and that `a = b`.
+We make the resulting Sigma protocol non-interactive by applying the Fiat-Shamir transformation with SHA256 as the challenge hash.
+For background on Sigma protocols and making them non-interactive see [[Sch]].
+
+To disambiguate the proof we first *tag* a SHA256 as specified in [[BIP340]] with a string that identifies the kind of statement we are proving: `tag_string = "eq(DLG(secp256k1),DL(secp256k1))"`.
```
nkohen: is one `DLG` and the other `DL` in the tag intentionally?

ecdsa_adaptor.md
```diff
@@ -0,0 +1,196 @@
...
+
+- `a`: The adaptor signature.
+- `y`: The decryption *secret key*.
+
+and is defined as:
+
+- Parse `a` as `(R, R_a, s_a, proof)`
+- Set `s` to `s_a * y^-1`
+- Negate `s` if it is high as specified in [[BIP62]]
+- Set `r` to the x-coordinate of `R` modulo `q`
```
nkohen: `q` is not defined

ecdsa_adaptor.md
```diff
@@ -0,0 +1,196 @@
...
+
+- `Y`: The encryption *public key* the adaptor signature was encrypted under.
+- `a`: The ECDSA adaptor signature
+- `sig`: The ECDSA signature decrypted from `a`
+
+and is defined as follows:
+
+- Parse `a` as `(R, R_a, s_a, proof)`
+- Parse `sig` as `(r,s)`
+- Set `r_implied` to the x-coordinate of `R` modulo `q`
```
nkohen: `q` not defined

nkohen (commented):
A couple nits

ecdsa_adaptor.md
```diff
@@ -0,0 +1,196 @@
+# ECDSA adaptor signatures
+
+DLCs work by conditioning a transaction on the revelation of an oracle's signature .
+Concretely the approach is to encrypt a transaction signature such that the oracle's released signature is the decryption key.
+Once the oracle has revealed a signature, either party may use it to decrypt the corresponding transaction signature and broadcast the transaction.
+By this method an oracle's view of a real world event enforces a particualr distribution of funds between two parties according to their private bet.
```
benthecarman: > particualr

particular

ecdsa_adaptor.md
```diff
@@ -0,0 +1,196 @@
...
+Once the oracle has revealed a signature, either party may use it to decrypt the corresponding transaction signature and broadcast the transaction.
+By this method an oracle's view of a real world event enforces a particualr distribution of funds between two parties according to their private bet. 
+
+## Adaptor Signatures
+
+*Adaptor signatures* are an efficient form of verifiable signature encryption that can achieve our required structure.
+Given a anticipated signature `Y` we may encrypt a signature (for the purposes of this document, an ECDSA signature) under `Y` such that once a party learns the discrete logarithm of `Y` they can decrypt it and get a valid ECDSA signature.
+
+An interesting aspect of adaptor signatures is that they leak the decryption key (the discrete logarithm of `Y`) when they are decrypted.
+In our case the leaked decryption key is is the scalar *s* value of a [[BIP340]] signature which turns out to be quite useful.
```
benthecarman: > key is is 

key is

benthecarman (commented):
All of the links have an extra `[]` around them it seems as well

ecdsa_adaptor.md
```diff
@@ -0,0 +1,196 @@
...
+The ability to recover the *s* value from the on-chain transaction actually improves the security of the protocol compared to the original DLC description in [[Dry]].
+To see why, imagine an oracle who engages in bets based on their own signatures.
+In order to not maintain their credibility they must publish the correct signature scalar `s_valid` publicly, but what is to stop them from privately generating a favorable and wrong `s_cheat` to settle their own bet?
+In the original DLC protocol this was possible and difficult to catch.
+Using adaptor signatures the wronged party can extract the malicious `s_cheat` value from the on-chain signature and use it along with `s_valid` to punish the oracle and produce a proof of fraud.
```
Tibo-lg: It's not clear what you mean by punish here. Maybe better to stick with poof of fraud.

ecdsa_adaptor.md
```diff
@@ -0,0 +1,196 @@
...
+An appropriate ECDSA adaptor signature scheme that can be used in Bitcoin prior to the Taproot soft-fork was originally posted to the Bitcoin mailing list by [[Mor]] and then further refined by [[Fou],[Aum]].
+The algorithms presented here resemble the scheme from [[Fou]] but the security proof of the scheme in [[Aum]] can be applied to it which requires a proof of knowledge of the discrete logarithm to be attached to each encryption key.
+Although we do not explicitly attach proofs of knowledge, in the DLC application the encryption key is the anticipated signature of an oracle which by definition the oracle must "know" and therefore we can safely omit the proof of knowledge.
+
+## Notation and Conventions
+
+- secp256k1 operations: We refer to *points* and *scalars* of the secp256k1 group and the operations between them.
+  - *n* denotes the curve order of secp256k1 `0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFEBAAEDCE6AF48A03BBFD25E8CD0364141`
+  - Scalars are integers between *0..(n-1)* inclusive.
+  - Scalars have three infix operations: multiplication (`*`) and addition (`+`) and subtraction (`-`) which are done modulo *n*.
```
Tibo-lg: > and addition
, addition

ecdsa_adaptor.md
LLFourn: yes I've changed them all to using a superscript.

Tibo-lg (commented):
Just some nits. Also it might be better to keep file names consistents with the others (maybe `Ecdsa-Adaptor.md` instead of `ecdsa_adaptor.md`.

ecdsa_adaptor.md
LLFourn: Yes. The idea I am working towards is that this string should fully describe all the elements that follow in the hash.
DLG implies that `G` is being used and therefore we don't bother hashing it as part of the statement.
So following `eq(DLG(secp256k1),DL(secp256k1))` you have three points `X, Y, Z` (G is implicit) but following  `eq(DL(secp256k1),DL(secp256k1))` you would have four points `W,X,Y,Z`.

LLFourn:
@benthecarman That was intentional. I tried to put references in [ ]. I haven't put a list of references at the bottom yet. I find just putting links annoying because then you have to figure out what word to put them on. Perhaps now is the time to make a convention for how to do this?

NicolasDorier:
@LLFourn trying to make your test pass:

First vector, I can't parse the adaptor sig,

```
036c2f34833942e6eebb9e01b71c1d5f6ea185e7af607805ab5a39a829ad7a30400220f57f812e043e0f5b5c8c22a5c6efe6f76cdaa71490f7b3a3adb7a39101bf7ad4ee4d049e573adee6647ab7d97445d87c4252d91a627935269e99703a7d4161a5b4e2675d09e0da65deb2ed6210e5e99244153e9600ed17d90f6f31853e2c5ff8680adcbd1393d1ae8eed5cd8a8e396275be95506bf40a20ac22f66ca875bf4
```

It seems to fail here 

https://github.com/nkohen/secp256k1/commit/33834591c81616b2e4fde1b0c0a2ae7f88af1147#diff-af7e1bf67c999123aba5f3f14622539c6ac8eb6ce92676ccaa1f5c418bed6574R45

Basically the first FE seems to be too big?

NicolasDorier:
What is the `verification_key`?

LLFourn:
@NicolasDorier This code doesn't conform the spec yet.

From the spec: 

- Set `a` to `R || R_a || s_a || proof`

So it can parsed like
```
036c2f34833942e6eebb9e01b71c1d5f6ea185e7af607805ab5a39a829ad7a3040  #R
0220f57f812e043e0f5b5c8c22a5c6efe6f76cdaa71490f7b3a3adb7a39101bf7a      #R_a
d4ee4d049e573adee6647ab7d97445d87c4252d91a627935269e99703a7d4161   #s_a
a5b4e2675d09e0da65deb2ed6210e5e99244153e9600ed17d90f6f31853e2c5ff8680adcbd1393d1ae8eed5cd8a8e396275be95506bf40a20ac22f66ca875bf4 # proof
```

To make the code conform I think there are only a few small changes:

- The tagged hash for for the DLEQ proof challenge should be tagged as in the spec
- The order here:
https://github.com/nkohen/secp256k1/commit/33834591c81616b2e4fde1b0c0a2ae7f88af1147#diff-af7e1bf67c999123aba5f3f14622539c6ac8eb6ce92676ccaa1f5c418bed6574R82-R87

needs to be instead:
```c
    secp256k1_sha256_initialize_tagged(&sha, algo16, algo16_len(algo16)); # this needs to be the tag from the spec instead
    secp256k1_dleq_hash_point(&sha, p1);
    secp256k1_dleq_hash_point(&sha, gen2);
    secp256k1_dleq_hash_point(&sha, p2); 
    secp256k1_dleq_hash_point(&sha, r1);
    secp256k1_dleq_hash_point(&sha, r2);
```
- The serialization and de serialization code needs to match what I've specified. I think there is a difference here in what is considered the "proof" and what is considered the "adaptor_sig". The proof in what I've specified is only 64 bytes. What is important is that deserialization takes a 162 byte buffer what the internals are like after that is not too important.

I think that's all you'll need to change to get the verification test vectors to pass. After that I would go over the signing code and make sure the nonces are produced in a way that at least approximates the spec (everything that is recommended to be hashed to produce a nonce it hashed).

@jesseposner has offered to do the work to make the above happen.

> What is the verification_key?

I meant the public signing key. I've changed the test vector to use this instead.

NicolasDorier:
does that mean that all tests vectors that we finally all pass from @nkohen need to follow this new spec?
If not, it should be same format, else we are testing different things.

LLFourn:
@NicolasDorier Yes. Any end-to-end test vectors need to be updated. My main goal with this spec is to separate the DLEQ proof from the ECDSA adaptor (since it might be independently useful). Obviously another approach could be just to spec what's in the branch but I thought it would be better to come up with something a little more thought out and easy to justify (not that the all the justifications are in the document yet).

NicolasDorier:
Not to say the spec is frozen, but at this stage several implementation follow it, and breaking it without good reason (like security fix/features) will just waste everybody's time.

Both those tests and the specs must be in sync. If somebody come to this repository, and implement DLC starting by this document, most of his effort won't help him to implement the DLC specification. (Formatting data correctly is as much time consuming as monkey porting the math)

So in the worst case: A developer come here, pass the tests, then find out that it is not following the spec, update his code to follow the spec, but then the unit tests he did to pass this vector won't pass anymore.

NicolasDorier:
```
036c2f34833942e6eebb9e01b71c1d5f6ea185e7af607805ab5a39a829ad7a3040  #R
0220f57f812e043e0f5b5c8c22a5c6efe6f76cdaa71490f7b3a3adb7a39101bf7a      #R_a
d4ee4d049e573adee6647ab7d97445d87c4252d91a627935269e99703a7d4161   #s_a
a5b4e2675d09e0da65deb2ed6210e5e99244153e9600ed17d90f6f31853e2c5ff8680adcbd1393d1ae8eed5cd8a8e396275be95506bf40a20ac22f66ca875bf4 # proof
```

For `R`, in the spec, the first byte seems to have meaning. https://github.com/nkohen/secp256k1/commit/33834591c81616b2e4fde1b0c0a2ae7f88af1147#diff-af7e1bf67c999123aba5f3f14622539c6ac8eb6ce92676ccaa1f5c418bed6574R48

LLFourn:
> Not to say the spec is frozen, but at this stage several implementation follow it, and breaking it without good reason (like security fix/features) will just waste everybody's time.

My feeling was when I was asked to do this specification the expectation was that there would be breaking changes to what exists in the branch.

The purpose of this spec is to do the DLEQ proof something like how it a generalized framework for expressing these kinds of proofs would specify it (I am working on such a framework). Perhaps this is premature but I think it's better to be as close as possible to "textbook".
The time waste may additionally come when someone wants to use a DLEQ proof for their protocol but does want to use the branch as it has some rough edges to it -- so they re-write an incompatible one.

> For R, in the spec, the first byte seems to have meaning. nkohen/secp256k1@3383459#diff-af7e1bf67c999123aba5f3f14622539c6ac8eb6ce92676ccaa1f5c418bed6574R48

Well spotted. It seems that points are not being SECG encoded but rather being prefixed with a 0 or a 1 instead of the usual 2 or 3.

nkohen:
I think breaking changes here are okay/good to have, but I will note that I do not plan to actually make any other part of the DLC spec or implementations conform to this new ECDSA adaptor spec until there is a stable branch on secp256k1-zkp that we can all use. Once such a branch exists, it should be significantly easier for everyone to switch to using that branch instead of my temporary one and I have code to auto-regenerate the test vectors (which other impls should be able to pass, simply by using the new crypto library)

test/ecdsa_adaptor.json
```diff
@@ -0,0 +1,56 @@
+[
+    {
+        "adaptor_sig": "036c2f34833942e6eebb9e01b71c1d5f6ea185e7af607805ab5a39a829ad7a30400220f57f812e043e0f5b5c8c22a5c6efe6f76cdaa71490f7b3a3adb7a39101bf7ad4ee4d049e573adee6647ab7d97445d87c4252d91a627935269e99703a7d4161a5b4e2675d09e0da65deb2ed6210e5e99244153e9600ed17d90f6f31853e2c5ff8680adcbd1393d1ae8eed5cd8a8e396275be95506bf40a20ac22f66ca875bf4",
+        "message_hash": "8131e6f4b45754f2c90bd06688ceeabc0c45055460729928b4eecf11026a9e2d",
+        "public_signing_key": "039b8327d929a0e45285c04d19c9fffbee065c266b701972922d807228120e43f3",
```
jesseposner: It would be nice to include the private keys for the `public_signing_key`s so that `ecdsa_adaptor_encrypt` can be tested with these vectors.

ECDSA-adaptor.md
```diff
@@ -0,0 +1,209 @@
...
+
+DLCs work by conditioning a transaction on the revelation of an oracle's signature .
+Concretely the approach is to encrypt a transaction signature such that the oracle's released signature is the decryption key.
+Once the oracle has revealed a signature, either party may use it to decrypt the corresponding transaction signature and broadcast the transaction.
+By this method an oracle's view of a real world event enforces a particular distribution of funds between two parties according to their private bet.
+
+## Adaptor Signatures
+
+*Adaptor signatures* are an efficient form of verifiable signature encryption that can achieve our required structure.
+Given a anticipated signature `Y` we may encrypt a signature (for the purposes of this document, an ECDSA signature) under `Y` such that once a party learns the discrete logarithm of `Y` they can decrypt it and get a valid ECDSA signature.
```
jesseposner:
```suggestion
Given an anticipated signature `Y` we may encrypt a signature (for the purposes of this document, an ECDSA signature) under `Y` such that once a party learns the discrete logarithm of `Y` they can decrypt it and get a valid ECDSA signature.
```

jesseposner: nit: `Y` isn't a signature per se, bur rather a commitment to a signature, or a public key derived from an anticipated signature

ECDSA-adaptor.md
```diff
@@ -0,0 +1,209 @@
...
+
+An interesting aspect of adaptor signatures is that they leak the decryption key (the discrete logarithm of `Y`) when they are decrypted.
```
jesseposner: It might help to be a little more explicit here:
```suggestion
An interesting aspect of adaptor signatures is that the decryption key (the discrete logarithm of the encryption key `Y`) can be derived if both the plaintext (the valid ECDSA signature) and the ciphertext (the adaptor signature) are known.
```

ECDSA-adaptor.md
```diff
@@ -0,0 +1,209 @@
...
+The algorithms presented here resemble the scheme from [[Fou]] but the security proof of the scheme in [[Aum]] can be applied to it which requires a proof of knowledge of the discrete logarithm to be attached to each encryption key.
+Although we do not explicitly attach proofs of knowledge, in the DLC application the encryption key is the anticipated signature of an oracle which by definition the oracle must "know" and therefore we can safely omit the proof of knowledge.
+
+## Notation and Conventions
+
+- secp256k1 operations: We refer to *points* and *scalars* of the secp256k1 group and the operations between them.
+  - *n* denotes the curve order of secp256k1 `0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFEBAAEDCE6AF48A03BBFD25E8CD0364141`
+  - Scalars are integers between *0..(n-1)* inclusive.
+  - Scalars have three infix operations: multiplication (`*`), addition (`+`) and subtraction (`-`) which are done modulo *n*.
+  - Scalars have a postfix operation (`⁻¹`) which denotes inversion modulo *n*.
```
jesseposner:
```suggestion
  - Scalars have a postfix operation (`⁻¹`) which denotes multiplicative inversion modulo *n*.
```

ECDSA-adaptor.md
```diff
@@ -0,0 +1,209 @@
...
+  - Points are members of the secp256k1 group.
+  - Points have two binary operations between them: addition (`+`) and subtraction (`-`) which represent the group operation and the inverse group operation respectively.
+  - Points and scalars have a infix multiplication (`*`) operation which for `s * P` means adding the point `P` to itself `s` times.
+  - *G* is the standard generator point of secp256k1 with coordinates:
```
jesseposner, line 35:
```suggestion
  - *G* is the generator point of secp256k1 with coordinates:
```

ECDSA-adaptor.md
```diff
@@ -0,0 +1,209 @@
...
+
+and is defined as:
+
+- Set `k` to `sample_nonce(Y || message_hash || x)`
+- Set `m` to `scalar(message_hash)`
+- Set `R_a` to `k * G`
+- Set `R` to `k * Y`
+- Set `proof` to `DLEQ_prove(k, (R_a, Y, R))`
+- Set `r` to the x-coordinate of `R` modulo `n` (i.e. convert it to a scalar)
+- Set `s_a` to `k^-1 (m + r * x)`
```
jesseposner:
```suggestion
- Set `s_a` to `k⁻¹ (m + r * x)`
```

ECDSA-adaptor.md
LLFourn: I think `Y` should just be "public key"? Just a typo I think.

test/ecdsa_adaptor.json
LLFourn: The problem is that I don't want to specify exactly how to do nonce generation (only what must be input into the nonce generation function). So what you get out from `ecdsa_adaptor_encrypt` is not going to be the same on all implementations since the spec allows nonce generation to be done in different ways

For example what I do in my own implementation is to hash all the things as described above, add randomness into the hash and then to use the output to seed ChaCha20 from which I generate all nonces. The reason I do this is that in my library you can compose arbitrary sigma ZKPs which may need hundreds of nonces. This would be an awkward way to specify this particular proof though since you only need one nonce for the proof.

ECDSA-adaptor.md
LLFourn: The name I use for the commitment/public key is "anticipated signature". It came from  Tadge: https://diyhpl.us/wiki/transcripts/discreet-log-contracts/.

I *think* I am using it correctly here.

test/ecdsa_adaptor.json
jesseposner: That makes sense. What if you included the nonce in the test vector? That way `ecdsa_adaptor_encrypt` could be tested with the private key and nonce from the test. Albeit, it would have to take a nonce as an optional input, which slightly changes the function signature from the spec.

ECDSA-adaptor.md
jesseposner: From that transcript, it looks like it's ambiguous as to whether "Anticipated Signature" refers to `Y` or just the scheme in general. However, my point might be a little pedantic and I don't think it matters all that much, and I can see the utility of having a shorthand like "Anticipated Signature."

jesseposner:
I implemented the spec in Python [here](https://github.com/jesseposner/dlc/blob/master/ecdsa_adaptor.py) and it runs and passes all the [test vectors](https://github.com/jesseposner/dlc/blob/master/ecdsa_adaptor.json).

jesseposner:
Here's my secp branch that updates the implementation to match the spec: https://github.com/jesseposner/secp256k1/tree/jp/ecdsa-adaptor-sigs

This commit has the diff showing the changes that are required to match the spec: https://github.com/bitcoin-core/secp256k1/commit/3246e15f9cb5aceefc6a16be5362928ff83ff463

I've also added the first test vector and it passes: https://github.com/bitcoin-core/secp256k1/commit/a8ae2126a41f24bc100a76e7deb00e0e8c21c2f0

I will be continuing to work on that branch to finish the TODO items and add more test coverage.

draft PR: https://github.com/ElementsProject/secp256k1-zkp/pull/117

test/ecdsa_adaptor.json
```diff
@@ -0,0 +1,56 @@
...
+        "error" : "s_a is the negated and therefore doesn't match R_a"
+    },
+    {
+        "adaptor_sig" : "024288a3e3d08ac29bab9a6396a0c4c2a74a923dbaca92c469fe38eb0bfde131b203127e25e61da3a9afbb0708eba169c1454f2a12785e3bd371b1a070d37c856b4a5adc9e4e988b4d30470fc94b838ebc202f6bf5fe66805cb957c89275cee34139d05ad315068f38bce87ffe269fed4cfb68438a536e64bda2d6bac695038b2499348a2a6fed030aa94952e36aaf002657e3ad5b1da9a2b90055675468e0011ee5",
+        "message_hash": "8131e6f4b45754f2c90bd06688ceeabc0c45055460729928b4eecf11026a9e2d",
+        "public_signing_key" : "03ef16dd7c75ca40cfeab2aa659f2201e857591df3de67494a4d1dae34587395e6",
+        "encryption_key" : "0254425b8f73754f16f758937fdab39a04ea61763f9c8b2ba8abe082882ddbf155",
+        "decryption_key" : "7f82dd63f4f75c33da444b72372be3aa43c0027a076bf9675eb7932695d127a4",
+        "signature": "bd775c1c2f753d6454659c695f3b3d57701c9f2be4b5dbd1c1997380d2550f8f1321d75e2d155437f4f5b918c9becb327096dd5e72343f39c51565b6110f78be",
+        "error": "signature R value does not match encrypted signature"
```
LLFourn: Hmm I don't think this test vector makes sense. It makes sense to test this aspect of the code but this test doesn't do it since this test will just fail when you check of the decrypted signature is the same as the one in the JSON.

test/ecdsa_adaptor.json
LLFourn: I think you're right. We could provide some separate signing vectors as well as a test to check decryption on its own,

NicolasDorier:
> I think breaking changes here are okay/good to have, but I will note that I do not plan to actually make any other part of the DLC spec or implementations conform to this new ECDSA adaptor spec until there is a stable branch on secp256k1-zkp

In this case, this spec should conform to the current spec, then we should just change this spec and the old spec later.
If we merge this we start having different spec inside the same repo, this is confusing as hell.

LLFourn:
@NicolasDorier I think there's no reason why @nkohen couldn't push changes to the other spec tests to this branch so they are both merged at the same time.

jesseposner (approved):

NicolasDorier:
I am pausing work on DLC until the protocol is settling, this is quite depressing to write code, take time to follow test vectors, only to get them regenerated to follow a different spec that does the same thing later on, it feels like running in a treadmill. Wake me up when things stop moving and the test vectors stop moving.

test/ecdsa_adaptor.json
LLFourn: There are now separate "recovery" tests for this.

test/ecdsa_adaptor.json
LLFourn: On second thought I think this is not worth it on its own -- if your signing passes the verification then it will work. The only security concern then is nonce generation which as I want to leave out of scope for this scheme.

LLFourn:
@jesseposner I've updated the spec to just use the tag "DLEQ". I've also added another kind of test vector that just tests recovery. Let me know when you've done the implementation :)

ECDSA-adaptor.md
```diff
@@ -0,0 +1,208 @@
...
+
+## Specification tests
+
+The test vectors for this specification are in [./test/ecdsa_adaptor.json] and come in two kinds:
+
+- `verification`: These test the verifier in three stages:
+  - Check `adaptor_sig` passes `ecdsa_adaptor_verify`.
+  - Check that `ecdsa_adaptor_decrypt` yields `signature`.
+  - Check that  `ecdsa_adaptor_recover` recovers `decryption_key`.
+- `recovery`: These test only the key recovery function. The recovery function should recovery the correct decryption key.
```
jesseposner:
```suggestion
- `recovery`: These test only the key recovery function. The recovery function should recover the correct decryption key.
```

jesseposner:
@LLFourn I've updated https://github.com/ElementsProject/secp256k1-zkp/pull/117 with the new changes and test vectors.

LLFourn:
@jesseposner Thanks. Will review soon.

jesseposner:
@LLFourn I'm going to be rebasing to clean up the commits soon in the PR, which should make it much easier to review as a whole (but more difficult to isolate the changes from the original branch).

jesseposner:
I've cleaned up the commits and it should be a lot easier to review now by proceeding commit by commit. I'm giving it one last round of polish today and then I will fill out the PR description and move it out of draft mode.

jesseposner:
@LLFourn The PR is now stable and out of draft mode. The only thing that should change at this point is updates to respond to comments and potentially adding some additional test cases.

jesseposner:
@LLFourn Any thoughts on including a [warning](https://github.com/ElementsProject/secp256k1-zkp/pull/117/files#diff-0bc5e1a03ce026e8fea9bfb91a5334cc545fbd7ba78ad83ae5489b52e4e48856R14) about the leakage of the DH key between the signing key and encryption key? I was thinking some guidance might be warranted here, for example suggesting that DH protocols not be used for these keys.

LLFourn:
This document is ready. As I said before I think it would be good if all spectests were updated to align with this one so all the documents are consistent.

test/ecdsa_adaptor.json
```diff
@@ -0,0 +1,83 @@
...
+        "comment": "plain valid adaptor signature"
+    },
+    {
+        "kind" : "verification",
+        "adaptor_sig" : "036035c89860ec62ad153f69b5b3077bcd08fbb0d28dc7f7f6df4a05cca35455be037043b63c56f6317d9928e8f91007335748c49824220db14ad10d80a5d00a9654af0996c1824c64c90b951bb2734aaecf78d4b36131a47238c3fa2ba25e2ced54255b06df696de1483c3767242a3728826e05f79e3981e12553355bba8a0131cd370e63e3da73106f638576a5aab0ea6d45c042574c0c8d0b14b8c7c01cfe9072",
+        "public_signing_key" : "035be5e9478209674a96e60f1f037f6176540fd001fa1d64694770c56a7709c42c",
+        "encryption_key" : "024eee18be9a5a5224000f916c80b393447989e7194bc0b0f1ad7a03369702bb51",
+        "decryption_key" : "db2debddb002473a001dd70b06f6c97bdcd1c46ba1001237fe0ee1aeffb2b6c4",
+        "signature" : "6035c89860ec62ad153f69b5b3077bcd08fbb0d28dc7f7f6df4a05cca35455be4ceacf921546c03dd1be596723ad1e7691bdac73d88cc36c421c5e7f08384305",
+        "message_hash": "8131e6f4b45754f2c90bd06688ceeabc0c45055460729928b4eecf11026a9e2d",
```
nkohen: message hash should be higher up for consistency with other tests

NicolasDorier:
A reminder that this should be merged (or not). I am waiting for the spec to be stable before continuing any work on DLC, so as long as this is in "maybe merge, maybe not", I am on the sideline. :/

PR #119, opened 2020-12-04
Tuesday December 8 2020 Meeting

nkohen:


Christewart:
Might be worth bringing up long term Taproot planning? It did get merged into bitcoin core, but there is still no activation plan AFAIK.

Christewart:
One other thing that might want to be mentioned: 

I we've at least internally talked about removing range events

https://github.com/discreetlogcontracts/dlcspecs/blob/master/Oracle.md#range

Should this be discussed?

PR #120, opened 2020-12-09
Specify Client-Side Oracle Validation

ariard:
A really raw WIP, I still need clarification on `oracle_info`/`oracle_announcement` formats. Also if we scope censorship-resistance of oracle-client communications.

### DIFF
--- a/Oracle-Validation.md
+++ b/Oracle-Validation.md
@@ -1,0 +1,103 @@
+# Client-Side Oracle Validation
+
+# Table of Contents
+
+* [Abstract](#abstract)
+* [General Requirements](#general-requirements)
+* [Oracle-Discovery](#oracle-discovery)
+* [Oracle-Validation](#oracle-validation)
+
+
+## Abstract
+
+The DLC protocol relies on trust-minimized oracles mapping real-world events to signatures, those ones
+then confidentially re-mapped by participants to a range of contract outcomes. A participant expected
+execution of the DLC is thus function of the oracle well-behaving as announced before event
+realization. In this current protocol version, oracle behavior is assumed to be trusted. This level
+of trust is expressed by the client oracle selection policy, and should faithfully reflect user
+oracle preferences.
+
+Oracle signature usage requires confidence that the associated `oracle_public_key` is owned by the
+correct remote subject (person, system or organization). The ownership is asserted by having the
+client verifying a signature which is bound to an identity anchor (domain name, LN node pubkey, ...).
+The evaluation of identity anchor correctness is beyond the scope of this protocol. A DLC client may
+use this identity anchor as a descriptor for a reputation system.
+
+# Oracle Discovery
+
+Actually, no oracle discovery protocol support is mandated. The following requirements are generic
+requirements, which should be swayed from only if protocol has higher built-in privacy mechanisms.
+
+## Requirements
+
+A node:
+* MUST persist its keyring across restart/shutdown.
+* SHOULD fetch any `oracle_public_key` advertised by a "good reputation" oracle.
+
+## Rationale
+
+A DLC client should desynchronize any oracle pubkey fetching from DLC contract setting, otherwise
+the fetched entity may deduce an usage of oracle upcoming events. By fetching any `oracle_public_key`
+ available, a DLC client is hindering better potential oracle usage among a wider anonymous set. 
+"Good-reputation" is enforced by the client oracle selection policy and its definition is beyond this
+current specification.
+
+# `oracle_announcement` Authentication
+
+[`oracle_announcement`](https://github.com/discreetlogcontracts/dlcspecs/blob/master/Oracle.md#oracle-announcements)  
+authentication happens at `offer_dlc` reception, thus verifying binding between an event and an
+oracle.
+
+## Requirements
+
+A node:
+- if the `oracle_public_key` is already a member of client oracle keyring:
+    - SHOULD NOT fetch the public key again
+- otherwise:
+    - SHOULD fetch the public key
+- if `oracle_public_key` has not been accepted by client oracle selection policy:
+    - MUST reject the `offer_dlc`
+- if `oracle_event` signature is not valid against the `oracle_public_ke`:
+    - MUST reject the `offer_dlc`
+- MAY pre-validate `oracle_event` signature against the `oracle_public_key` at the time of its publication
+
+## Rationale
+
+Fetching `oracle_public_key` at `offer_dlc` is a privacy leak vector for the message receiver.
+Assuming oracle keys fetching is done through some privacy-preserving channel, if the DLC 
+counterparty can trigger communications through this channel it might hinder DLC client anonymity
+toward the oracle.
+
+Rejecting announcement for which the `oracle_public_key` hasn't been accepted by client oracle
+selection policy and announcement hasn't been authenticated prevent usage of malicious oracles.
+Authenticating the event also prevents entering in a DLC position where the oracle never intends
+to make the attestation, and thus losing the timevalue of the collateral. Valid announcements
+are also a building block of fraud proofs.
+
+# `oracle_attestation` Authentication
+
+## Requirements
+
+A node:
+- if an announcement for this `event_id` hasn't been previously accepted OR announcement's `oracle_public_key` and attestation's `oracle_public_key` are not equal:
+   - MUST ignore the `oracle_attestation`
+- if the range of `signatures` is not valid against the nonces committed in the accepted announcement:
+   - MUST ignore the `oracle_attestation`
+
+## Rationale
+
+Rejecting unrequested `oracle_attestation` prevents attestations injection by an attacker to discover
+other oracles actually consumed by this DLC client. Accepting unrequested _valid_ `oracle_attestation`
+and broadcasting a CET transaction on the p2p network are observable behaviors revealing DLC presenve
+for the corresponding event and oracle.
+
+Authenticating the `oracle_attestation` against a previously accepted announcement prevent CPU DoS
+against a DLC client feeded with expensive-to-validate messages.
+
+# Authors
+
+Antoine Riard <antoine.riard@gmail.com>
+
+![Creative Commons License](https://i.creativecommons.org/l/by/4.0/88x31.png "License CC-BY")
+<br>
+This work is licensed under a [Creative Commons Attribution 4.0 International License](http://creativecommons.org/licenses/by/4.0/).


Oracle-Validation.md
```diff
@@ -0,0 +1,73 @@
...
+* [Abstract](#abstract)
+* [General Requirements](#general-requirements)
+* [Oracle-Discovery](#oracle-discovery)
+* [Oracle-Validation](#oracle-validation)
+
+
+## Abstract
+
+The DLC protocol relies on trust-minized oracles mapping real-world events to signatures, those ones
+then confidentialy re-mapped by participants to a range of contract outcomes. A participant expected
```
benthecarman: > confidentialy

confidentially

Oracle-Validation.md
```diff
@@ -0,0 +1,73 @@
...
+* MUST fetch oracle pubkey.
+* MUST fetch any oracle pubkey available.
+* MUST persist its keyring across restart/shutdown.
+
+## Rationale
+
+A DLC client should desynchronize any oracle pubkey fetching from DLC contract setting, otherwise
+the fetched entity may deduce an ongoing usage of oracle events. 
+
+A DLC client should fetch any oracle pubkey available to mask better its actual orage usage.
```
benthecarman: > orage

not sure what this is meant to be

Oracle-Validation.md
```diff
@@ -0,0 +1,73 @@
...
+Oracle event authentication happens at `offer_dlc` reception. It ensures the offered DLC event is
+effectively to be announced by a previously identified oracle.
+
+## Requirements
+
+A node:
+* MUST verify `oracle_info` signature against the claimed oracle pubkey
+* MUST verify `oracle_announcement` signature against the negotiated oracle pubkey
+
+TODO: does `oracle_info`/`oracle_announcement` contain oracle identifier/pubkey/signature ?
```
benthecarman: They currently do

Oracle-Validation.md
```diff
@@ -0,0 +1,73 @@
...
+Actually, no oracle discovery protocol support is mandated. The following requirements are generic
+requirements, which should be swayed from only if protocol has higher built-in privacy mechanisms.
+
+## Requirements
+
+A node:
+* MUST fetch oracle pubkey.
+* MUST fetch any oracle pubkey available.
+* MUST persist its keyring across restart/shutdown.
+
```
benthecarman, line 36: Should we also recommend a encrypted & authenticated channel so they have less of chance of being man-in-the-middle'd

benthecarman: Also should we add recommendations of where to obtain the oracle's keys? ie directly from oracle vs oracle explorer

Oracle-Validation.md
naumenkogs: orage coin good

Oracle-Validation.md
ariard: Yes I think rn we don't have specific communication protocols for oracle discovery, you might have
* a DNS seed (e.g BOLT 10) relying on trusting DNS servers
* a manual copy-paste from few oracle repositories (e.g 1ml.com)
* a oracle existence gossips network (e.g BOLT 7) with some attributes filtered by user preferences
* a hardcoded initial list provided by DLC client vendors
* ....

Even if you answer "What's a good oracle to me ?", you still have to find the list of such "good" oracles without them being hidden or unknown from you, or if you have some central repo avoiding this one becoming a bait ?

> Should we also recommend a encrypted & authenticated channel so they have less of chance of being man-in-the-middle'd

First, how do you ensure that the pubkey you're opening a E&A transport with is _really_ owned by the entity you expect to talk with ?

Oracle-Validation.md
LLFourn: From memory there is a public key outside the signed announcement in the p2p protocol.
I think forcing a human readable `oracle_id` like "myoracle.com" inside the announcement would be a good idea.

Oracle-Validation.md
```diff
@@ -0,0 +1,73 @@
...
+## Rationale
+
+Identifying oracle pubkey against an identity anchor prevents usage of event signer disallowed by
+the client oracle selection policy.
+
+Authenticating oracle event prevents time-value DoS where a DLC is entered relying on a known honest oracle which doesn't have committed to the DLC event signing.
+
+
+TODO: sybil attack where honest connections to Olivia are censored until `refund_tx` is valid ?
```
LLFourn: What aspect of this is a sybill attack?

ariard:
Finally updated this PR, incorporating new `oracle_announcement`/`oracle_attestation`. 

I think this [comment](https://github.com/discreetlogcontracts/dlcspecs/pull/120#discussion_r538986097) merits its own meeting discussion, how do we make oracle discovery accessible is a really good question.

Oracle-Validation.md
Tibo-lg: > Should we also recommend a encrypted & authenticated channel so they have less of chance of being man-in-the-middle'd

I don't see the point of an authenticated channel to receive oracle information if the oracle announcements (and attestations) are themselves authenticated.

 > Also should we add recommendations of where to obtain the oracle's keys? ie directly from oracle vs oracle explorer

I think it's up to the users who they trust more, and I don't really see a technical rational for choosing one over the other? (and thus I don't think we should recommend anything, at least at this stage).

Oracle-Validation.md
```diff
@@ -0,0 +1,100 @@
...
+
+* [Abstract](#abstract)
+* [General Requirements](#general-requirements)
+* [Oracle-Discovery](#oracle-discovery)
+* [Oracle-Validation](#oracle-validation)
+
+
+## Abstract
+
+The DLC protocol relies on trust-minized oracles mapping real-world events to signatures, those ones
```
nkohen: > trust-minized

trust-minimized

Oracle-Validation.md
```diff
@@ -0,0 +1,100 @@
...
+
+Rejecting unrequested `oracle_attestation` prevents attestations injection by an attacker to discover
+other oracles actually consumed by this DLC client. Accepting unrequested _valid_ `oracle_attestation`
+and broadcasting a CET transaction on the p2p network are observable behaviors revealing DLC presenve
+for the corresponding event and oracle.
+
+Authenticating the `oracle_attestation` against a previously accepted announcement prevent CPU DoS
+against a DLC client feeded with expensive-to-validate messages.
+
+TODO: `oracle_attestation` should be signed against `oracle_public_key` to prevent forging of invalid `oracle_attestation` by an attacker ?
```
nkohen: Not sure what this means? Aren't `oracle_attestation`s unforgeable?

Oracle-Validation.md
```diff
@@ -0,0 +1,100 @@
...
+
+# Authors
+
+Antoine Riard <antoine.riard@gmail.com>
+
+TODO: License
```
nkohen: ping

Oracle-Validation.md
ariard: Depends if oracle discovery is happening through any kind of p2p mechanisms. If you learn oracle through a DNS record and the server cache is poisoned do you consider this as a sybil attack ? But yes better to call this oracle censorship, which is beyond the scope of this PR.

Oracle-Validation.md
ariard: Olivia announces event_id=XYZ with pubkey.Olivia. It's committed by Alice and Bob in their DLC. Mallory, a malicious oracle is connected to Alice for an unrelated event_id=ABC. XYZ is a public event announcement, so Mallory _might_ guess that Alice consumes this oracle. 

When event XYZ maturity comes,  Mallory feeds Alice with an attestation {event_id=XYZ, oracle_public_key=Olivia, sigs=invalid}. Is there a risk of  triggering Alice to severe her connection for faulty sigs or return a specific error to Mallory leaking event_id=XYZ usage ?

Oracle-Validation.md
ariard: > I don't see the point of an authenticated channel to receive oracle information if the oracle > announcements (and attestations) are themselves authenticated.

It _might_ be good to heartbeat your connections with a oracle to detect any kind of oracle censorship attempt. Wouldn't achieve its effect with unauthenticated channel. Though such heartbeat might not be great to preserve your privacy toward the oracle, trade-off...

> I think it's up to the users who they trust more, and I don't really see a technical rational for choosing one over the other? (and thus I don't think we should recommend anything, at least at this stage).

Once we have lazyness/fraud proofs we might have some kind of scoring and selection engine? But overall agree too early for now..

ariard:
Addressed review comments or answered at f335dc1

Yet-to-be-solved [comment](https://github.com/discreetlogcontracts/dlcspecs/pull/120#discussion_r585996737).

Oracle-Validation.md
LLFourn: In the sense that only Olivia can create a valid attestation under her announcement -- yes this is a requirement generally. Any invalid attestation should just be ignored. I don't *think* there is a need to sign the attestation.

ariard: Just _ignoring_ is a good behavior, otherwise it might be exploited in weird ways. Will add a "if invalid, MUST ignore". I believe CPU DoS can't really be avoided...

Oracle-Validation.md
Tibo-lg: > It might be good to heartbeat your connections with a oracle to detect any kind of oracle censorship attempt. Wouldn't achieve its effect with unauthenticated channel. Though such heartbeat might not be great to preserve your privacy toward the oracle, trade-off...

Yeah not great for privacy, and I don't really see a big upside. If you get censored there is not much you can do I think? And if there is something you can do, you should be able to do it even if you detect it only at maturation time if you setup a long enough refund time lock.

> Once we have lazyness/fraud proofs we might have some kind of scoring and selection engine? But overall agree too early for now..

Yes we might. But still I personally feel it will be up to the users which scoring or selection engine they'd want to chose, or if they just trust an oracle more than those. I think what I'm saying is that such considerations might not have their place in the specifications.

ariard: >  If you get censored there is not much you can do I think? And if there is something you can do, you should be able to do it even if you detect it only at maturation time if you setup a long enough refund time lock.

First we should dissociate oracle censorship from bitcoin p2p censorship. They may overlap or not. Secondly, you might have redundant communication channels, only enabled in case of emergency (including worst-case scenario sending a mail to a human operator to get involved). And if it's a tx-relay censorship, it might be hard to dissociate from a congestioned mempool...

That said, no a worry to address those issues now.

ariard:
Updated at d0f377c, requiring to ignore invalid `oracle_attestations`.

Ready for a new round of reviews.

Oracle-Validation.md
```diff
@@ -0,0 +1,100 @@
...
+   - MUST ignore the `oracle_attestation`
+- if the range of `signatures` is not valid against the nonces committed in the accepted announcement
+   - MUST ignore the `oracle_attestation`
+
+## Rationale
+
+Rejecting unrequested `oracle_attestation` prevents attestations injection by an attacker to discover
+other oracles actually consumed by this DLC client. Accepting unrequested _valid_ `oracle_attestation`
+and broadcasting a CET transaction on the p2p network are observable behaviors revealing DLC presenve
+for the corresponding event and oracle.
```
Tibo-lg, line 92: I don't really follow this part. Do you mean an unrequested `oracle_attestation` but for which you actually have a DLC? In that case I'm not sure how we could propagate attestation in a P2P setting.

ariard: It's all depend of we propagate messages in this p2p settings. 

If it's something similar to tx-relay on the base layer, you do have an inventory-getdata-tx messages sequence. If the inventory is already known by receiver you won't emit a getdata to save on bandwidth/CPU.

If it's something like Lightning gossips, without any announcement, effectively this requirement would be a blocker to process attestations. 

So yeah I agree, I think that's a small mitigation against DoS assuming our p2p network fits the first model. Likely to early as a reasoning :sweat_smile:

Tibo-lg: Thanks for the clarification :)

nkohen (approved):
This will require some additional language once we merge the oracle changes (separate announcement and attestation keys, signatures from nonce keys which need to be validated) but lgtm for now

Tibo-lg (approved):
I think it will have to be revisited later but it's a good basis for iterating.

Oracle-Validation.md
```diff
@@ -0,0 +1,98 @@
...
+# `oracle_announcement` Authentication
+
+[`oracle_announcement`](https://github.com/discreetlogcontracts/dlcspecs/blob/master/Oracle.md#oracle-announcements)  
+authentication happens at `offer_dlc` reception, thus verifying binding between an event and an
+oracle.
+
+## Requirements
+
+A node:
+- SHOULD NOT fetch `oracle_public_key` if it's not already a member of client oracle keyring
```
LLFourn: This double negative is wrong. Second "not" shouldn't bet there right?

Oracle-Validation.md
```diff
@@ -0,0 +1,98 @@
...
+
+Fetching `oracle_public_key` at `offer_dlc` is a privacy leak vector for the message receiver.
+Assuming oracle keys fetching is done through some privacy-preserving channel, if the DLC 
+counterparty can trigger communications through this channel it might hinder DLC client anonymity
+toward the oracle.
+
+Rejecting announcement for which the `oracle_public_key` hasn't been accepted by client oracle
+selection policy and announcement hasn't been authenticated prevent usage of lazy or malicious
+oracles. Authenticating event also prevent entering in a DLC position for which the event won't
+be signed, and thus losing timevalue of collaterals.
```
LLFourn: Style improvements.
```suggestion
oracles. Authenticating the event also prevents entering a DLC position where the oracle
never intends to make the attestation, and thus losing the timevalue of the collateral.
```

Oracle-Validation.md
```diff
@@ -0,0 +1,98 @@
...
+## Rationale
+
+Fetching `oracle_public_key` at `offer_dlc` is a privacy leak vector for the message receiver.
+Assuming oracle keys fetching is done through some privacy-preserving channel, if the DLC 
+counterparty can trigger communications through this channel it might hinder DLC client anonymity
+toward the oracle.
+
+Rejecting announcement for which the `oracle_public_key` hasn't been accepted by client oracle
+selection policy and announcement hasn't been authenticated prevent usage of lazy or malicious
+oracles. Authenticating event also prevent entering in a DLC position for which the event won't
```
LLFourn: what is a lazy oracle and how is it different to a malicious oracle?

ariard:
@nkohen Let me know what needs to get done on this PR. I think Lloyd is okay to land it and integrate expected changes around attestation/announcement keys later.

LLFourn (commented):
This will need to be revised very soon with oracle key changes and attestation changes (concept of attestation signature will be gone). It might be better to revise it before merging. Leave it up to you guys.

Oracle-Validation.md
ariard: Good catch!

Oracle-Validation.md
ariard: I think a malicious oracle might be qualified by an equivocation proof. A lazy would be one faulting to commit to a yet-to-be-designed-and-proposed-on-the-ml-publication space :) 

Note, I made the same point during my fraud point review that we should have better security definitions of notions like equivocation, invalidity, falseness and maybe lazyness. See https://github.com/discreetlogcontracts/dlcspecs/pull/152

Dropping for now we can revise later.

ariard:
Updated at a8cac0c integrating @LLFourn  small points.

LLFourn (approved):

ariard:
ping @nkohen :)

Oracle-Validation.md
```diff
@@ -0,0 +1,98 @@
...
+A DLC client should desynchronize any oracle pubkey fetching from DLC contract setting, otherwise
+the fetched entity may deduce an usage of oracle upcoming events. By fetching any `oracle_public_key`
+ available, a DLC client is hindering better potential oracle usage among a wider anonymous set. 
+"Good-reputation" is enforced by the client oracle selection policy and its definition is beyond this
+current specification.
+
+# `oracle_announcement` Authentication
+
+[`oracle_announcement`](https://github.com/discreetlogcontracts/dlcspecs/blob/master/Oracle.md#oracle-announcements)  
+authentication happens at `offer_dlc` reception, thus verifying binding between an event and an
```
nkohen, line 48: Isn't it also authenticated before `offer_dlc` construction?

Oracle-Validation.md
```diff
@@ -0,0 +1,98 @@
...
+oracle.
+
+## Requirements
+
+A node:
+- SHOULD NOT fetch `oracle_public_key` if it's already a member of client oracle keyring
+- if `oracle_public_key` has not been accepted by client oracle selection policy:
+    - MUST reject the `offer_dlc`
```
nkohen, line 59: is this really a MUST and not a SHOULD? Above you say "SHOULD fetch any `oracle_public_key`  ..."

Oracle-Validation.md
```diff
@@ -0,0 +1,98 @@
...
+Fetching `oracle_public_key` at `offer_dlc` is a privacy leak vector for the message receiver.
+Assuming oracle keys fetching is done through some privacy-preserving channel, if the DLC 
+counterparty can trigger communications through this channel it might hinder DLC client anonymity
+toward the oracle.
+
+Rejecting announcement for which the `oracle_public_key` hasn't been accepted by client oracle
+selection policy and announcement hasn't been authenticated prevent usage of malicious oracles.
+Authenticating the event also prevents entering in a DLC position where the oracle never intends
+to make the attestation, and thus losing the timevalue of the collateral.
+
```
nkohen, line 76: probably also worth mentioning that valid announcements are required in generating fraud proofs in the case of fraud later on.

nkohen (commented):
Just a couple nits

Oracle-Validation.md
ariard: IIUC your question, your suggesting that a DLC client could pre-authenticated events as soon as they're publicly announced by an oracle but _before_ they're proposed as a contract by a counterparty ?

I think so, added a MAY likewise :
"MAY pre-validate `oracle_event` signature against the `oracle_public_key` at the time of its publication"

Beyond the really slight perf improvement, it might also avoid to fetch `oracle_public_key` on `offer_dlc` reception, that way preserve receiver anonymity towards the oracle.

Oracle-Validation.md
ariard: Above is a SHOULD NOT ?

If you already know the `oracle_public_key`, avoid to fetch it again to preserve anonymity towards the oracle. 

If you don't know the `oracle_public_key`, fetch it to authenticate the `oracle_announcement`. 

If this `oracle_public_key` has been blacklisted by your oracle selection policy, keep it as part of you keyring, otherwise you won't be able to reject offending `offer_dlc`.

If this `oracle_public_key` has been accepted by your oracle, accept the proposed `offer_dlc`.

Do you have another flow in mind ? Note, you might have a size issue on your keyring, but for now let's shrug about tricky DoS.

That's a good remark, clarified with new requirements.

ariard:
Thanks for the review, updated at 3f181f7 to address Nadav comments.

PR #123, opened 2020-12-17
Enforce size prefixes for enumeration messages

Tibo-lg:
Not having size prefixes in some of the messages makes it difficult (for me at least) too reuse LN code that parses TLVs without forwarding information about the messages size (AFAICT at least). It feels to me like having a prefix is a very small overhead, and that in any case it might be better to be consistent in doing that all the time instead of distinguishing between cases where it can be omitted and those where it cannot.

### DIFF
--- a/Messaging.md
+++ b/Messaging.md
@@ -93,13 +93,14 @@ This type contains information about a contracts outcomes and their correspondin
 
 1. type: 42768 (`contract_info_v0`)
 2. data:
+   * [`bigsize`:`nb_outcomes`]
    * [`sha256`:`outcome_1`]
    * [`u64`:`outcome_1_local_payout`]
    * ...
    * [`sha256`:`outcome_n`]
    * [`u64`:`outcome_n_local_payout`]
 
-This type of contract info is a simple enumeration where the value `n` is omitted from being explicitly included as it can be derived from the length field of the TLV.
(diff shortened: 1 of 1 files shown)


benthecarman (approved):

Messaging.md
```diff
@@ -93,13 +93,14 @@ This type contains information about a contracts outcomes and their correspondin
 
 1. type: 42768 (`contract_info_v0`)
 2. data:
+   * [`u16`:`nb_outcomes`]
```
benthecarman: UInt16 is probably not enough for a very large contract, it's worth using UInt32

Messaging.md
```diff
@@ -150,13 +151,14 @@ This type contains CET signatures and any necessary information linking the sign
 
 1. type: 42774 (`cet_adaptor_signatures_v0`)
 2. data:
+   * [`u16`:`nb_signatures`]
```
benthecarman: ditto

benthecarman (changes requested):

Messaging.md
Tibo-lg: updated to use `bigsize` (a bit more efficient if number of outcome is small)

nkohen (approved):

benthecarman (approved):
ACK

PR #126, opened 2020-12-22
Create `oracle_attestation` TLV

### LINKED ISSUE #116: Create OracleAttestmentTLV
Would be nice to have a TLV that is used for the oracle sending the user the signatures/attestments

benthecarman:
Closes #116

### DIFF
--- a/Oracle.md
+++ b/Oracle.md
@@ -23,7 +23,9 @@ This necessary information is committed to in a so-called [_event descriptor_](#
    - [Oracle events](#oracle-events)
       - [Version 0 `oracle_event`](#version-0-oracle_event)
 - [Oracle announcements](#oracle-announcements)
-      - [Version 0 `oracle_announcement`](#version-0-oracle_announcement)
+  - [Version 0 `oracle_announcement`](#version-0-oracle_announcement)
+- [Oracle Attestations](#oracle-attestations)
+  - [Version 0 `oracle_attestation`](#version-0-oracle_attestation)
 - [Signing Algorithm](#signing-algorithm)
 
 ## Event descriptor
@@ -197,6 +199,30 @@ The TLV serialization of oracle announcements is as follow.
    * [`oracle_event`:`oracle_event`]
 
 where `signature` is a Schnorr signature over a sha256 hash of the serialized `oracle_event`, using the tag `announcement/v0`.
+
+## Oracle Attestations
+
+After an event occurs, and the oracle creates signatures to attest the outcome, it needs to give them to users.
+An oracle can use an attestation tlv to give users this information.
+
+The TLV serialization of oracle attestations is as follows.
+
+#### Version 0 `oracle_attestation`
+
+1. type: 55400 (`oracle_attestation_v0`)
+2. data:
+    * [`string`:`event_id`]
+    * [`x_point`:`oracle_public_key`]
+    * [`u16`: `nb_signatures`]
+    * [`signature`:`signature_1`]
+    * ...
+    * [`signature`:`signature_n`]
+    * [`string`:`outcome_1`]
+    * ...
+    * [`string`:`outcome_n`]
+
+Where the signatures are ordered the same as the nonces in their original `oracle_event`.
+The outcomes should be the message signed, ordered the same as the signatures.
 
 ## Signing Algorithm
 


Oracle.md
```diff
@@ -198,6 +200,27 @@ The TLV serialization of oracle announcements is as follow.
...
+After an event occurs, and the oracle creates signatures to attest the outcome, it needs to give them to users.
+An oracle can use an attestation tlv to give users this information.
+
+The TLV serialization of oracle attestations is as follows.
+
+#### Version 0 `oracle_attestation`
+
+1. type: 55400 (`oracle_attestation_v0`)
+2. data:
+    * [`x_point`:`oracle_public_key`]
```
Tibo-lg, line 215: Personally I have something similar, but with two more things:
* the event id
* the list of values that were signed
I guess we can argue if these are necessary or not, but at least on my side it makes things much easier to handle.

benthecarman: These seem reasonable to me.

Oracle.md
```diff
@@ -198,6 +200,31 @@ The TLV serialization of oracle announcements is as follow.
...
+
+The TLV serialization of oracle attestations is as follows.
+
+#### Version 0 `oracle_attestation`
+
+1. type: 55400 (`oracle_attestation_v0`)
+2. data:
+    * [`string`:`event_id`]
+    * [`x_point`:`oracle_public_key`]
+    * [`u16`: `nb_attestations`]
```
Tibo-lg: nit: I find it confusing to call these attestations since that's already the name of the enclosing structure. What about just 'signature'? Also it might be ok to have a single `u16` for the number as there should be exactly the same number for each I guess.

Tibo-lg (approved):
LGTM

Oracle.md
```diff
@@ -198,6 +200,30 @@ The TLV serialization of oracle announcements is as follow.
...
+2. data:
+    * [`string`:`event_id`]
+    * [`x_point`:`oracle_public_key`]
+    * [`u16`: `nb_signatures`]
+    * [`signature`:`signature_1`]
+    * ...
+    * [`signature`:`signature_n`]
+    * [`string`:`outcome_1`]
+    * ...
+    * [`string`:`outcome_n`]
```
LLFourn, line 222: Why not just have the outcome as a single string? The respective messages for each signature can then just be inferred from the single outcome string. That seems cleaner to me so clients can figure out what happened without have to reconstruct it from the list of strings.

Tibo-lg: For me it feels cleaner to put exactly what the oracle signed without requiring work from client to reconstruct it, but I'm not strongly opposed to your idea. Also in the (maybe unlikely) case where the oracle signs multiple unrelated strings for a single event we could then not reuse that.

benthecarman: I agree with @Tibo-lg. This should be clearer and less prone to mistakes.

nkohen: I agree that this should be kept as is. TLV is a serialization format which is not meant for users to every really deal with directly, their software can handle the concatenation for them and this should be kept "protocol-level"

nkohen (approved):

rorp:
Should we add a field for a signing timestamp and a signature for the whole TLV?

nkohen:
> Should we add a field for a signing timestamp and a signature for the whole TLV?

@rorp could you elaborate on what this would provide as a benefit? It isn't immediately obvious to me why either would be useful.

rorp:
> > Should we add a field for a signing timestamp and a signature for the whole TLV?
> 
> @rorp could you elaborate on what this would provide as a benefit? It isn't immediately obvious to me why either would be useful.

Timestamps can be useful for audit, and signatures can used to prevent timestamps from forging.

nkohen:
but there's no way to validate that the timestamp is real, they can easily lie about it can't they?

rorp:
They can lie about the outcome as well, and if the outcome is timestamped it's easier to verify it. 

Also if they issue multiple attestations for the same event, timestamps can help the users to develop some meaningful prioritization strategy.

Tibo-lg:
> They can lie about the outcome as well, and if the outcome is timestamped it's easier to verify it.

I don't follow what is easier to verify with a timestamp?

> Also if they issue multiple attestations for the same event, timestamps can help the users to develop some meaningful prioritization strategy.

I also don't understand what kind of strategy could be developed using timestamps? If I have two signatures available from a single oracle, my strategy would be to use the one that gives me the highest payout regardless of the timestamp (and also exclude the oracle from future contracts :) ).

I could see the usefulness of a signature over the whole TLV as to guarantee that it hasn't been tampered with, although I'm not sure it would be worth it since one can just verify each signature individually (and check that they match the value included in the TLV).

benthecarman:
I agree with @Tibo-lg & @nkohen. I don't see how a timestamp or signature over the entire tlv will make any real benefit

PR #127, opened 2020-12-22
Fix uint16 -> u16 in Digit Decomp Descriptor

benthecarman:


Tibo-lg (approved):
LGTM

Though now I wonder if we should use `i16` and `i32` for signed numbers to be consistent?

nkohen (approved):

PR #128, opened 2020-12-30
Multi-Oracle Support

nkohen:
Built on top of #110

TODO:

- [x] Target master when #110 is merged
- [x] Finish introductory and motivation sections
- [x] Add pictures to the 2-of-2 section
- [ ] Add links to my reference implementation
- [ ] Add examples
- [ ] Add test vectors (future PR)
- [ ] Offer/Accept integration
- [ ] Add Support for signed numeric outcomes


MultiOracle.md
```diff
@@ -0,0 +1,450 @@
...
+single-oracle DLCs.
+All work is done by DLC participants using existing DLC oracles as they are.
+
+This document begins by examining the case of enumerated outcome event DLCs, which can also be
+applied to numeric outcome DLCs in cases where all oracles are expected and required to produce exactly
+the same events.
+The majority of this document, however, is devoted to treating the numeric case where oracles are allowed
+to have bounded disagreements.
+
+In any of these cases, using multiple oracles does introduce a multiplier on total the number of CETs.
```
Tibo-lg: nit: "total the" -> "the total"

MultiOracle.md
```diff
@@ -0,0 +1,450 @@
...
+  * [n-of-n Oracles with Bounded Error](#n-of-n-oracles-with-bounded-error)
+    * [Analysis](#analysis)
+  * [t-of-n Oracles with Bounded Error](#t-of-n-oracles-with-bounded-error)
+* [Reference Implementations](#reference-implementations)
+* [Authors](#authors)
+
+## Multiple Oracles for Enumerated Outcome Events
+
+In the case of enumerated outcome events, where there is no macroscopic structure to the set of
+CETs and all oracles of a given event are expected to sign the same outcomes, then there is little
```
Tibo-lg: nit: "then there is little work to do" -> "little work is required"

MultiOracle.md
```diff
@@ -0,0 +1,450 @@
...
+This proposal can be thought of as defining the maximally coarse CET coverage algorithm which results in the fewest
+possible multiplier on the number of CETs.
+
+At a high level the algorithm works as follows (many details and some special cases are omitted).
+Designate one oracle in every t-of-t grouping to be the primary oracle and generate the set of CETs required for the DLC
+in question as would be done for a single oracle.
+This primary oracle functionally determines the execution price, meaning that the UX should be the same as if this
+primary oracle was used on its own.
+Then, for each of these CETs representing the primary oracle's attested result, label the first and last outcome covered
+by this CET as `[start, end]`  and compute the largest allowed range `[end - maxError, start + maxError]` for
```
Tibo-lg: I don't know if it's an error or if I'm just not understanding, but aren't the end and start inverted in the second statement?

MultiOracle.md
```diff
@@ -0,0 +1,450 @@
...
+secondary oracles.
+Compute the set of CETs required to cover this range (using the [CET compression algorithm](CETCompression.md#cet-compression)) and discard all but the
+largest CETs in this set, which by definition cover the vast majority of the range.
+This discarding is the primary opinion decidedly made from the design space by this proposal, and discarding any less
+leads to better guarantees at the expense of more CETs.
+The resulting CET(s) can then either be used (in the case of maximize coverage) or shrunk until they cover as little as
+possible while still containing `[start - minFail, end + minFail]` by repeatedly cutting them in half on either side.
+This process results in 1-3 secondary oracle CETs for ever single CET of the primary oracle.
```
Tibo-lg: What do you mean by "1-3 secondary oracle"? Also ever -> every

MultiOracle.md
```diff
@@ -0,0 +1,450 @@
...
+
+Let `[start, end] = computeCETIntervalBinary(cetDigits, numDigits)`,
+let `halfMaxError = 2^(maxErrorExp - 1)`,
+and let `maxNum = (2^numDigits) - 1`.
+
+We then split into cases:
+
+TODO: Pictures here
+
+* **Case** Small CET (`end - start + 1 < maxError`)
```
Tibo-lg: I think I mentioned already, but I'm not a big fan of this naming. It confuses me to use the term CET for what I think is outcome value. I see that of course there is a relation but it would IMHO be better to have distinct terms.

MultiOracle.md
Tibo-lg: Maybe change 1-3 to 1 to 3, and mention that it's just how the algorithm work.

Tibo-lg (commented):
Awesome work! Just some nits, and probably need to spend a bit more time on the algos.

MultiOracle.md
nkohen: So in the general case, `end - start < maxError` (because most CETs are small) but I agree this is very unclear here, will do better

MultiOracle.md
```diff
@@ -0,0 +1,450 @@
...
+This specification aims to significantly reduce oracle risk by introducing mechanisms with which to
+construct DLCs where some threshold of multiple oracles is required for successful execution.
+
+Using multiple oracles multiplies the cost of bribery and other attacks, while simultaneously offering
+protection against unresponsive and otherwise corrupted oracles.
+
+Using multiple oracles also provides a better environment for contract updates to a new oracle should
+one of the oracles being used announce that it will become unresponsive or has lost control of its keys.
+
+A key feature of this proposal is that no change in behavior is required of oracles  when compared to
```
benthecarman: nit: 

> oracles  when

there's an extra space here

MultiOracle.md
```diff
@@ -0,0 +1,450 @@
...
+
+For a more detailed rationale for this approach see [this section](CETCompression.md#adaptor-points-with-multiple-signatures)'s rationale.
+
+### t-of-n Oracles
+
+To construct a DLC for an enumerated outcome event using some threshold t-of-n oracles,
+we construct CETs for all outcomes for all possible combinations of t-of-t oracles.
+
+For example, say there are three oracles, Olivia, Oren, and Ollivander, who will be attesting
+two one of two outcomes, A or B.
```
benthecarman: > two one of two outcomes

_to_ one of two outcomes

MultiOracle.md
```diff
@@ -0,0 +1,450 @@
...
+* Olivia and Oren sign B.
+* Olivia and Ollivander sign B.
+* Oren and Ollivander sign B.
+
+In general, the resulting DLC will have as many CETs as it did in the single oracle case,
+with a multiplier of `n C t`.
+
+#### Example Algorithm
+
+```scala
```
benthecarman, line 106: Python!

MultiOracle.md
```diff
@@ -0,0 +1,450 @@
...
+The only additional work that must be done is that there must be a decision procedure to determine
+which oracle in a given group of `t` oracles is primary, and additionally what values should be used
+for `minFail` and `maxError` for a given group of `t` oracles (as this can vary between groups).
+
+The simplest candidate would be a negotiated ordered ranking of the oracles where the highest ranking
+oracle in any group of `t` oracles is chosen as primary.
+The values `minFail` and `maxError` can also be computed as some function of the rankings of a group
+of `t` oracles, or else they can be set as constant parameters across all groupings.
+
+TODO: Decide how this is done and communicated.
```
benthecarman, line 450: Isn't this done, just ranked by the order of them in the oracle info

benthecarman (changes requested):

MultiOracle.md
nkohen: I'm hoping that this will be made clear by the pictures I intend to add

MultiOracle.md
nkohen: That is how we currently do it in bitcoin-s but I'm open to the idea of doing something more general such as allowing the parameters to vary by choice of `t` oracles as mentioned in the previous paragraph

MultiOracle.md
```diff
@@ -0,0 +1,450 @@
+# Multiple Oracle Support
+
+## Introduction
+
+The Discreet Log Contract security model is entirely reliant on the existence of trustworthy oracles.
+However, even if an oracle appears trustworthy and has a good history of truth-telling, there is always
+some amount of risk associated with the oracle becoming unresponsive, or worse, corrupted.
+This risk is particularly worrisome for large value DLCs, as well as high volume DLCs where the pool
```
ariard: Didn't get it at first read that by high volume you mean "heavily used event outcome". 

For a more refined analysis, we should dissociate the per-event DLC pool from the per-oracle DLC pool. In the absence of multi-event DLC, the first one is a subset of the second one.

MultiOracle.md
```diff
@@ -0,0 +1,450 @@
...
+of funds that could profitably bribe or fund attacks on oracles is relatively larger.
+This specification aims to significantly reduce oracle risk by introducing mechanisms with which to
+construct DLCs where some threshold of multiple oracles is required for successful execution.
+
+Using multiple oracles multiplies the cost of bribery and other attacks, while simultaneously offering
+protection against unresponsive and otherwise corrupted oracles.
```
ariard, line 14: Do we dissociate unresponsiveness from lazyiness ? I don't think it's worthy for now but in the future you might have emergency communication channels with your oracle in case of primary one being down (BGP hijacks, network outage).

MultiOracle.md
```diff
@@ -0,0 +1,450 @@
...
+
+Using multiple oracles also provides a better environment for contract updates to a new oracle should
+one of the oracles being used announce that it will become unresponsive or has lost control of its keys.
```
ariard, line 17: Given oracle outcome event signing is public, it doesn't prevent formation of oracle coalitions. Should this be mentioned ?

MultiOracle.md
```diff
@@ -0,0 +1,450 @@
...
+This document begins by examining the case of enumerated outcome event DLCs, which can also be
+applied to numeric outcome DLCs in cases where all oracles are expected and required to produce exactly
+the same events.
+The majority of this document, however, is devoted to treating the numeric case where oracles are allowed
+to have bounded disagreements.
+
+In any of these cases, using multiple oracles does introduce a multiplier on the total number of CETs.
+In general, this multiplier grows exponentially with the number of oracles, but remains low enough
+for small numbers of oracles to enable the most common use cases (2-of-3 and 3-of-5) without much trouble.
+Larger numbers of oracles can still be used in cases where they are needed, but the CET cost incurred may be large.
```
ariard: Is the complexity of signing states equal to _n * m_, with _n_ the number of oracles and _m_ the number of outcomes. 

Also it's mostly a time cost, memory should be okay. I think it's still a matter of weeks for 12! or 13! states.

MultiOracle.md
```diff
@@ -0,0 +1,450 @@
...
+
+The n-of-n case is accomplished by simple point addition in the same manner as is done to combine
+multiple digit signatures in numeric outcome DLCs from digit decomposition.
+This specification opts to reduce the threshold (t-of-n) case to a combination of many t-of-t constructions.
+
+If two oracles for an enumerated outcome have similar, but not exactly equivalent enumerations of all
+possible outcomes, some care must be taken to ensure expected behavior.
+For example, if one weather oracle has only the events `["rainy", "sunny", "cloudy"]` while a second
+oracle has more events, `["rainy", "sunny", "partly cloudy", "cloudy"]`, then the DLC participants
+must decide what should happen in the case that the second oracle signs the message `"partly cloudy"`.
```
ariard, line 65: Do we assume this is out-of-band for now ? Maybe it could be better to specify the oracle-range approach as the default one. Identify a range supported by all oracles and outlaw any element outside as a refund case.

I'm worried manual negotiation with user involvement scales really well and maybe even footgunish.

MultiOracle.md
```diff
@@ -0,0 +1,450 @@
...
+They can either choose to not support that event (refund), or construct CETs for when the first party signs
+`"sunny"`, `"cloudy"`, or either (that is, two individual CETs).
+
+### n-of-n Oracles
+
+In much the same way as we create [adaptor points for oracles publishing multiple signatures](CETCompression.md#adaptor-points-with-multiple-signatures), we
+can similarly compute aggregate adaptor points for multiple oracles' signatures by simply using
+`s(1..n) * G = (s1 + s2 + ... + sn) * G = s1 * G + s2 * G + ... + sn * G` where
+`si` is the `i`th oracle's signature scalar.
+When the oracles then broadcast their `n` signatures, the corresponding adaptor secret
```
ariard: Better "After each `i`th oracle has broadcast its `i` signatures" ? Otherwise you may think oracles are acting as a group signer.

MultiOracle.md
```diff
@@ -0,0 +1,450 @@
...
+Specifically distance is used so that outcomes are related if they are sufficiently near each other.
+
+### Design
+
+It turns out that there is an exceedingly large design space of approaches that can be taken to support multiple
+oracles with some acceptable variance.
+It seems to be generally true, however, that a trade-off exists between the precision of guarantees that a procedure
+can make, and the size of the multiplier on the number of CETs required for the DLC.
+This specification chooses to prioritize CET reduction at the expense of precise guarantees so as to more easily
+support more oracles, which should generally counteract some of the imprecision surrounding variance support.
```
ariard: I agree with the intuition, better to care about security of the DLC by increasing the number of oracles. If we do have users who really care about precise, maybe we could have microscopic structure but that would require oracle specific support ?

So maybe oracle could sign the same event but offering different structures ?

MultiOracle.md
```diff
@@ -0,0 +1,450 @@
...
+Although this resulting design is extremely opinionated in some sense, it does leave three parameters unspecified
+to be negotiated or chosen by DLC participants.
+
+The first two parameters are the orders of magnitude of expected and of allowed difference between oracles.
```
ariard, line 181: You should indicate the difference unit, integers, unsigned, floats. Unless it's assumed by the spec ?

MultiOracle.md
```diff
@@ -0,0 +1,450 @@
...
+By expected difference we mean the differences for which support is required by users, and by allowed difference
+we mean the differences for which support is acceptable but not required.
```
ariard: Can you precise what do you mean by "acceptable but not required" ? Does it mean if implementation A allows the difference but it's outside of support requirement of implementation B, you do have a DLC failure ?

MultiOracle.md
```diff
@@ -0,0 +1,450 @@
...
+greater than `maxError` are supported, there are no guarantees made for differences in between these two parameters.
+Second, that `minFail` and `maxError` are required to be powers of `2` and not arbitrary numbers so that in reality, allowed
+variance is only expressed in terms of order of magnitude, which is a somewhat coarse measure.
+These two things together also imply that `minFail` must be at most half of `maxError` so that the gap between `minFail`
+and `maxError` where no strict guarantees can be made is of significant size.
+
+The third parameter offers a minimal remedy to these lack of guarantees.
+It is a boolean flag (`true` or `false`) called `maximizeCoverage`.
+This parameter arises from the fact that there is a decision to be made when covering the `minFail` variance to
+either cover as much as possible (up to `maxError`) or as little as possible, and this decision has no consequence
```
ariard, line 198: `minFail` is less or equal OR less. Unclear by using "up to `maxError`".

ariard (commented):
Stellar work :) Just a high-level parse for now!

MultiOracle.md
nkohen: I agree with you but this is likely out of scope for this document and should be included elsewhere

MultiOracle.md
nkohen: not quite sure what you mean by this?

MultiOracle.md
nkohen: No the complexity is much worse than linear as it is exponential

MultiOracle.md
nkohen: The language here is intentionally vague and will be changed after this discussion is resolved: https://mailmanlists.org/pipermail/dlc-dev/2021-February/000024.html

MultiOracle.md
nkohen: So actually I don't believe any changes are required oracle-side to support more precise bounds, it would be an entirely client-side procedure, but the number of CET's required is pretty immense...

MultiOracle.md
nkohen: This spec assumes the same units and such as are specified in the `oracle_event`

MultiOracle.md
nkohen: Not sure what you mean

MultiOracle.md
```diff
@@ -0,0 +1,461 @@
...
+should be noted that these are still only probabilistic as even if `maximizeCoverage` is set to `true`, there can still
+be differences of `minSupport + 1` which are unsupported in the worst case and likewise, differences of `maxError - 1`
+may still succeed even if `maximizeCoverage` is set to `false` in the worst case.
+Yet in the average case this maximizing coverage will make it much more likely that cases in the `maxError`-`minSupport` gap
+will succeed and minimizing coverage will make it much more likely that these cases fail (are not supported).
+
+Here is a diagram which illustrates how the error bounds `minSupport` and `maxError` are used to constrain the selection
+of secondary oracle intervals:
+
+![image](images/SecondaryDLCOracle.png)
```
Tibo-lg: nit: This png really doesn't work with my dark mode setting

MultiOracle.md
```diff
@@ -0,0 +1,461 @@
...
+This proposal can be thought of as defining the maximally coarse coverage algorithm which results in the fewest
+possible multiplier on the number of adaptor signatures.
+
+At a high level the algorithm works as follows (many details and some special cases are omitted).
+Designate one oracle in every t-of-t grouping to be the primary oracle and generate the set of digit prefixes and corresponding
+adaptor points required for the DLC in question as would be done for a single oracle.
+This primary oracle functionally determines the execution price, meaning that the UX should be the same as if this
+primary oracle was used on its own.
+Then, for each of these digit prefixes representing the primary oracle's attested result, label the first and last outcome covered
+by this digit prefix as `[start, end]`  and compute the largest allowed range `[end - maxError, start + maxError]` for
```
Tibo-lg: I think the picture makes it clearer, but I still feel it might be nice to add the explanation that you gave me last time (that in general [end - start] is smaller than maxError) or maybe explicitly refer to the figure.

Tibo-lg (commented):
Really tiny nits. I think it's much clearer without using CET.

MultiOracle.md
nkohen: Any suggestions? (directed generally towards everyone, not just Tibo)

matthewjablack: Maybe just add a white background for each image ⬇️

SecondaryDLCOracle
SmallPrimaryInterval
LargePrimaryInterval

nkohen: I thought about doing this but it does make the images less portable :/ maybe the best bet is to have both versions on the spec repo and reference the white background version in the spec itself but where people can still use the version without the background

matthewjablack: If you use an svg you might be able to use css to specify dark vs light theme: https://github.community/t/support-theme-context-for-images-in-light-vs-dark-mode/147981/17

MultiOracle.md
```diff
@@ -0,0 +1,462 @@
...
+to one of two outcomes, A or B.
+If we wish to construct a 2-of-3 DLC over these outcomes using these oracles, we will construct
+a total of 6 adaptor signatures:
+
+* Olivia and Oren sign A.
+* Olivia and Ollivander sign A.
+* Oren and Ollivander sign A.
+* Olivia and Oren sign B.
+* Olivia and Ollivander sign B.
+* Oren and Ollivander sign B.
```
matthewjablack, line 99: I'm assuming each of these 6 cases includes a Primary and Secondary oracle. Perhaps this should be specified here.

MultiOracle.md
```diff
@@ -0,0 +1,462 @@
...
+* `numToVec` takes as input an integer representing the left endpoint of a digit prefix's range, the number
+  of binary digits to be signed by the oracle, and a number of digits to ignore/omit, and returns
+  the binary digits corresponding to that digit prefix.
+
+We take as input the number of digits to be signed by the oracle, `numDigits`, the binary digits of the primary
+oracle's digit prefix which we wish to cover with the secondary oracle, `primaryDigits`, `maxErrorExp`, `minSupportExp`,
+and a boolean flag `maximizeCoverage` which signals whether to maximize the probability of success for
+outcomes differing by less than `maxError` and more than `minSupport` or not (in which case failure is maximized).
+
+Let `[start, end] = computePrefixIntervalBinary(primaryDigits, numDigits)`,
```
matthewjablack, line 279: Perhaps provide an example of `primaryDigits`

MultiOracle.md
matthewjablack: I suppose order-ranked oracle info could be added as a `negotiation_field` to `dlc_accept` message and the `initiator` could choose to either sign or not.

matthewjablack (commented):
Just a few points for clarification

nkohen:
> I suppose order-ranked oracle info could be added as a negotiation_field to dlc_accept message and the initiator could choose to either sign or not.

@matthewjablack Currently it works the other way where the offerer sets a preference ranking (via the oracle_info of the contract_info) and the accepter can choose not to accept. Open to other ideas but I don't think adding a negotiation field makes sense in cases of actual disagreement (unlike rounding where it can make sense to merge both party's preferences)

MultiOracle.md
nkohen: No actually that is not the case here, there is only such distinctions as "primary" and "secondary" oracles in the numeric with differences setting as in all other places all oracles agree so there is no use for such a distinction.

MultiOracle.md
nkohen: I think what might cover this without requiring an abrupt change of direction is a link to the reference implementation (which will be added before merging this PR) and may a "see examples here" with a link to test vectors which will be example inputs to `computePrefixIntervalBinary` with their corresponding outputs.

Thoughts?

MultiOracle.md
nkohen: I ended up doing what I thought of doing here ^

MultiOracle.md
matthewjablack: I think that would do it. Perhaps there should also be a link to the [CET Compression concrete example](https://github.com/discreetlogcontracts/dlcspecs/blob/898e81a23c30ac4bc66209b20697e76e4a1da7bd/CETCompression.md#concrete-example).

matthewjablack:
> @matthewjablack Currently it works the other way where the offerer sets a preference ranking (via the oracle_info of the contract_info) and the accepter can choose not to accept.

@nkohen On second thought, this makes a lot more sense. Better not to complicate messaging protocol with negotiation fields.

MultiOracle.md
ariard: Let's define a oracle coalition as a set of oracles colluding on a event A outcome to influence market behaviors.

Olivia is committing to real-word event X with maturing time Y. Then she is browsing through the recent announcements log until she finds event X is also committed by Oliver. She reaches out to him and propose to enter into a coalition to sign X=Y when the most expected outcome is X=Z. Meanwhile, they buy a lot of DLC puts betting on X=Y thus making a lot of profits from the manipulation.

Those profits may overpass the loss of reputation and as such not fully-mitigable by fraud proofs.

Should we just open an issue to document "Oracle coalitions and Harmful Market Behaviors", that's long-term concerns for now ?

MultiOracle.md
```diff
@@ -0,0 +1,462 @@
...
+This document begins by examining the case of enumerated outcome event DLCs, which can also be
+applied to numeric outcome DLCs in cases where all oracles are expected and required to produce exactly
+the same events.
+The majority of this document, however, is devoted to treating the numeric case where oracles are allowed
+to have bounded disagreements.
+
+In any of these cases, using multiple oracles does introduce a multiplier on the total number of adaptor signatures needed.
+In general, the adaptor signature set grows exponentially with the number of oracles, but remains small enough
+for small numbers of oracles to enable the most common use cases (2-of-3 and 3-of-5) without much trouble.
+Larger numbers of oracles can still be used in cases where they are needed, but the adaptor signature cost incurred may be large.
```
ariard, line 32: Though may precise it's a signature _computation_ cost increase, not an _onchain footprint_ one.

MultiOracle.md
```diff
@@ -0,0 +1,462 @@
...
+some amount of risk associated with the oracle becoming unresponsive, or worse, corrupted.
+This risk is particularly worrisome for large value DLCs, as well as widely used DLC oracle events where
+the pool of funds that could profitably bribe or fund attacks on oracles is relatively larger.
+This specification aims to significantly reduce oracle risk by introducing mechanisms with which to
+construct DLCs where some threshold of multiple oracles is required for successful execution.
+
+Using multiple oracles multiplies the cost of bribery and other attacks, while simultaneously offering
+protection against unresponsive and otherwise corrupted oracles.
+
+Using multiple oracles also provides a better environment for contract updates to a new oracle should
```
ariard, line 16: A trade-off not documented here is the lower fault-tolerance of your N-of-M DLC contract itself. If you're using a high N or N=M, one or few unresponsive/corrupted oracles are enough to halt your contract ?

"One of the limitation of multiple oracles is a decreased fault-tolerance of the contract if the threshold is high".

MultiOracle.md
ariard: I've been through the thread, w.r.t to enumerated event, I would prefer option 2.

I think the meta-enum have 2 issues :
* imply human _cognition_ for mapping oracle's events disjoint from the best identity set among all oracle events. I.e mapping non-ambiguously "drizzly" to "rainy" or "precipations" if both elements are present in the identity set
* compatibility with _compound_ events where you're creating DLC as a combination of oil price and the latest NBA game, there is not even a identity map that you have to successively extend for ambiguous cases

I think the list approach doesn't suffer of those issues and can be fully automated ? Further, I'm expecting more oracles to pre-coordinate among themselves to present the same event structure rather than having the complexity client-side.

MultiOracle.md
ariard: `minFail` might be <= to `maxError` or only `<` to `maxError` ?

MultiOracle.md
```diff
@@ -0,0 +1,462 @@
...
+This specification chooses to prioritize adaptor signature reduction at the expense of precise guarantees so as to more
+easily support more oracles, which should generally counteract some of the imprecision surrounding variance support.
+Although this resulting design is extremely opinionated in some sense, it does leave three parameters unspecified
+to be negotiated or chosen by DLC participants.
+
+The first two parameters are the orders of magnitude of expected and of allowed difference between oracles.
+By expected difference we mean the differences for which support is required by users, and by allowed difference
+we mean the differences for which support is acceptable but not required so that they may or may not be covered
+by the multi-oracle DLC construction algorithm.
+These parameters will, from now on, be called `minSupport` and `maxError` as they also represent the lower bound
```
ariard, line 185: I'm confused by this part though I get the idea by looking on the illustration.

AFAICT, you have a primary interval to which you can superset a secondary interval thanks to `minSupport` bounds. But how do you decide on a primary interval as a DLC client ?

Further, should we speak about a tertiary interval that you can build thanks to `maxError` ?

Overall I think this part of the specification would benefit from a split in at least 3 sections: 
* "multi-oracle numeric range" data struct fields definition
* requirements on data struct as enforced by DLC counterparties
* design rational

I can take it latter if you wanna but I feel it's needed for this part of the spec to age well :)

ariard (commented):
I'm currently confused by the "Multiple Oracles for Numeric Outcome Events" design section I left few questions I would be glad to be answered before reviewing the algorithm and remaining parts.

PR #130, opened 2021-01-12
Generalized contract_info TLV

nkohen:
Updated TLVs to allow support for numeric outcome and multiple oracle DLCs.

### DIFF
--- a/Messaging.md
+++ b/Messaging.md
@@ -154,6 +154,7 @@ This type of oracle info is for single-oracle events.
 
 1. type: 42786 (`oracle_info_v1`)
 2. data:
+   * [`u16`:`threshold`]
    * [`u16`:`num_oracles`]
    * [`oracle_announcement`:`oracle_announcement_1`]
    * ...
@@ -166,6 +167,7 @@ that exactly correspond to the messages being signed by the other oracles.
 
 1. type: 55340 (`oracle_info_v2`)
 2. data:
+   * [`u16`:`threshold`]
    * [`u16`:`num_oracles`]
    * [`oracle_announcement`:`oracle_announcement_1`]
    * ...
--- a/Messaging.md
+++ b/Messaging.md
@@ -113,7 +113,7 @@ This type contains information about a contract's outcomes, their corresponding 
 
 This type contains information about a contract's outcomes and their corresponding payouts.
 
-To save space, only one side's POV is included in this message as the other can be derived using `remote_payout = total_collateral - local_payout`.
+To save space, only the offerer's POV is included in this message as the other can be derived using `accept_payout = total_collateral - offer_payout`.
 
 #### Version 0 `contract_descriptor`
 
@@ -138,6 +138,8 @@ This type represents an enumerated outcome contract.
 
 This type represents a numeric outcome contract.
 
+The type `rounding_intervals` is defined [here](NumericOutcome.md#rounding-interval-serialization).
+
 ### The `oracle_info` Type
 
 This type contains information about the oracles to be used in executing a DLC.
@@ -161,7 +163,8 @@ This type of oracle info is for single-oracle events.
    * [`oracle_announcment`:`oracle_announcement_num_oracles`]
 
 This type of oracle info is for multi-oracle events where all oracles are signing messages
-that exactly correspond to the messages being signed by the other oracles.
+that exactly correspond to the messages being signed by the other oracles, and `threshold`
+oracles must be in agreement for execution to happen.
 
 #### Version 2 `oracle_info`
 
@@ -176,12 +179,12 @@ that exactly correspond to the messages being signed by the other oracles.
 
 The order of the oracle announcements represents a total ordering of preference on the oracles.
 
-This type of oracle info is for multi-oracle numeric events where there is some allotted
-difference between oracles allowed and specified in `oracle_params`.
+This type of oracle info is for multi-oracle numeric events where allowed differences in the values
+signed by oracles is specified in `oracle_params`.
 
 ### The `oracle_params` Type
 
-This type contains information about the oracles being used beyond their announcements.
+contains information about how oracle information is used in a given contract.
 
 #### Version 0 `oracle_params` Type
 
@@ -191,8 +194,8 @@ This type contains information about the oracles being used beyond their announc
    * [`u16`:`minFailExp`]
    * [`bool`:`maximize_coverage`]
 
-This type is used when there is a shared error bounds requirement for any set of oracles in a
-multi-oracle numeric outcome DLC with allowed (bounded) error.
+This type is used when the error bound requirements for any set of oracles `threshold` oracles in a
+multi-oracle numeric outcome DLC with allowed error is the same.
 
 ### The `negotiation_fields` Type
 
--- a/Messaging.md
+++ b/Messaging.md
@@ -15,8 +15,18 @@ All data fields are unsigned big-endian unless otherwise specified.
   * [DLC Specific Types](#dlc-specific-types)
     * [The `contract_info` Type](#the-contract_info-type)
       * [Version 0 `contract_info`](#version-0-contract_info)
+    * [The `contract_descriptor` Type](#the-contract_descriptor-type)
+      * [Version 0 `contract_descriptor`](#version-0-contract_descriptor)
+      * [Version 1 `contract_descriptor`](#version-1-contract_descriptor)
     * [The `oracle_info` Type](#the-oracle_info-type)
       * [Version 0 `oracle_info`](#version-0-oracle_info)
+      * [Version 1 `oracle_info`](#version-1-oracle_info)
+      * [Version 2 `oracle_info`](#version-2-oracle_info)
+    * [The `oracle_params` Type](#the-oracle_params-type)
+      * [Version 0 `oracle_params`](#version-0-oracle_params)
+    * [The `negotiation_fields` Type](#the-negotiation_fields-type)
+      * [Version 0 `negotiation_fields`](#version-0-negotiation_fields)
+      * [Version 1 `negotiation_fields`](#version-1-negotiation_fields)
     * [The `funding_input` Type](#the-funding_input-type)
       * [Version 0 `funding_input`](#version-0-funding_input)
     * [The `cet_adaptor_signatures` Type](#the-cet_adaptor_signatures-type)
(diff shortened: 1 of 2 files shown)


Messaging.md
```diff
@@ -87,33 +97,123 @@ The following DLC-specific types are used throughout the specification. All type
...
+#### Version 0 `oracle_params` Type
+
+1. type: 55338 (`oracle_params_v0`)
+2. data
+   * [`u16`:`maxErrorExp`]
+   * [`u16`:`minFailExp`]
+   * [`bool`:`maximize_coverage`]
+
+This type is used when there is a shared error bounds requirement for any set of oracles in a
+multi-oracle numeric outcome DLC with allowed (bounded) error.
```
benthecarman: This is probably waiting on #128 but should add a link to explain what this is / how it is used

Messaging.md
```diff
@@ -87,33 +97,123 @@ The following DLC-specific types are used throughout the specification. All type
...
+This type represents an enumerated outcome contract.
+
+#### Version 1 `contract_descriptor`
 
-This type of contract info is a simple enumeration of pairs of the hash of an outcome value with the corresponding payout for the local party.
+1. type: 42784 (`contract_descriptor_v1`)
+2. data:
+   * [`u16`:`num_digits`]
+   * [`payout_function`:`payout_function`]
+   * [`rounding_intervals`:`rounding_intervals`]
```
benthecarman, line 138: `rounding_intervals` serialization is never defined

benthecarman (changes requested):

Messaging.md
```diff
@@ -87,33 +97,123 @@ The following DLC-specific types are used throughout the specification. All type
 
 ### The `contract_info` Type
 
-This type contains information about a contracts outcomes and their corresponding payouts. To save space, only one side's POV is included in this message as the other can be derived using `remote_payout = total_collateral - local_payout`.
+This type contains information about a contract's outcomes, their corresponding payouts, and the oracles to be used.
 
 #### Version 0 `contract_info`
 
-1. type: 42768 (`contract_info_v0`)
+1. type: 55342 (`contract_info_v0`)
```
Tibo-lg, line 104: Just to make sure, we are keeping v0 but changing it is that ok? I guess it should be fine since we don't need to be backward compatible at that point but just wanted to make sure.

Messaging.md
```diff
@@ -87,33 +97,123 @@ The following DLC-specific types are used throughout the specification. All type
...
 2. data:
-   * [`bigsize`:`nb_outcomes`]
-   * [`sha256`:`outcome_1`]
-   * [`u64`:`outcome_1_local_payout`]
+   * [`u64`:`total_collateral`]
```
Tibo-lg, line 106: IIRC before you were against putting this to allow the accepting party to set the collateral they want to put, do we give up on that (I'm fine either way, just to understand).

Messaging.md
```diff
@@ -87,33 +97,123 @@ The following DLC-specific types are used throughout the specification. All type
...
+   * [`contract_descriptor`:`contract_descriptor`]
+   * [`oracle_info`:`oracle_info`]
+
+`total_collateral` is the Satoshi-denominated value of the sum of all party's collateral.
+
+### The `contract_descriptor` Type
+
+This type contains information about a contract's outcomes and their corresponding payouts.
+
+To save space, only one side's POV is included in this message as the other can be derived using `remote_payout = total_collateral - local_payout`.
```
Tibo-lg: should we switch to accept/offer?

Messaging.md
```diff
@@ -87,33 +97,123 @@ The following DLC-specific types are used throughout the specification. All type
...
+
+1. type: 42786 (`oracle_info_v1`)
+2. data:
+   * [`u16`:`threshold`]
+   * [`u16`:`num_oracles`]
+   * [`oracle_announcement`:`oracle_announcement_1`]
+   * ...
+   * [`oracle_announcment`:`oracle_announcement_num_oracles`]
+
+This type of oracle info is for multi-oracle events where all oracles are signing messages
```
Tibo-lg: suggestion: "where at least a `threshold` number of oracles are expected to sign the exact same outcome value."

Messaging.md
```diff
@@ -87,33 +97,123 @@ The following DLC-specific types are used throughout the specification. All type
...
+   * [`u16`:`threshold`]
+   * [`u16`:`num_oracles`]
+   * [`oracle_announcement`:`oracle_announcement_1`]
+   * ...
+   * [`oracle_announcment`:`oracle_announcement_num_oracles`]
+   * [`oracle_params`:`oracle_params`]
+
+The order of the oracle announcements represents a total ordering of preference on the oracles.
+
+This type of oracle info is for multi-oracle numeric events where there is some allotted
```
Tibo-lg: suggestion: "where differences in the value signed by oracles are allowed as specified in `oracle_params`" (at least allotted and allowed in the same sentence feels weird)

Messaging.md
```diff
@@ -87,33 +97,123 @@ The following DLC-specific types are used throughout the specification. All type
...
+difference between oracles allowed and specified in `oracle_params`.
+
+### The `oracle_params` Type
+
+This type contains information about the oracles being used beyond their announcements.
```
Tibo-lg: suggestion: "contains information about how oracle information is used in a given contract"

Messaging.md
```diff
@@ -87,33 +97,123 @@ The following DLC-specific types are used throughout the specification. All type
...
+
+#### Version 0 `oracle_params` Type
+
+1. type: 55338 (`oracle_params_v0`)
+2. data
+   * [`u16`:`maxErrorExp`]
+   * [`u16`:`minFailExp`]
+   * [`bool`:`maximize_coverage`]
+
+This type is used when there is a shared error bounds requirement for any set of oracles in a
```
Tibo-lg: what do you mean by "a shared error bound"?

Tibo-lg (commented):
Looks good, feel free too ignore my suggestions but I where I put them is in general because I feel the sentences are difficult to get if you don't already have a good understanding of how things are supposed to work (which is maybe fine, but it might also not hurt to make things a bit more intelligible IMHO).

Messaging.md
nkohen: yup, this is me with a white flag, at least for now

Messaging.md
nkohen: Yes, the old type is now a `contract_descriptor_v0` which is what the old contract info was at a byte level

Messaging.md
nkohen: I'm not sure I follow, this only currently works if all oracles are signing corresponding things?

Messaging.md
```diff
@@ -15,8 +15,18 @@ All data fields are unsigned big-endian unless otherwise specified.
...
       * [Version 0 `contract_info`](#version-0-contract_info)
+    * [The `contract_descriptor` Type](#the-contract_descriptor-type)
+      * [Version 0 `contract_descriptor`](#version-0-contract_descriptor)
+      * [Version 1 `contract_descriptor`](#version-1-contract_descriptor)
     * [The `oracle_info` Type](#the-oracle_info-type)
       * [Version 0 `oracle_info`](#version-0-oracle_info)
+      * [Version 1 `oracle_info`](#version-1-oracle_info)
+      * [Version 2 `oracle_info`](#version-2-oracle_info)
+    * [The `oracle_params` Type](#the-oracle_params-type)
+      * [Version 0 `oracle_params`](#version-0-oracle_params)
```
Tibo-lg, line 26: nit: this link doesn't seem to be working

Messaging.md
Tibo-lg: understood

Messaging.md
```diff
@@ -87,33 +97,126 @@ The following DLC-specific types are used throughout the specification. All type
...
+   * [`contract_descriptor`:`contract_descriptor`]
+   * [`oracle_info`:`oracle_info`]
+
+`total_collateral` is the Satoshi-denominated value of the sum of all party's collateral.
+
+### The `contract_descriptor` Type
+
+This type contains information about a contract's outcomes and their corresponding payouts.
+
+To save space, only the offerer's POV is included in this message as the other can be derived using `accept_payout = total_collateral - offer_payout`.
```
Tibo-lg: nit: I guess POV means Payout Value? It's not very clear to me, and it's never defined so might be better to just spell it out.

Messaging.md
```diff
@@ -87,33 +97,126 @@ The following DLC-specific types are used throughout the specification. All type
...
+
+1. type: 42784 (`contract_descriptor_v1`)
+2. data:
+   * [`u16`:`num_digits`]
+   * [`payout_function`:`payout_function`]
+   * [`rounding_intervals`:`rounding_intervals`]
+
+This type represents a numeric outcome contract.
+
+The type `rounding_intervals` is defined [here](NumericOutcome.md#rounding-interval-serialization).
```
Tibo-lg, line 143: nit: Maybe nice to link to payout_function as well?

Messaging.md
Tibo-lg: Yeah but IIUC it works if you have `threshold` oracles that sign the same value and n - threshold sign whatever? I was just trying to make the sentence more digestable but from your comment I guess I failed :D

Messaging.md
```diff
@@ -87,33 +97,126 @@ The following DLC-specific types are used throughout the specification. All type
...
+   * [`oracle_params`:`oracle_params`]
+
+The order of the oracle announcements represents a total ordering of preference on the oracles.
+
+This type of oracle info is for multi-oracle numeric events where allowed differences in the values
+signed by oracles is specified in `oracle_params`.
+
+### The `oracle_params` Type
+
+contains information about how oracle information is used in a given contract.
```
Tibo-lg: nit: capitalize?

Messaging.md
```diff
@@ -87,33 +97,126 @@ The following DLC-specific types are used throughout the specification. All type
...
+2. data:
+   * (empty)
+
+This type signifies that the accepter has no negotiation fields.
+
+#### Version 1 `negotiation_fields`
+
+1. type: 55336 (`negotiation_fields_v1`)
+2. data
+   * [`rounding_intervals`: `rounding_intervals`]
```
Tibo-lg, line 218: This is used both here and line 137 is that on purpose?

Protocol.md
```diff
@@ -165,6 +163,7 @@ and closing transactions.
    * [`spk`:`change_spk`]
    * [`cet_adaptor_signatures`:`cet_adaptor_signatures`]
    * [`signature`:`refund_signature`]
+   * [`negotiation_fields`:`negotiation_fields`]
```
Tibo-lg, line 166: Should the type definition be moved here? I feel it's becoming a bit unclear which types are defined here and which are defined in `Messaging.md`, do we have a clear rule for that?

Tibo-lg (commented):
Again mainly nits

Messaging.md
nkohen: Funny enough both point of view and payout value work here lol, will fix tho

Messaging.md
nkohen: Ah I think I understand, will do my best on this one

Messaging.md
nkohen: Yes, you'll recall that both offerer and accepter get to give their max rounding and then these are merged before being used in CET computation.

Protocol.md
nkohen: This is a good question, I don't have a good answer but someone should come up with one and open a PR to make things uniform. I've opened an issue (#141)

Messaging.md
Tibo-lg: Oh right!

Protocol.md
Tibo-lg: Perfect

Messaging.md
Tibo-lg: Reads much better thanks!

Tibo-lg (approved):

PR #136, opened 2021-02-06
Define input & output ordering

### LINKED ISSUE #18: Ordering of Inputs and Outputs in transactions

benthecarman:
Closes #18

### DIFF
--- a/Protocol.md
+++ b/Protocol.md
@@ -70,10 +70,13 @@ the funding transaction and CETs.
    * [`contract_info`:`contract_info`]
    * [`point`:`funding_pubkey`]
    * [`spk`:`payout_spk`]
+   * [`u64`:`payout_serial_id`]
    * [`u64`:`total_collateral_satoshis`]
    * [`u16`:`num_funding_inputs`]
    * [`num_funding_inputs*funding_input`:`funding_inputs`]
    * [`spk`:`change_spk`]
+   * [`u64`:`change_serial_id`]
+   * [`u64`:`fund_output_serial_id`]
    * [`u64`:`feerate_per_vb`]
    * [`u32`:`contract_maturity_bound`]
    * [`u32`:`contract_timeout`]
@@ -98,6 +101,15 @@ the sender and `funding_inputs` contains outputs, outpoints, and expected weight
 of the sender's funding inputs. `change_spk` specifies the script pubkey that funding
 change should be sent to.
 
+`payout_serial_id` is a randomly chosen number which uniquely identifies the payout output of the offering party.
+Outputs in the contract execution transaction will be sorted by `payout_serial_id`.
+
+`change_serial_id` is a randomly chosen number which uniquely identifies the change output.
+Outputs in the funding transaction will be sorted by `change_serial_id` and `fund_output_serial_id`.
+
+`fund_output_serial_id` is a randomly chosen number which uniquely identifies the funding output.
+Outputs in the funding transaction will be sorted by `change_serial_id` and `fund_output_serial_id`.
+
 `feerate_per_vb` indicates the fee rate in satoshi per virtual byte that both
 sides will use to compute fees in the funding transaction, as described in the
 [transaction specification](Transactions.md).
@@ -114,6 +126,8 @@ The sending node MUST:
   - set `total_collateral_satoshis` to a value greater than or equal to 1000.
   - set `contract_maturity_bound` and `contract_timeout` to either both be UNIX timestamps, or both be block heights as distinguished [here](https://en.bitcoin.it/wiki/NLockTime).
   - set `contract_maturity_bound` to be less than `contract_timeout`.
+  - use a unique `input_serial_id` for each input
+  - set `change_serial_id` and `fund_output_serial_id` to different values
 
 The sending node SHOULD:
 
@@ -144,6 +158,8 @@ The receiving node MUST reject the contract if:
   - it considers `feerate_per_vb` too small for timely processing or unreasonably large.
   - `funding_pubkey` is not a valid secp256k1 pubkey in compressed format.
   - `funding_inputs` do not contribute at least `total_collateral_satoshis` plus full [fee payment](Transactions.md#fee-payment).
+  - Any `input_serial_id` is duplicated
+  - The `fund_output_serial_id` and `change_serial_id` are not set to different value
 
 ### The `accept_dlc` Message
 
@@ -158,9 +174,11 @@ and closing transactions.
    * [`u64`:`total_collateral_satoshis`]
    * [`point`:`funding_pubkey`]
    * [`spk`:`payout_spk`]
+   * [`u64`:`payout_serial_id`]
    * [`u16`:`num_funding_inputs`]
    * [`num_funding_inputs*funding_input`:`funding_inputs`]
    * [`spk`:`change_spk`]
+   * [`u64`:`change_serial_id`]
    * [`cet_adaptor_signatures`:`cet_adaptor_signatures`]
    * [`signature`:`refund_signature`]
    * [`negotiation_fields`:`negotiation_fields`]
@@ -175,6 +193,10 @@ The sender MUST:
   - set `cet_adaptor_signatures` to valid adaptor signatures, using its `funding_pubkey` for each CET, as defined in the [transaction specification](Transactions.md#contract-execution-transaction) and using signature public keys computed using the `offer_dlc`'s `contract_info` and `oracle_info` as adaptor points.
   - include an adaptor signature in `cet_adaptor_signatures` for every event specified in the `offer_dlc`'s `contract_info`.
   - set `refund_signature` to the valid signature, using its `funding_pubkey` for the refund transaction, as defined in the [transaction specification](Transactions.md#refund-transaction).
+  - use a unique `input_serial_id` for each input
+  - use a unique `change_serial_id`
+  - set `change_serial_id` so that it is not equal to the `fund_output_serial_id`
+  - use a unique `payout_serial_id`
 
 The sender SHOULD:
 
@@ -187,8 +209,16 @@ The receiver:
     - MAY reject the contract.
   - if `cet_adaptor_signatures` or `refund_signature` fail validation:
     - MUST reject the contract.
-- if `funding_inputs` do not contribute at least `total_collateral_satoshis` plus [fee payment](Transactions.md#fee-payment)
-  - MUST reject the contract.
+  - if `funding_inputs` do not contribute at least `total_collateral_satoshis` plus [fee payment](Transactions.md#fee-payment)
+    - MUST reject the contract.
+  - if any `input_serial_id` is duplicated
+    - MUST reject the contract.
+  - if any `change_serial_id` is duplicated
+    - MUST reject the contract.
+  - if any `change_serial_id` is equal to the `fund_output_serial_id`
+    - MUST reject the contract.
+  - if any `payout_serial_id` is duplicated
+    - MUST reject the contract.
 
 Other fields have the same requirements as their counterparts in `offer_dlc`.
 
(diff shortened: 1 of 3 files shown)


Protocol.md
```diff
@@ -71,10 +71,12 @@ the funding transaction and CETs.
    * [`oracle_info`:`oracle_info`]
    * [`point`:`funding_pubkey`]
    * [`spk`:`payout_spk`]
+   * [`u64`:`payout_serial_id`]
    * [`u64`:`total_collateral_satoshis`]
    * [`u16`:`num_funding_inputs`]
    * [`num_funding_inputs*funding_input`:`funding_inputs`]
    * [`spk`:`change_spk`]
+   * [`u64`:`output_serial_id`]
```
nkohen: is this meant to be called `change_serial_id`? or else is `payout_serial_id` meant to be `cet_output_serial_id`?

Protocol.md
```diff
@@ -116,6 +124,7 @@ The sending node MUST:
   - set `total_collateral_satoshis` to a value greater than or equal to 1000.
   - set `contract_maturity_bound` and `contract_timeout` to either both be UNIX timestamps, or both be block heights as distinguished [here](https://en.bitcoin.it/wiki/NLockTime).
   - set `contract_maturity_bound` to be less than `contract_timeout`.
+  - use a unique `input_serial_id` for each input
```
nkohen, line 129: should there be a similar requirement for output ids?

Transactions.md
```diff
@@ -30,9 +30,8 @@ The funding inputs and change output script public keys are negotiated in the of
 
 ## Funding Transaction Input and Output Ordering
 
-The Offerer/Initiator's inputs come first and in the order that they were listed in the Offer message, the Accepter's inputs come after this in the order they are listed in the Accept message. The funding output comes first, followed by the Offerer's change output if applicable, and then the Accepter's change output if applicable.
-
-Note that this will likely change in the future.
+The inputs are sorted by each `funding_input`'s `input_serial_id` in ascending order.
+The outputs are sorted by `output_serial_id` in ascending order with the funding output having an `output_serial_id` of `0x8000000000000000`
```
nkohen: the constant probably deserves a rationale, I can see it is half-way in the middle of the domain but saying why it isn't also random is probably worth-while

nkohen (commented):
Sorry this took me a while to get around to, looking good :)

Protocol.md
benthecarman: I originally had it as `change_serial_id` but changed to `output_serial_id` because it's weird to give the funding output a `change_serial_id`

Protocol.md
benthecarman: This is in the offer and there is only one change & payout id, so they can't have a duplicate

benthecarman: Added the requirements for the accept message

Protocol.md
Tibo-lg: What about calling it `fund_output_serial_id`? It would be less ambiguous IMHO.

Protocol.md
```diff
@@ -146,6 +155,8 @@ The receiving node MUST reject the contract if:
   - it considers `feerate_per_vb` too small for timely processing or unreasonably large.
   - `funding_pubkey` is not a valid secp256k1 pubkey in compressed format.
   - `funding_inputs` do not contribute at least `total_collateral_satoshis` plus full [fee payment](Transactions.md#fee-payment).
+  - Any `input_serial_id` is duplicated
+  - The `output_serial_id` equal to `0x8000000000000000`
```
Tibo-lg: I think this requirement should be stated above somewhere

Protocol.md
```diff
@@ -160,9 +171,11 @@ and closing transactions.
    * [`u64`:`total_collateral_satoshis`]
    * [`point`:`funding_pubkey`]
    * [`spk`:`payout_spk`]
+   * [`u64`:`payout_serial_id`]
    * [`u16`:`num_funding_inputs`]
    * [`num_funding_inputs*funding_input`:`funding_inputs`]
    * [`spk`:`change_spk`]
+   * [`u64`:`change_serial_id`]
```
Tibo-lg, line 181: I think you forgot to update this?

Transactions.md
```diff
@@ -30,9 +30,11 @@ The funding inputs and change output script public keys are negotiated in the of
 
 ## Funding Transaction Input and Output Ordering
 
-The Offerer/Initiator's inputs come first and in the order that they were listed in the Offer message, the Accepter's inputs come after this in the order they are listed in the Accept message. The funding output comes first, followed by the Offerer's change output if applicable, and then the Accepter's change output if applicable.
+The inputs are sorted by each `funding_input`'s `input_serial_id` in ascending order.
+The outputs are sorted by `output_serial_id` in ascending order with the funding output having an `output_serial_id` of `0x8000000000000000`
 
-Note that this will likely change in the future.
+The funding output has a `output_serial_id` of `0x8000000000000000` since it is the midpoint of an uint64.
+This is chosen to give the funding output the most random index among the outputs.
```
Tibo-lg: I'm a bit dubious about that. Tbh this const makes thing more complicated, and I have trouble understanding how you can make something more random by using a constant (and so this rational didn't do a good enough job at convincing me :) )

benthecarman: Updated, lmk what you think

Protocol.md
benthecarman: Used @Tibo-lg's suggestion

Transactions.md
```diff
@@ -30,9 +30,12 @@ The funding inputs and change output script public keys are negotiated in the of
 
 ## Funding Transaction Input and Output Ordering
 
-The Offerer/Initiator's inputs come first and in the order that they were listed in the Offer message, the Accepter's inputs come after this in the order they are listed in the Accept message. The funding output comes first, followed by the Offerer's change output if applicable, and then the Accepter's change output if applicable.
+The inputs are sorted by each `funding_input`'s `input_serial_id` in ascending order.
+The outputs are sorted by `fund_output_serial_id` in ascending order with the funding output having an `fund_output_serial_id` of `0x8000000000000000`
 
-Note that this will likely change in the future.
+The funding output has a `fund_output_serial_id` of `0x8000000000000000` since it is the midpoint of an uint64.
+This is chosen to give the funding output so each user's output has an equal likelihood of being before or after the funding output.
```
nkohen: I'm still confused why this is here? Why not just have a random number like the rest? The dual funded lightning channel proposal gives the funding output no special treatment.

The issue here is that this introduces a fingerprint as the funding output is more likely to be near the middle instead of in a random place.

benthecarman: Tried to be clever thinking this was an equal distribution, ran some simulations and now see that it does favor having the funding output in the middle.

benthecarman: Updated to give it its own serial id

nkohen (approved):

Protocol.md
```diff
@@ -98,6 +101,15 @@ the sender and `funding_inputs` contains outputs, outpoints, and expected weight
 of the sender's funding inputs. `change_spk` specifies the script pubkey that funding
 change should be sent to.
 
+`payout_serial_id` is a randomly chosen number which uniquely identifies the payout output.
```
Tibo-lg: nit: add "of the offering party".

Protocol.md
```diff
@@ -114,6 +126,8 @@ The sending node MUST:
   - set `total_collateral_satoshis` to a value greater than or equal to 1000.
   - set `contract_maturity_bound` and `contract_timeout` to either both be UNIX timestamps, or both be block heights as distinguished [here](https://en.bitcoin.it/wiki/NLockTime).
   - set `contract_maturity_bound` to be less than `contract_timeout`.
+  - use a unique `input_serial_id` for each input
+  - set `change_serial_id` & `fund_output_serial_id` to different values
```
Tibo-lg: nit: it stresses me to use a symbol in the middle of a sentence

Protocol.md
```diff
@@ -144,6 +158,8 @@ The receiving node MUST reject the contract if:
   - it considers `feerate_per_vb` too small for timely processing or unreasonably large.
   - `funding_pubkey` is not a valid secp256k1 pubkey in compressed format.
   - `funding_inputs` do not contribute at least `total_collateral_satoshis` plus full [fee payment](Transactions.md#fee-payment).
+  - Any `input_serial_id` is duplicated
+  - The `fund_output_serial_id` and `change_serial_id` must be set to different values
```
Tibo-lg: nit: change to "are not set to different values" (it doesn't fit with the other sentences)

Protocol.md
```diff
@@ -187,8 +209,15 @@ The receiver:
...
-  - MUST reject the contract.
+  - if `funding_inputs` do not contribute at least `total_collateral_satoshis` plus [fee payment](Transactions.md#fee-payment)
+    - MUST reject the contract.
+
+The receiving node MUST reject the contract if:
+
+  - Any `input_serial_id` is duplicated
+  - Any `change_serial_id` is duplicated
+  - Any `change_serial_id` is equal to the `fund_output_serial_id`
+  - Any `payout_serial_id` is duplicated
```
Tibo-lg: nit: it's a bit weird to me that these are split in two, is there a reason for that?

Protocol.md
```diff
@@ -235,4 +264,4 @@ Nadav Kohen <nadavk25@gmail.com>
 
 ![Creative Commons License](https://i.creativecommons.org/l/by/4.0/88x31.png "License CC-BY")
 <br>
-This work is licensed under a [Creative Commons Attribution 4.0 International License](http://creativecommons.org/licenses/by/4.0/).
\ No newline at end of file
+This work is licensed under a [Creative Commons Attribution 4.0 International License](http://creativecommons.org/licenses/by/4.0/).
```
Tibo-lg, line 268: What's this diff about?

Transactions.md
```diff
@@ -30,9 +30,8 @@ The funding inputs and change output script public keys are negotiated in the of
 
 ## Funding Transaction Input and Output Ordering
 
-The Offerer/Initiator's inputs come first and in the order that they were listed in the Offer message, the Accepter's inputs come after this in the order they are listed in the Accept message. The funding output comes first, followed by the Offerer's change output if applicable, and then the Accepter's change output if applicable.
-
-Note that this will likely change in the future.
+The inputs are sorted by each `funding_input`'s `input_serial_id` in ascending order.
+The outputs are sorted by `change_serial_id` in ascending order with the funding output using the `fund_output_serial_id`.
```
Tibo-lg: suggestion: 
"The outputs are sorted in ascending order based on their respective `change_serial_id` or `fund_output_serial_id`"

Tibo-lg (commented):
Just nits

Protocol.md
benthecarman: My editor just added a new line to the end

benthecarman:
Responded to @Tibo-lg's review

Tibo-lg (approved):
Thanks for fixing the nits!

PR #137, opened 2021-02-11
Restrictions on Script Public Keys with standardness rules

### LINKED ISSUE #53: Restrictions on Funding Change Script Public Keys

benthecarman:
Closes #53 

standardness rules taken from: https://github.com/lightningnetwork/lightning-rfc/pull/672

### DIFF
--- a/Protocol.md
+++ b/Protocol.md
@@ -8,6 +8,7 @@
           * [The `offer_dlc` Message](#the-offer_dlc-message)
           * [The `accept_dlc` Message](#the-accept_dlc-message)
           * [The `sign_dlc` Message](#the-sign_dlc-message)
+          * [Script Pubkey Standardness Definition](#script-pubkey-standardness-definition)
 * [Authors](#authors)
 
 # Contract
@@ -122,6 +123,7 @@ The sending node MUST:
 
   - set undefined bits in `contract_flags` to 0.
   - ensure the `chain_hash` value identifies the chain it wishes to open the contract within.
+  - set `payout_spk` and `change_spk` to a [standard script pubkey](#script-pubkey-standardness-definition)
   - set `funding_pubkey` to a valid secp256k1 pubkey in compressed format.
   - set `total_collateral_satoshis` to a value greater than or equal to 1000.
   - set `contract_maturity_bound` and `contract_timeout` to either both be UNIX timestamps, or both be block heights as distinguished [here](https://en.bitcoin.it/wiki/NLockTime).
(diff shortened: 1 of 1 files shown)


nkohen (approved):

Protocol.md
```diff
@@ -228,10 +234,25 @@ The recipient:
...
 
+### Script Pubkey Standardness Definition
+
+ For a script pub key to be valid it must be in one of the following forms:
+
+    1. `OP_DUP` `OP_HASH160` `20` 20-bytes `OP_EQUALVERIFY` `OP_CHECKSIG` (pay to pubkey hash), OR
+    2. `OP_HASH160` `20` 20-bytes `OP_EQUAL` (pay to script hash), OR
+    3. `OP_0` `20` 20-bytes (version 0 pay to witness pubkey hash), OR
+    4. `OP_0` `32` 32-bytes (version 0 pay to witness script hash), OR
+    5. `OP_1` through `OP_16` inclusive, followed by a single push of 2 to 40 bytes
```
ariard, line 277: Are those rules scoping funding inputs script pubkey ? Note, non-Taproot and v1+ witness program spends aren't standards and will be rejected by the current Bitcoin Core policy (see `AreInputsStandard` in `src/policy/policy.cpp`). 

Unless we fully verify inputs pubkeys and witnesses against standardness rules, a malicious counterparty should be able to fail propagation of the funding tx...

Protocol.md
```diff
@@ -228,10 +234,25 @@ The recipient:
...
+       (witness program versions 1 through 16)
+
+  These script pub key forms include only standard forms accepted by the Bitcoin network, which ensures the resulting transaction will propagate to miners.
```
ariard: I would phrase this differently, "accepted by the wider set of deployed Bitcoin clients in the network, which increase the chances of successful propagation to miners". Policy rules are heterogeneous by design.

Tibo-lg (approved):

Protocol.md
benthecarman: Added to apply to funding inputs

ariard: Ooops, my bad, I meaned that _input_ scriptpubkey standardness isn't the same that _output_ one. Non-Taproot and v1+ witness program are excluded for inputs. So I would formalize differently,
"If the scriptpubkey is sent to: P2PKH, P2SH, P2WPH, P2WSH, segwit v1+"
"If the scriptpubkey is spent from : P2WPKH, P2WSH"

Note the restriction on inputs excluding P2SH, P2PKH, THIS IS UNSAFE TO USE NON-SEGWIT INPUTS !

A malicious counterparty can malleate the input, confirm the funding, break the sigs of second-stage dlcs/refund and thus take in hostage collateral funds.

benthecarman: Fixed to use the same wording for inputs as the BOLTs

ariard (commented):
SGTM 2256f71 :)

PR #139, opened 2021-02-15
Removed ranged event descriptors

nkohen:


Tibo-lg (changes requested):
nit: 

* change "Three kinds of event outcomes are defined (note that a single event can be composed of several types)." to "Two kinds..."

* remove in table of content the [Range]

Tibo-lg (approved):

PR #140, opened 2021-02-15
Added links to the dlc-dev mailing list

nkohen:


Christewart (approved):

Tibo-lg (approved):
LGTM (maybe I would have said "subscribe to" instead of "check out" but I don't care that much tbh :) )

PR #143, opened 2021-02-17
Added contract_info_v1 for Disjoint Union DLCs

nkohen:
I'm open to the idea of just replacing `contract_info_v0` instead of adding a new one, if anyone has an opinion one way or the other make sure to comment.


Tibo-lg (commented):
Does this work if I want disjoint events and have multi oracle in some of them?

Tibo-lg (approved):
Ignore my comment above, I had missed that oracle_info covers also multi oracle. So LGTM

benthecarman (approved):

PR #151, opened 2021-03-02
Add diagrams 🎨

matthewjablack:
This PR adds diagrams showing an overview of on-chain and off-chain transactions involved with entering and exiting a DLC. 

It also includes a fee-bumping diagram and depiction of participant/outsider view of a DLC from a block explorer. 

Thanks to @Abhimanyu1994 for putting these together 🙏 🔥  and @nkohen for the initial inspiration

**Overview:**

dlc_overview

**Overview + RBF:**

dlc_overview_rbf

**Participant View:**

dlc_participant_view

**Outsider View:**

dlc_outsider_view


Introduction.md
```diff
@@ -9,6 +9,8 @@ by every party involved.
 DLC [were originally proposed by Thaddeus Dryja](https://adiabat.github.io/dlc.pdf).
 The goal of DLC is to enable setting up contracts between two parties directly on the Bitcoin blockchain using an oracle to determine the contract outcome.
 
+![DLC Overview](/images/dlc_overview.png)
+
```
ghost, line 13: There is a typo in this image. It should be `UTXOs` instead of `UXTOS`

ghost (changes requested):
There is only one typo. Everything else looks good and will be helpful.

Introduction.md
benthecarman: In `dlc_overview_rbf` as well

matthewjablack: Good catch, updated

ghost: It's still incorrect in `dlc_overview`. `X` is at wrong place. 

Incorrect: U**X**TOs
Correct: U**TX**Os

matthewjablack: 2nd time is the charm haha. Updated thanks

Introduction.md
```diff
@@ -9,6 +9,8 @@ by every party involved.
 DLC [were originally proposed by Thaddeus Dryja](https://adiabat.github.io/dlc.pdf).
 The goal of DLC is to enable setting up contracts between two parties directly on the Bitcoin blockchain using an oracle to determine the contract outcome.
 
+![DLC Overview](/images/dlc_overview.png)
```
nkohen: For me (locally, not on github) these images only get properly linked in without the `/` prefix which I don't think should be here (on all pic links)

matthewjablack: Works either way for me so I'll switch to no `/` prefix.

nkohen (approved):

PR #152, opened 2021-03-03
Fraud Proofs

nkohen:
Introduces oracle fraud proofs!

### DIFF
--- a/FraudProofs.md
+++ b/FraudProofs.md
@@ -31,15 +31,16 @@ The fundamental building block of DLC oracle fraud proofs is a proof that a spec
 attested to a specific outcome, which can then be used with other information to prove fraud.
 
 In theory, it would be best if DLC oracles could deliver their announcements and attestations
-using some form of broadcast channel, but in practice it is expected that oracle messages will
-often times be received by users via APIs and other more private and P2P means.
+using some form of public broadcast channel, but in practice it is expected that oracle messages
+will oftentimes be received by users via APIs and other more private and certain P2P means.
 
 If DLC oracle broadcast channels did exist in such a way that oracles could not commit fraud privately,
 then oracle attestation proofs would be as simple as wrapping the publicly broadcast oracle messages,
 but practically speaking oracles have the ability to cheat without broadcasting fraudulent attestations publicly.
 
-Luckily, enough information is leaked to cheated DLC participants on-chain that they can still
-always generate proofs which show that a fraudulent oracle attestation must exist.
+Luckily in the case that anyone loses funds due to a fraudulently unlocked CET, enough information is leaked
+to the cheated DLC participants on-chain that they can still always generate proofs which show that a fraudulent
+oracle attestation must exist.
 
 #### Version 0 `oracle_attestation_proof`
 
@@ -59,8 +60,14 @@ on-chain information as the difference between the broadcast CET's signature and
 If `num_oracles = 1`, then the `aggregate_oracle_attestation` is directly equal to the attestation released by the one oracle. 
 As such, if one has access directly to an oracle's attestation, then this proof should use `num_oracles = 1`.
 
-The `oracle_announcements` and `oracle_outcomes` are used to compute a signature point `S` corresponding to an anticipation of
-these oracles attesting to these outcomes.
+The `oracle_outcomes`, which are the strings signed by the oracles, are either known directly from an `oracle_attestation`
+message (in the `outcome` field(s)), or they can be computed from an on-chain CET by using one's (decrypted) signature along
+with their adaptor signature to compute the `aggregate_oracle_attestation` which can be used to search the space of all
+possible `oracle_outcomes` by comparing `aggregate_oracle_attestation*G` to the adaptor point corresponding to any
+given `oracle_outcomes`. 
+
+The `oracle_announcements` and `oracle_outcomes` are used by a verifier to compute a signature point `S` corresponding to an
+anticipation of these oracles attesting to these outcomes.
 This signature point is used to validate that `aggregate_oracle_attestation*G = S` which proves that there must exist an attestation
 from each of the oracles in `oracle_announcements` of their corresponding outcome in `oracle_outcomes`.
 
--- a/FraudProofs.md
+++ b/FraudProofs.md
@@ -2,19 +2,47 @@
 
 ## Introduction
 
-TODO
+In the Discreet Log Contract oracle model, oracles act as oblivious public entities while
+their users act privately.
+
+Due to DLC oracles' public nature, should they ever commit fraudulent acts such as spreading
+false, invalid, or conflicting attestations, their fraud should be easily and compactly provable
+and publicly verifiable.
+After all, oracle fraud comes in the form of an unforgeable attestation of false or invalid events,
+or else in the form of multiple conflicting unforgeable attestations.
+
+This document specifies a standard format for serializing fraud proofs as well as specifying their verification.
 
 ## Table of Contents
 
 * [Proof of Oracle Attestation](#proof-of-oracle-attestation)
+  * [Version 0 `oracle_attestation_proof`](#version-0-oracle_attestation_proof)
 * [Claim of False Oracle Attestation](#claim-of-false-oracle-attestation)
+  * [Version 0 `false_oracle_attestation_claim` ](#version-0-false_oracle_attestation_claim)
 * [Proof of Invalid Oracle Attestation](#proof-of-invalid-oracle-attestation)
+  * [Version 0 `invalid_oracle_attestation_proof`](#version-0-invalid_oracle_attestation_proof)
 * [Proof of Oracle Equivocation](#proof-of-oracle-equivocation)
+  * [Version 0 `oracle_equivocation_proof`](#version-0-oracle_equivocation_proof)
+  * [Version 1 `oracle_equivocation_proof`](#version-1-oracle_equivocation_proof)
 * [Authors](#authors)
 
 ## Proof of Oracle Attestation
 
-TODO
+The fundamental building block of DLC oracle fraud proofs is a proof that a specific oracle
+attested to a specific outcome, which can then be used with other information to prove fraud.
+
+In theory, it would be best if DLC oracles could deliver their announcements and attestations
+using some form of broadcast channel, but in practice it is expected that oracle messages will
+often times be received by users via APIs and other more private and P2P means.
+
+If DLC oracle broadcast channels did exist in such a way that oracles could not commit fraud privately,
+then oracle attestation proofs would be as simple as wrapping the publicly broadcast oracle messages,
+but practically speaking oracles have the ability to cheat without broadcasting fraudulent attestations publicly.
+
+Luckily, enough information is leaked to cheated DLC participants on-chain that they can still
+always generate proofs which show that a fraudulent oracle attestation must exist.
+
+#### Version 0 `oracle_attestation_proof`
 
 1. type: ??? (`oracle_attestation_proof_v0`)
 2. data:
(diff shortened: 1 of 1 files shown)


FraudProofs.md
```diff
@@ -0,0 +1,158 @@
+# Fraud Proofs
+
+## Introduction
+
+In the Discreet Log Contract oracle model, oracles act as oblivious public entities while
+their users act privately.
+
+Due to DLC oracles' public nature, should they ever commit fraudulent acts such as spreading
+false, invalid, or conflicting attestations, their fraud should be easily and compactly provable
+and publicly verifiable.
```
Tibo-lg: mmmmm very picky but IMHO saying it can be compactly provable doesn't really hold for "false" ones

Tibo-lg: And not sure what you mean by "publicly verifiable".

FraudProofs.md
```diff
@@ -0,0 +1,158 @@
...
+   * [`num_oracles*oracle_outcome`:`oracle_outcomes`]
+   * [`32*bytes`:`aggregate_oracle_attestation`]
+   * [`u16`:`oracle_index`]
+3. subtype: `oracle_outcome`
+4. data:
+   * [`nb_signatures*string`:`outcomes`]
+
+The most important piece of this proof is the `aggregate_oracle_attestation` which is recoverable from
+on-chain information as the difference between the broadcast CET's signature and its corresponding adaptor signature.
+In the case that one has access directly to the `oracle_attestation` in question, then this can be used in place of an aggregate.
```
Tibo-lg: It's not clear to me if you're referring to attestation (which in the specs refer to a message) or a signature. Given the type  assume the latter but then better to distinguish names I feel. Also how do you know which one it is (aggregate or not)?

FraudProofs.md
```diff
@@ -0,0 +1,158 @@
...
+these oracles attesting to these outcomes.
+This signature point is used to validate that `aggregate_oracle_attestation*G = S` which proves that there must exist an attestation
+from each of the oracles in `oracle_announcements` of their corresponding outcome in `oracle_outcomes`.
+
+Finally, this proof references a specific oracle by `oracle_index` which is the index in `oracle_announcements` and `oracle_outcomes`
+corresponding to the referenced oracle's announcement and outcome to which they attested. 
+
+## Claim of False Oracle Attestation
+
+The most straightforward kind of oracle fraud occurs when an oracle attests to a false event outcome,
```
Tibo-lg, line 80: Really detail, but I feel incorrect would be better than false.

FraudProofs.md
```diff
@@ -0,0 +1,158 @@
...
+#### Version 1 `oracle_equivocation_proof`
+
+1. type: ??? (oracle_equivocation_proof_v1)
+2. data:
+   * [`oracle_announcement`:`oracle_announcement`]
+   * [`nb_signatures*string`:`outcomes_1`]
+   * [`32*bytes`:`oracle_attestation_1`]
+   * [`nb_signatures*string`:`outcomes_2`]
+   * [`32*bytes`:`oracle_attestation_2`]
+   * [`32*bytes`:`oracle_private_key`]
```
Tibo-lg: Is it necessary to put this in since anyway it can be recomputed?

Tibo-lg (commented):
Looks good! Just some minor comments

FraudProofs.md
nkohen: changed it to `In the case that one has access directly to an oracle's attestation, then this can be used as the aggregate.` (because it is just an aggregate of one).

FraudProofs.md
nkohen: I worry about the differentiation between these attestations and "invalid" ones (i.e. not in the announcement) which would both be considered "incorrect" I think? Open to names other than false though

FraudProofs.md
nkohen: it is not, and I'm open to removing it just thought it might be fun :P

Tibo-lg: Sorry if I'm being boring but I'd rather not have it.

FraudProofs.md
Tibo-lg: Yeah you're right and I cannot come up with something better.

Tibo-lg (approved):

FraudProofs.md
```diff
@@ -0,0 +1,154 @@
...
+  * [Version 1 `oracle_equivocation_proof`](#version-1-oracle_equivocation_proof)
+* [Authors](#authors)
+
+## Proof of Oracle Attestation
+
+The fundamental building block of DLC oracle fraud proofs is a proof that a specific oracle
+attested to a specific outcome, which can then be used with other information to prove fraud.
+
+In theory, it would be best if DLC oracles could deliver their announcements and attestations
+using some form of broadcast channel, but in practice it is expected that oracle messages will
```
ariard: I would rather say anonymous network. Note, I'm not sure how you dissociate broadcast channel from P2P means in this sentence.

Maybe say that p2p propagation of oracles events is left for future spec efforts ?

Also "that oracle messages will often times be received". Is that an english sentence ?

FraudProofs.md
```diff
@@ -0,0 +1,154 @@
...
+often times be received by users via APIs and other more private and P2P means.
+
+If DLC oracle broadcast channels did exist in such a way that oracles could not commit fraud privately,
```
ariard, line 37: I see what you  mean by broadcast channel. I think what we want is a strong primitive than broadcast channel, which is proof-of-publication. At least different in the sense of protocol participants joining the network in the future can still authenticate the proofs and with some expectations on the censorship-resistance property.

FraudProofs.md
```diff
@@ -0,0 +1,154 @@
...
+then oracle attestation proofs would be as simple as wrapping the publicly broadcast oracle messages,
```
ariard, line 38: Better to say "If DLC oracle broadcast channels did exist, oracle attestation proofs would be wrapped as a publicly broadcast oracle message and as such oracles could not commit fraud privately". 

If we made mandatory to publish attestation (lazyness==fraud) I think oracles couldn't cheat by restraining themselves from broadcast.

FraudProofs.md
```diff
@@ -0,0 +1,154 @@
...
+but practically speaking oracles have the ability to cheat without broadcasting fraudulent attestations publicly.
+
+Luckily, enough information is leaked to cheated DLC participants on-chain that they can still
+always generate proofs which show that a fraudulent oracle attestation must exist.
```
ariard: I think this point is only restrained to equivocation ? Invalid attestation won't allow to build a valid onchain transaction.

W.r.t to falseness, I think this is the same if it's published through an attestation or directly leaked onchain. It's more a "client-side" notion without a cryptographic definition.

FraudProofs.md
```diff
@@ -0,0 +1,154 @@
...
+
+The most important piece of this proof is the `aggregate_oracle_attestation` which is recoverable from
+on-chain information as the difference between the broadcast CET's signature and its corresponding adaptor signature.
+If `num_oracles = 1`, then the `aggregate_oracle_attestation` is directly equal to the attestation released by the one oracle. 
+As such, if one has access directly to an oracle's attestation, then this proof should use `num_oracles = 1`.
+
+The `oracle_announcements` and `oracle_outcomes` are used to compute a signature point `S` corresponding to an anticipation of
+these oracles attesting to these outcomes.
+This signature point is used to validate that `aggregate_oracle_attestation*G = S` which proves that there must exist an attestation
+from each of the oracles in `oracle_announcements` of their corresponding outcome in `oracle_outcomes`.
```
ariard, line 73: I think you should define `oracles_outcome`. Even if it's  signature, where does it come from ? Likely you should write an onchain section. 

"A DLC client discovering a CET spending funding output without an adaptor secret it's aware of:
       * MUST build an `oracle_attestation_proof`"

FraudProofs.md
```diff
@@ -0,0 +1,154 @@
...
+3. subtype: `oracle_outcome`
+4. data:
+   * [`nb_signatures*string`:`outcomes`]
+
+The most important piece of this proof is the `aggregate_oracle_attestation` which is recoverable from
+on-chain information as the difference between the broadcast CET's signature and its corresponding adaptor signature.
+If `num_oracles = 1`, then the `aggregate_oracle_attestation` is directly equal to the attestation released by the one oracle. 
+As such, if one has access directly to an oracle's attestation, then this proof should use `num_oracles = 1`.
+
+The `oracle_announcements` and `oracle_outcomes` are used to compute a signature point `S` corresponding to an anticipation of
```
ariard: Can you explain how the computation is done or point to another part of the spec where it's described ?

FraudProofs.md
```diff
@@ -0,0 +1,154 @@
...
+often times be received by users via APIs and other more private and P2P means.
+
+If DLC oracle broadcast channels did exist in such a way that oracles could not commit fraud privately,
+then oracle attestation proofs would be as simple as wrapping the publicly broadcast oracle messages,
+but practically speaking oracles have the ability to cheat without broadcasting fraudulent attestations publicly.
+
+Luckily, enough information is leaked to cheated DLC participants on-chain that they can still
+always generate proofs which show that a fraudulent oracle attestation must exist.
+
+#### Version 0 `oracle_attestation_proof`
```
ariard, line 45: What about `oracle_revoked_attestation_proof` ? By broadcasting onchain an adaptor secret in contradiction with a known, authenticated `oracle_attestation`, the oracle did revoke such attestation.

I like how it can be compared to the Lightning model. There a honest participant will prove to the blockchain that a _commitment  state) is invalid by publishing a previously-known _revocation secret_.

In the DLC model, substitute the invalid _commitment state_ by the adaptor secret representing an outcome and the _revocation secret_ by the `oracle_announcement`.

FraudProofs.md
```diff
@@ -0,0 +1,154 @@
...
+Unlike false attestations, this kind of fraud can be automatically validated by client software.
+
+#### Version 0 `invalid_oracle_attestation_proof`
+
+1. type: ??? (`invalid_oracle_attestation_proof_v0`)
+2. data:
+   * [`oracle_attestation_proof`: `oracle_attestation_proof`]
+
+This type simply consists of an `oracle_attestation_proof` proving that a specific oracle attested to
+a specific `oracle_outcome` where this outcome is not allowed by that oracle's `oracle_announcement`.
```
ariard, line 117: Again, can you define what's an `oracle_outcome` and from where it's discovered :) ?

FraudProofs.md
```diff
@@ -0,0 +1,154 @@
...
+
+An important thing to keep in mind is that in the case of invalid attestations, there are no concerns related
+to recovering the `oracle_outcome` from on-chain information because there are no CETs or adaptor
+signatures built for such an outcome.
+In other words, this kind of fraud is only detectable and only important if the `oracle_outcome` is known
```
ariard, line 122: Shouldn't this say if the oracle_outcome is valid but the corresponding `oracle_announcement` is unknown ?

ariard (commented):
Just achieve a first parse. I feel this proposal would benefit a lot to have :
* basic security games to define new security notions like laziness, equivocation, invalidity, falseness
* fully specify proving/verifying algorithms for each kind of proof (like we're doing for #114)
* describe state machine for onchain/offchain interactivity (like we've done for #87)

Let me know what you think about it.

FraudProofs.md
nkohen: > I would rather say anonymous network. Note, I'm not sure how you dissociate broadcast channel from P2P means in this sentence.

The distinction I'm trying to make is between a twitter-like model where you can't easily tweet and have it appear different to different people as opposed to a more P2P setting where it is easier to lie to certain kinds of nodes "in private" and the much more private interaction of a REST API.

I changed the wording a little to say "public broadcast channel" and "certain P2P means."

> Maybe say that p2p propagation of oracles events is left for future spec efforts ?

Doesn't really seem relevant here

> Also "that oracle messages will often times be received". Is that an english sentence ?

often times has been changed to oftentimes but otherwise, I think so?

FraudProofs.md
nkohen: Isn't that essentially what the entire `Proof of Oracle Attestation` section is devoted to? proof-of-publication?

FraudProofs.md
nkohen: > Better to say "If DLC oracle broadcast channels did exist, oracle attestation proofs would be wrapped as a publicly broadcast oracle message and as such oracles could not commit fraud privately".

This was more confusing to me than what is there right now, oracle attestation proofs wrap oracle messages, not the other way around? And it seems weird to mention that you can't commit fraud privately after as it is that fact which causes the wrapping to be valid, not the other way around though I'll admit this is a lesser problem than the first point.

nkohen: Open to what others think though

FraudProofs.md
nkohen: > Invalid attestation won't allow to build a valid onchain transaction.

Ah right, I change the wording to make this a more conditional sentence.

> W.r.t to falseness, I think this is the same if it's published through an attestation or directly leaked onchain. It's more a "client-side" notion without a cryptographic definition.

 This is incorrect, if you are using more than one nonce (due to numeric outcomes, multiple oracles, or both) then you will never be able to recover oracle attestations from on-chain information as you only have access to aggregate attestations.

FraudProofs.md
nkohen: I wasn't able to understand what you were suggesting here, is what you're describing not exactly what an equivocation proof is?

FraudProofs.md
nkohen: No. If the announcement is not known then there is no feasible way to prove fraud or attestation of any kind because you don't even know what keys are let alone messages, not really sure what it would mean for an `oracle_announcement` to be unknown and for anything in this document to matter.

What is being said here is that if you know the `oracle_announcement` and `oracle_outcome` and that outcome is not in the announcement then you can prove fraud, but if you don't know the `oracle_outcome` then you can't know that a given message is an attestation of anything because without knowledge of the outcome it is just a random 32 bytes.

FraudProofs.md
nkohen: Made this issue for adding a good place to link to (also has the description if you're curious): https://github.com/discreetlogcontracts/dlcspecs/issues/156

FraudProofs.md
nkohen: added a "definition" of "the strings signed by the oracles."

As for having an actual "on-chian section" I think that is likely out-of-scope for this PR and maybe even of v0 (although I'd be open to having this be created before v0 was out, just not likely a high priority).

FraudProofs.md
```diff
@@ -0,0 +1,161 @@
...
+then oracle attestation proofs would be as simple as wrapping the publicly broadcast oracle messages,
+but practically speaking oracles have the ability to cheat without broadcasting fraudulent attestations publicly.
+
+Luckily in the case that anyone loses funds due to a fraudulently unlocked CET, enough information is leaked
+to the cheated DLC participants on-chain that they can still always generate proofs which show that a fraudulent
+oracle attestation must exist.
+
+#### Version 0 `oracle_attestation_proof`
+
+1. type: ??? (`oracle_attestation_proof_v0`)
```
benthecarman, line 47: Need to pick type

FraudProofs.md
```diff
@@ -0,0 +1,161 @@
...
+   * [`num_oracles*oracle_outcome`:`oracle_outcomes`]
+   * [`32*bytes`:`aggregate_oracle_attestation`]
+   * [`u16`:`oracle_index`]
+3. subtype: `oracle_outcome`
+4. data:
+   * [`nb_signatures*string`:`outcomes`]
+
+The most important piece of this proof is the `aggregate_oracle_attestation` which is recoverable from
+on-chain information as the difference between the broadcast CET's signature and its corresponding adaptor signature.
+If `num_oracles = 1`, then the `aggregate_oracle_attestation` is directly equal to the attestation released by the one oracle.
```
benthecarman: > then the `aggregate_oracle_attestation` is directly equal to the attestation released by the one oracle. 

Wouldn't this be false for numeric events

FraudProofs.md
```diff
@@ -0,0 +1,161 @@
...
+
+As such, this specification only provides a standard format for making such claims which is generally
+intended for use with other external infrastructure.
+That said, it is possible for DLC client software to accept the manual importing of these claims after
+which user's can be taken through manual validation step-by-step.
+Such a process can serve as a manual means of black-listing oracle keys.
+
+#### Version 0 `false_oracle_attestation_claim`
+
+1. type: ??? (`false_oracle_attestation_claim_v0`)
```
benthecarman, line 95: need to pick type

FraudProofs.md
```diff
@@ -0,0 +1,161 @@
...
+## Proof of Oracle Equivocation
+
+Another form of oracle fraud for which there exists a straightforward automatic validation is equivocation.
+Equivocation is when an oracle attests to multiple messages for the same event.
+
+Validation in this case simply requires validating that both attestations exist using `oracle_attestation_proof`s.
+
+#### Version 0 `oracle_equivocation_proof`
+
+1. type: ??? (`oracle_equivocation_proof_v0`)
```
benthecarman, line 135: need to pick type

FraudProofs.md
```diff
@@ -0,0 +1,161 @@
...
+2. data:
+   * [`oracle_attestation_proof`:`oracle_attestation_proof_1`]
+   * [`oracle_attestation_proof`:`oracle_attestation_proof_2`]
+
+The most general kind of oracle equivocation proof is simply the combination of two `oracle_attestation_proof`s
+for the same `oracle_announcement` but for different `oracle_outcome`s.
+
+#### Version 1 `oracle_equivocation_proof`
+
+1. type: ??? (oracle_equivocation_proof_v1)
```
benthecarman, line 145: need to pick type

Tibo-lg:
I realized that with the new attestation scheme there is another possible kind of fraud from oracles, which is to produce two different announcements. I don't think this is covered here?

Christewart:
Does this have two implementations ? cc @Tibo-lg @nkohen @matthewjablack @benthecarman  (I need to look into bitcoin-s, but maybe ben/nadav know)

Tibo-lg:
> Does this have two implementations ? cc @Tibo-lg @nkohen @matthewjablack @benthecarman (I need to look into bitcoin-s, but maybe ben/nadav know)

I don't have it implemented on my side at least.

PR #153, opened 2021-03-05
Use no hash oracle attestation

Tibo-lg:
Update the signature scheme to use @llfourn [proposal](https://mailmanlists.org/pipermail/dlc-dev/2020-December/000002.html)

Also removed the requirement for NFC standardization of UTF8 as it does not seem strictly necessary since we are not hashing strings anymore. If anybody can think of a reason that we should keep it please comment.

### DIFF
--- a/CETCompression.md
+++ b/CETCompression.md
@@ -41,9 +41,8 @@ digits where signatures were ignored.
 ## Adaptor Points with Multiple Signatures
 
 Given public key `P` and nonces `R1, ..., Rn` we can compute `n` individual signature points for
-a given event `(d1, ..., dn)` in the usual way: `si * G = Ri + H(P, Ri, di)*P`.
-To compute a composite adaptor point for all events which agree on the first `m` digits, where
-`m` is any positive number less than or equal to `n`, the sum of the corresponding signature
+a given event `(d1, ..., dn)` as follow: `si * G = Ri + di * P`.
+To compute a composite adaptor point for all events which agree on the first `m` digits, where `m` is any positive number less than or equal to `n`, the sum of the corresponding signature
 points is used: `s(1..m) * G = (s1 + s2 + ... + sm) * G = s1 * G + s2 * G + ... + sm * G`.
 
 When the oracle broadcasts its `n` signatures `s1, ..., sn`, the corresponding adaptor secret can be
--- a/Oracle.md
+++ b/Oracle.md
@@ -3,31 +3,35 @@
 ## Introduction
 
 For the purpose of these specifications, an event is a digital representation of a real world fact.
-An [oracle](./Introduction.md#Oracle) is an entity that commits to publishing one or more signatures over a (number of) event outcome(s) ahead of time by releasing one or more [R-values](./Introduction.md#R-value) as well as necessary information for two parties to build a set of [CETs](./Introduction.md#Contract-Execution-Transaction-(CET)).
+An [oracle](./Introduction.md#Oracle) is an entity that commits to publishing one or more attestations over a (number of) event outcome(s) ahead of time by releasing one or more [R-values](./Introduction.md#R-value) as well as necessary information for two parties to build a set of [CETs](./Introduction.md#Contract-Execution-Transaction-(CET)).
 This necessary information is committed to in a so-called [_event descriptor_](#Event-descriptor) that will be further detailed in this document.
 
 ## Table of Contents
 
-- [Event descriptor](#event-descriptor)
-   - [Simple enumeration](#simple-enumeration)
-      - [Example: Weather tomorrow](#example-weather-tomorrow)
-   - [Digit decomposition](#digit-decomposition)
-      - [Example: BTC/USD rate](#example-btcusd-rate)
-   - [Serialization and signing of outcome values](#serialization-and-signing-of-outcome-values)
-   - [Serialization of event descriptors](#serialization-of-event-descriptors)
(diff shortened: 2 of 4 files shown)


CETCompression.md
```diff
@@ -41,9 +41,8 @@ digits where signatures were ignored.
 ## Adaptor Points with Multiple Signatures
 
 Given public key `P` and nonces `R1, ..., Rn` we can compute `n` individual signature points for
-a given event `(d1, ..., dn)` in the usual way: `si * G = Ri + H(P, Ri, di)*P`.
-To compute a composite adaptor point for all events which agree on the first `m` digits, where
-`m` is any positive number less than or equal to `n`, the sum of the corresponding signature
+a given event `(d1, ..., dn)` as follow: `si * G = Ri + cj * P` where `cj` is the value of the digit being signed.
```
nkohen: why not just write this as `si * G = Ri + di * P`?

nkohen: also, just to clarify with @LLFourn, it is okay to use `d_i = 0` correct? Even though it reveals the nonce key?

Oracle.md
```diff
@@ -90,15 +90,13 @@ For example, if the rate at maturity time is $20000, the oracle will sign the di
 
 If the rate is greater than $999999, the oracle signs the digits `9`, `9`, `9`, `9`, `9`, `9`.
 
-### Serialization and signing of outcome values
+### Outcome signing
```
nkohen: Should probably be consistent and call these attestations to disambiguate from signing, might also be worth some `ctrl + f` to make sure this doesn't happen elsewhere

Oracle.md
```diff
@@ -189,20 +185,19 @@ The TLV serialization of oracle attestations is as follows.
...
 
-The algorithm `Sign(sk, message, tag)` is defined as:
+The algorithm `Sign(sk, ci, r)` is defined as:
 
-* Let H = `tag_hash("DLC/oracle/" || tag)`
-* Let `m` = H(`message`)
-* Return `BIP340_sign(sk, m)`
+* Let `ci` the index of the outcome in the ordered list of possible outcomes that could be attested by the oracle. This list is either explicitly given by the oracle in the case of enumerated outcomes, or implicitly in the case of numerical decomposition.
+* Let `r` the pre-image of `R` announced by the oracle,
+* Return `s = r + ci * x`
```
nkohen, line 204: probably worth a link out to Lloyd's paper here or some other place.

CETCompression.md
Tibo-lg: I thought the `d` refers to the digits themselves, not their value, is that incorrect?

nkohen: I'm not sure I understand the distinction, isn't `d = 0 or 1` for example?

Tibo-lg: Nevermind you're right fixed.

Oracle.md
Tibo-lg: Replaced everything with attestation and attest. Sounds a bit weird to me but let me know what you think.

Oracle.md
Tibo-lg: Added

benthecarman:
> Also removed the requirement for NFC standardization of UTF8 as it does not seem strictly necessary since we are not hashing strings anymore. If anybody can think of a reason that we should keep it please comment.

This still is needed for things like the `eventid` and for enum outcomes. We hash the entire `event_tlv` when signing the announcement so these need to be standardized.

Tibo-lg:
> This still is needed for things like the eventid and for enum outcomes. We hash the entire event_tlv when signing the announcement so these need to be standardized.

Good point. Though thinking about it a bit more, I'm actually not sure why we are using canonical encoding in the first place. Looking at it more closely, it seems that it is useful when trying to compare or search strings, which is not what we are doing. The only reason we would need it is if the codes would somehow be automatically changed based on implementations if they were not canonical, but AFAICT it is not the case.

So can someone give a good reason for using it in the first place? Then at least I could add it as a rational to the document.

benthecarman:
> So can someone give a good reason for using it in the first place? Then at least I could add it as a rational to the document.

Imo there should at least be one standard so people know how to correctly interpret these bytes to display to the user.

Tibo-lg:
> Imo there should at least be one standard so people know how to correctly interpret these bytes to display to the user.

So that's the job of utf8. From my understanding, what NFC or NFD provide is a guarantee that two strings that "display" the same will have the same encoding, and it seems to be mainly useful for comparing or searching strings. But for our context I can't really figure out what is the benefit.

LLFourn:
> 
> 
> > Also removed the requirement for NFC standardization of UTF8 as it does not seem strictly necessary since we are not hashing strings anymore. If anybody can think of a reason that we should keep it please comment.
> 
> This still is needed for things like the `eventid` and for enum outcomes. We hash the entire `event_tlv` when signing the announcement so these need to be standardized.

Why does this mean it needs to be standardized in the case of event ids and enum outcomes?

> Imo there should at least be one standard so people know how to correctly interpret these bytes to display to the user.

They are just strings so why can't they be displayed to the user normalized or not?

benthecarman:
> From my understanding, what NFC or NFD provide is a guarantee that two strings that "display" the same will have the same encoding, and it seems to be mainly useful for comparing or searching strings. But for our context I can't really figure out what is the benefit.

I misunderstood what normalization meant, if it doesn't change the actual serialization then it would be fine to exclude and leave it up to the implementation

PR #154, opened 2021-03-05
Adaptor point optimizations

Tibo-lg:
Added some possible optimizations to signature point computation for numerical outcome DLCs.

This is written on top of #153

### DIFF
--- a/NumericOutcomeCompression.md
+++ b/NumericOutcomeCompression.md
@@ -35,6 +35,9 @@ where attestations are ignored.
     * [Counting Adaptor Points](#counting-adaptor-points)
     * [Optimizations](#optimizations)
   * [Algorithms](#algorithms)
+  * [Adaptor Point Computation Optimizations](#adaptor-point-computation-optimizations)
+    * [Pre-computing](#pre-computing)
+    * [Memoization](#memoization)
   * [Reference Implementations](#reference-implementations)
 * [Authors](#authors)
 
@@ -48,6 +51,8 @@ points is used: `s(1..m) * G = (s1 + s2 + ... + sm) * G = s1 * G + s2 * G + ... 
 
 When the oracle broadcasts its `n` attestations `s1, ..., sn`, the corresponding aggreate adaptor secret
 can be computed as `s(1..m) = s1 + s2 + ... + sm` which can be used to broadcast a corresponding CET.
+
+Note that [optimizations](#Adaptor-Point-Computation-Optimizations) can be applied when computing signature points.
 
 #### Rationale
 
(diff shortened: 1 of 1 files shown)


CETCompression.md
```diff
@@ -48,6 +48,8 @@ points is used: `s(1..m) * G = (s1 + s2 + ... + sm) * G = s1 * G + s2 * G + ...
 When the oracle broadcasts its `n` signatures `s1, ..., sn`, the corresponding adaptor secret can be
 computed as `s(1..m) = s1 + s2 + ... + sm` which can be used to broadcast the CET.
 
+Note that [optimizations](#Adaptor-Point-Computation-Optimizations) can be applied when computing large number of signature points.
```
nkohen: nit: no need to say "large number of"

CETCompression.md
```diff
@@ -368,6 +370,43 @@ def groupByIgnoringDigits(start: Long, end: Long, base: Int, numDigits: Int): Ve
 }
 ```
 
+## Adaptor Point Computation Optimizations
+
+Computing a large number of adaptor points for events with outcome composed of many digits is a computationally intensive process.
```
nkohen: probably worth noting for context that in reference implementations, signature point computation accounts for a significant proportion of runtime

CETCompression.md
```diff
@@ -368,6 +370,43 @@ def groupByIgnoringDigits(start: Long, end: Long, base: Int, numDigits: Int): Ve
...
+...
+S9_0 = R9 + 0 * P
+S9_1 = R9 + 1 * P
+```
+
+Then compute the aggregate signature points by combining the Si_j.
+For example, to compute the signature point for the value `0 0 0 0 0 0 0 0 0 0`, compute:
+```
+S = S0_0 + S1_0 + ... + S9_0
+```
```
nkohen: probably worth a second example just to drive home how this is actually an optimization and not just a round-about way of computing the same thing

CETCompression.md
```diff
@@ -368,6 +370,43 @@ def groupByIgnoringDigits(start: Long, end: Long, base: Int, numDigits: Int): Ve
...
+
+In addition to the above pre-computation, memoization can be used to further improve performance.
+Assuming an event with outcomes decomposed in base 2 with 10 digits, the algorithm works as follow:
+1. Compute `S0_0` for the value `0`
+2. Compute `S = S0_0 + S1_0` for the value `0 0`
+3. Compute `S' = S0_0 + S1_0 + S2_0` for the value `0 0 0`
+...
+10. Compute `S'' = S0_0 + S1_0 + ... + S8_0` for the value `0 0 0 0 0 0 0 0 0`
+11. Compute `S''' = S0_0 + S1_0 + ... + S9_0` for the value `0 0 0 0 0 0 0 0 0 0`
+12. Compute `S'''' = S'' + S9_1` for the value `0 0 0 0 0 0 0 0 0 1`
```
nkohen: This seems like more of an example than an algorithm imo

CETCompression.md
```diff
@@ -368,6 +370,43 @@ def groupByIgnoringDigits(start: Long, end: Long, base: Int, numDigits: Int): Ve
...
+
+Then compute the aggregate signature points by combining the Si_j.
+For example, to compute the signature point for the value `0 0 0 0 0 0 0 0 0 0`, compute:
+```
+S = S0_0 + S1_0 + ... + S9_0
+```
+
+### Memoization
+
+In addition to the above pre-computation, memoization can be used to further improve performance.
```
nkohen: Just to be safe, probably worth linking to https://en.wikipedia.org/wiki/Memoization on "memoization"

nkohen (commented):
This will be nice to have :) Also reading this has made me curious, @LLFourn do you know if secp's normal point multiplication is actually optimized to be fast for small scalars or if it will actually be more efficient to do something dumber like repeated addition? If the latter is the case it is probably worth a mention in the specs (under optimizations)

CETCompression.md
Tibo-lg: updated

CETCompression.md
Tibo-lg: updated but not entirely satisfied with the wording so let me know if you have better suggestions

CETCompression.md
Tibo-lg: Added a second example

CETCompression.md
Tibo-lg: Added TS code (thought rust was maybe not the most readable)

CETCompression.md
Tibo-lg: done

CETCompression.md
nkohen: good enough for me :)

CETCompression.md
nkohen: honestly I think it might be nice to have the example first so that it is easier to make sense of the algorithm but otherwise looks fine to me (will of course look more closely later when I implement this)

Tibo-lg: Good point updated to have the example first.  I don't think this is something urgent to merge so we can maybe wait until you have it implemented (and I cleaned up my implementation :) )

Tibo-lg:
@nkohen @LLFourn do you guys think this can be merged since we're not changing the way we compute signature points any time soon?

LLFourn:
Yes :)

Tibo-lg:
@Christewart @benthecarman Could one of you guy check this so I feel like I don't abuse my admin powers by merging it?

nkohen (approved):
I think this is good for merge though I think it might be nice to also add a quick link in here to your blog post about optimizations

Tibo-lg:
> I think this is good for merge though I think it might be nice to also add a quick link in here to your blog post about optimizations

[Added](https://github.com/discreetlogcontracts/dlcspecs/pull/154/files#diff-8b022b8ac6bc8877a6cd2ae8e585914e277281d272025c234f3fac5588822f1aR382)

PR #157, opened 2021-03-12
Numeric Outcome DLC refactor

nkohen:
Re-worked CET language in numeric outcome specifications and added a diagram, as well as some general clean-up.

(Renames `CETCompression.md` -> `NumericOutcomeCompression.md`)

Related to #77.

PR #158, opened 2021-03-12
Adaptor point doc

### LINKED ISSUE #156: Explicitly Describe Adaptor Point Computation
It appears that the current specification, we do have a description of adaptor point computation
```
Given public key `P` and nonces `R1, ..., Rn` we can compute `n` individual signature points for
a given event `(d1, ..., dn)` in the usual way: `si * G = Ri + H(P, Ri, di)*P`.
```
(and its updated non-BIP340 version on the PR that updates things)

But, this is in the CETCompression specification in a section describing how we handle the "special case" where there is more than one nonce.

Nowhere is the "usual way: `s*G = R + H(P, R, m)*P`" included elsewhere in the specification such as in either the `contract_info` or `oracle_announcement` definitions or else maybe in the Protocol.md specification, and this should probably be included in one of those places.

nkohen:
Built on top of #157.

Fixes #156.

Adds a specification document for Adaptor Point computation and ordering.


nkohen:
Currently this contains a detailed outline which I plan to turn into a specification tomorrow.

LLFourn:
NIT: I think "adaptor point" is not the right name since you can use them without using adaptor signatures. These points are the same thing if you are using non-adaptor sig DLCs. What about attestation points?

nkohen:
> NIT: I think "adaptor point" is not the right name since you can use them without using adaptor signatures. These points are the same thing if you are using non-adaptor sig DLCs. What about attestation points?

Done.

AttestationPoint.md
```diff
@@ -0,0 +1,108 @@
...
+This means that in order to spend any execution branch of a DLC, one must reveal the scalar pre-image for a point.
+As is described in this document, the points used for adaptor signing correspond to anticipations of all
+possible oracle attestations such that exactly one adaptor secret is revealed, unlocking exactly one signature,
+making exactly one Contract Execution Transaction (CET) valid for on-chain publication.
+
+By an anticipation of an oracle attestation, we mean an elliptic curve point `S` whose scalar pre-image `s` (i.e. `S = s*G`)
+is an oracle attestation of a specific message, `m` (or the sum of multiple such attestations).
+The point `S` can be thought of as an encryption key to be used on the signature of the CET which corresponds to the
+event where `m` is attested to so that this signature can only be used should this attestation be broadcast.
+The key point is that `S` can be computed *in advance* given information in oracle announcements so that anticipation points
```
Christewart, line 15: nit link [oracle announcement](https://github.com/discreetlogcontracts/dlcspecs/blob/master/Oracle.md#oracle-announcements)

AttestationPoint.md
```diff
@@ -0,0 +1,108 @@
...
+
+* **CET** - A Contract Execution Transaction (CET) is a Bitcoin transaction which spends the DLC funding output as an input
+  and outputs the DLC participants payouts.
+  Note that every adaptor point, attestation point, adaptor signature, and oracle outcome will map to some CET but that this mapping
+  is not one-to-one since multiple points/signatures/outcomes can map to the same CET if those outcomes result in equal payouts.
+* **Adaptor Point** - An Adaptor Point is an elliptic curve point used as an encryption key for a signature so that the encrypted signature
+  is known as an adaptor signature.
+  This encryption is used to make signatures (and transitively, transactions) conditional on the scalar pre-image of this point becoming known.
+  Often times an adaptor point is assigned meaning through its functional use in a Bitcoin contracting scheme such as a DLC.
+  Specifically for DLCs, adaptor points always correspond to oracle attestations so that they are always attestation points and
```
Christewart, line 43: I would bold this sentence to make sure it is not missed (i missed it my first time through :/)

Christewart:
Needs rebase, otherwise is this good to merge? AFAICT this is just extra documentation. I see it's still in "Draft" status. Any reason for that @nkohen ?

AttestationPoint.md
```diff
@@ -0,0 +1,108 @@
+# Deterministic Attestation Point Computation
+
+## Introduction
+
+Discreet Log Contracts (DLCs) are oracle contracts enforced using a set of adaptor signature "point-locks."
+This means that in order to spend any execution branch of a DLC, one must reveal the scalar pre-image for a point.
+As is described in this document, the points used for adaptor signing correspond to anticipations of all
+possible oracle attestations such that exactly one adaptor secret is revealed, unlocking exactly one signature,
+making exactly one Contract Execution Transaction (CET) valid for on-chain publication.
```
Tibo-lg, line 9: nit: it might be better to link to the glossary than to redefine acronyms every time?

AttestationPoint.md
```diff
@@ -0,0 +1,108 @@
...
+
+By an anticipation of an oracle attestation, we mean an elliptic curve point `S` whose scalar pre-image `s` (i.e. `S = s*G`)
+is an oracle attestation of a specific message, `m` (or the sum of multiple such attestations).
+The point `S` can be thought of as an encryption key to be used on the signature of the CET which corresponds to the
+event where `m` is attested to so that this signature can only be used should this attestation be broadcast.
+The key point is that `S` can be computed *in advance* given information in oracle announcements so that anticipation points
+(aka attestation points or sometimes adaptor points) can be used to construct a DLC and later the corresponding attestations
```
Tibo-lg, line 16: I feel there is a distinction between attestation point and adaptor point, as an adaptor point can be the sum of multiple attestation points.

AttestationPoint.md
```diff
@@ -0,0 +1,108 @@
...
+are used to execute that DLC.
+
+At a high level, this is done using the fact that oracle attestations, `s`, are validated against public key information in the usual way
+by checking that `s*G` is equal to a point computed in another way from public information, thus this point can be used as an attestation point.
```
Tibo-lg, line 20: I feel this sentence is unclear.

AttestationPoint.md
```diff
@@ -0,0 +1,108 @@
...
+* [Disambiguation of Terms](#disambiguation-of-terms)
+* [Attestation Point Computation](#attestation-point-computation)
+  * [Single Attestation Points](#single-attestation-points)
+  * [Multiple Attestation Aggregate Points](#multiple-attestation-aggregate-points)
+* [Single Oracle Enumerated Outcome Attestation Points](#single-oracle-enumerated-outcome-attestation-points)
+* [Single Oracle Numeric Outcome Attestation Points](#single-oracle-numeric-outcome-attestation-points)
+* [Multiple Oracle Attestation Points](#multiple-oracle-attestation-points)
+* [Authors](#authors)
+
+## Disambiguation of Terms
```
Tibo-lg, line 33: This should be in the glossary IMHO.

AttestationPoint.md
```diff
@@ -0,0 +1,108 @@
...
+
+* **CET** - A Contract Execution Transaction (CET) is a Bitcoin transaction which spends the DLC funding output as an input
```
Tibo-lg, line 35: I would remove the part about "funding output as an input" so we don't have to redefine things once we specify DLC channels.

AttestationPoint.md
```diff
@@ -0,0 +1,108 @@
...
+  and outputs the DLC participants payouts.
+  Note that every adaptor point, attestation point, adaptor signature, and oracle outcome will map to some CET but that this mapping
+  is not one-to-one since multiple points/signatures/outcomes can map to the same CET if those outcomes result in equal payouts.
+* **Adaptor Point** - An Adaptor Point is an elliptic curve point used as an encryption key for a signature so that the encrypted signature
+  is known as an adaptor signature.
```
Tibo-lg, line 40: what do you mean by "is known as an adaptor signature"?

AttestationPoint.md
```diff
@@ -0,0 +1,108 @@
...
+  This encryption is used to make signatures (and transitively, transactions) conditional on the scalar pre-image of this point becoming known.
```
Tibo-lg, line 41: This is unclear IMHO. It's not the signatures or transactions, that are conditional, but the ability to spend a UTXO.

AttestationPoint.md
```diff
@@ -0,0 +1,108 @@
...
+  Often times an adaptor point is assigned meaning through its functional use in a Bitcoin contracting scheme such as a DLC.
+  Specifically for DLCs, adaptor points always correspond to oracle attestations so that they are always attestation points and
+  these two terms are occasionally used inter-changeably in these specifications.
```
Tibo-lg, line 44: Again I don't really agree with that personally.

AttestationPoint.md
```diff
@@ -0,0 +1,108 @@
...
+* **Attestation Point** - Also known as an anticipation point, an Attestation Point is an elliptic curve point `S = s*G` such that its scalar pre-image,
+  `s`, is an oracle attestation (or sum of multiple oracle attestations).
+  These points can be computed without explicitly knowing the scalar pre-image using only an oracle's public announcement information.
```
Tibo-lg, line 47: I think it might be worth noting that it requires knowledge of the `R` value.

AttestationPoint.md
```diff
@@ -0,0 +1,108 @@
...
+  These points are used in the DLC specification as Adaptor Points.
+* **Adaptor Signature** - An Adaptor Signature is an encrypted (on-chain-valid) digital signature which is constructed by a signer with their key
+  along with an Adaptor Point, which is an encryption (public) key.
```
Tibo-lg, line 50: Better not to redefine, simply link to the adaptor point definition.

AttestationPoint.md
```diff
@@ -0,0 +1,108 @@
...
+  This signature can be validated without decryption so long as the verifier knows not only the message being signed and the signers public key,
```
Tibo-lg, line 51: nit: signer's?

AttestationPoint.md
```diff
@@ -0,0 +1,108 @@
...
+  but also the Adaptor Point used to encrypt the signature.
+  The adaptor signature can only be decrypted into a valid on-chain signature using knowledge of the scalar pre-image of the Adaptor Point.
```
Tibo-lg, line 53: Maybe worth introducing "adaptor secret"?

AttestationPoint.md
```diff
@@ -0,0 +1,108 @@
...
+  This process also leaks this scalar pre-image to the adaptor signer as the difference between the decrypted and encrypted signature is this secret.
+* **Digit Prefix** - When an oracle attests to a numeric outcomes such as `77 = 001001101` each bit is individually attested to.
+  As such, when DLC payouts are constant on some interval, DLC participants often only use a prefix of the bits signed and not all of them to
+  correspond to the set of outcomes where that prefix is followed by any bit-string, e.g. the Digit Prefix `001001` corresponds to the inclusive
+  interval of outcomes `[72, 79]` (assuming a `num_digits` of `9`) because `72 = 001001000` and `79 = 001001111`.
+* **Oracle Outcome** - TODO
+
+TODO: Don't forget to link to this from Protocol and Messaging docs as well as the two Numeric Outcome docs.
```
Tibo-lg, line 61: I guess needs to be done :)

Tibo-lg (commented):
This doesn't look like it's finished (all sections at the end look like TODOs, but put some comments.

AttestationPoint.md
Ichiro0219, line 40: Maybe better to add  2 definition
・the explanation of "the scalar pre-images" for oracle's each Attestation Points, these are included in oracle_attestation as signatures

・the explanation of "the scalar pre-image" for Adaptor Point that is used as an decryption key for adaptor signatures

PR #162, opened 2021-04-13
Move oracle related TLV info to Messaging.md

Tibo-lg:
The TOC of `Messaging.md` including some parts about oracle messages, but actually they were not there. As we discussed moving all TLV serialization info to Messaging I moved them there and updated the TOC.

### DIFF
--- a/Messaging.md
+++ b/Messaging.md
@@ -9,41 +9,43 @@ All data fields are unsigned big-endian unless otherwise specified.
 
 ## Table of Contents
 
-* [Connection Handling and Multiplexing](#connection-handling-and-multiplexing)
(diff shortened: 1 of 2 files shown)


nkohen (changes requested):
Looks good but I think you need to modify the table of contents for both files moving the sections from one to the other

Tibo-lg:
> Looks good but I think you need to modify the table of contents for both files moving the sections from one to the other

Good point thanks updated.

benthecarman (approved):
ACK d7a58e540db40ee8e48f1cd42ee557ea7a4f23c5

nkohen (approved):

PR #163, opened 2021-04-23
Update messaging and serialization

Tibo-lg:
This PR is a proposal to update the update the serialization to make it more efficient, concise and upgradable. The current message format specifies hard coded type prefixes for embedded type that don't require them, making it a maintenance burden and inefficient. It also lacks any extensibility or upgradability mechanism.

Changes include:
* Add protocol version to `offer_dlc`
* Add tlv streams to wire messages
* Define three types of subtypes and specific serialization format for each of them (and update type definitions accordingly)

I hesitated to move the wire message definitions to `Messaging.md` as previously discussed but to keep the diff small I didn't.


Protocol.md
```diff
@@ -81,6 +82,11 @@ the funding transaction and CETs.
    * [`u64`:`feerate_per_vb`]
    * [`u32`:`cet_locktime`]
    * [`u32`:`refund_locktime`]
+   * [`offer_tlvs`: `tlvs`]
+
+[//]: # (TODO: `protocol_version` should be advertised by DLC nodes in the future)
+The `protocol_version` value denotes the version of the protocol that the offering node is proposing.
+If the receiving node does not support the requested version, it should ignore the message.
```
benthecarman, line 90: maybe we should define a advertisement message now and just use that as an initial message in current setups so we don't have to define a new `offer_dlc_vx` once we have this.

Messaging.md
```diff
@@ -103,19 +137,19 @@ The following DLC-specific types are used throughout the specification. All type
 
 This type contains information about a contract's outcomes, their corresponding payouts, and the oracles to be used.
 
-#### Version 0 `contract_info`
+#### `single_contract_info`
 
-1. type: 55342 (`contract_info_v0`)
+1. type: 1
```
benthecarman: I am confused why we are removing these types

Protocol.md
Tibo-lg: let me know if I misunderstand you, but I feel having an advertisement message would not remove the need to have a prrotocol version field (if that's what you're proposing?). Nodes could potentially support multiple protocol version so announcing it in the offer message still seem necessary to me.

Messaging.md
Tibo-lg: Before we used type tags in two cases:
* for types that are embedded in others. In this case I couldn't see any use of the tag since if you know exactly what you want to parse a tag doesn't bring anything (happy to be corrected if I missed something).
* for cases where there are multiple version of a field (this case). I couldn't see any reason to use type numbers that are assigned by LN for wire messages, as this is both less efficient and a waste of them AFAICT (again happy to be corrected if I'm wrong).

Messaging.md
```diff
@@ -11,39 +11,43 @@ All data fields are unsigned big-endian unless otherwise specified.
...
+      - [`multi_oracle_info`](#multi_oracle_info)
    - [The `oracle_params` Type](#the-oracle_params-type)
-      - [Version 0 `oracle_params`](#version-0-oracle_params)
+      - [`oracle_params`](#oracle_params)
    - [The `negotiation_fields` Type](#the-negotiation_fields-type)
-      - [Version 0 `negotiation_fields`](#version-0-negotiation_fields)
-      - [Version 1 `negotiation_fields`](#version-1-negotiation_fields)
-      - [Version 2 `negotiation_fields`](#version-2-negotiation_fields)
+      - [`single_negotiation_fields`](#single_negotiation_fields)
+      - [Version 2 `disjoint_negotiation_fields`](#version-2-disjoint_negotiation_fields)
```
nkohen: is there a reason this is the only version number left?

Messaging.md
```diff
@@ -53,13 +57,43 @@ Implementations MUST use a single connection per peer; contract messages (which
...
-This means that types on outer-messages will be represented with `u16` integers (defined below) and their length
-is omitted from their encoding because the transport layer has the length in a separate unencrypted field.
-Meanwhile all typed sub-messages (which follow TLV format) will have their types represented using `bigsize` integers
-(defined below) and their lengths (also `bigsize`) are included in their encodings.
+We reuse the [Lightning Message Format](https://github.com/lightningnetwork/lightning-rfc/blob/master/01-messaging.md#lightning-message-format) for messages sent over the wire, the [Type-Length-Value Format](https://github.com/lightningnetwork/lightning-rfc/blob/master/01-messaging.md#type-length-value-format) (TLV) for message extensibility and extra features support, and a custom format for [sub-types](#sub-types).
+
+### Wire Messages
+
+Any encoded binary blob that can be sent over the wire will follow the Lightning Message Format.
+This means that wire messages are prefixed with a `u16` field (defined below) and their length is omitted from their encoding because the transport layer has the length in a separate un-encrypted field..
```
nkohen: nit: double period at the end

Messaging.md
```diff
@@ -53,13 +57,43 @@ Implementations MUST use a single connection per peer; contract messages (which
...
+
+#### Plain sub-type
+
+Plain sub-types do not have any particular format, and their field can simply be replaced in place where they are used.
+
+#### Sibling sub-type
+
+Sibling sub-types are prefixed with a `bigsize` type identifier.
+The type identifiers are specific to a set of type variants, and can thus be reused across different sub-types.
+Type identifiers are defined within the type definitions, starting from `1` and increasing by `1`.
```
nkohen: do we not want to support the "it's okay to be odd" design here?

Messaging.md
```diff
@@ -53,13 +57,43 @@ Implementations MUST use a single connection per peer; contract messages (which
...
+* [Sibling sub-types](#sibling-sub-type) to support multiple variants of a field,
+* [Optional sub-types](#optional-sub-type) to support optional fields.
+
+#### Plain sub-type
+
+Plain sub-types do not have any particular format, and their field can simply be replaced in place where they are used.
+
+#### Sibling sub-type
+
+Sibling sub-types are prefixed with a `bigsize` type identifier.
```
nkohen, line 91: no `bigsize` length requirement?

Messaging.md
```diff
@@ -53,13 +57,43 @@ Implementations MUST use a single connection per peer; contract messages (which
...
+The type identifiers are specific to a set of type variants, and can thus be reused across different sub-types.
+Type identifiers are defined within the type definitions, starting from `1` and increasing by `1`.
+
+#### Optional sub-type
+
+Optional sub-types are also prefixed with `bigsize` type identifier, but the identifiers are defined in the containing structures using the `(Optional: x)` notation where `x` is a number denoting the identifier, starting at `1` and increasing by `1`.
```
nkohen: ditto both above comments

Messaging.md
```diff
@@ -53,13 +57,43 @@ Implementations MUST use a single connection per peer; contract messages (which
...
+Optional sub-types should be gathered as the last fields of a type.
+In the absence of any optional sub-type in a structure, a zero byte `0x00` should be written to enable parsing.
```
nkohen: Isn't this redundant if we also support TLV streams?

Messaging.md
nkohen: If we are doing this kind of thing it might be nice to explicitly instantiate the super types and have

1. extends: [super type name]
2. type: 1
3. data: ...

Messaging.md
```diff
@@ -169,52 +203,42 @@ The type `rounding_intervals` is defined [here](NumericOutcome.md#rounding-inter
...
 
-1. type: 42786 (`oracle_info_v1`)
+1. type: 2
 2. data:
    * [`u16`:`threshold`]
    * [`u16`:`num_oracles`]
    * [`oracle_announcement`:`oracle_announcement_1`]
    * ...
    * [`oracle_announcement`:`oracle_announcement_num_oracles`]
+   * [`oracle_params`:`oracle_params`] (Optional: 1)
```
nkohen: nit: notationally I think I would prefer
```
[`Optional(oracle_params, 1)`:`oracle_params`]
```
or something like that

Messaging.md
```diff
@@ -379,25 +389,27 @@ See [the Oracle specifications](./Oracle.md#oracle-event) for more details.
 This type contains an `oracle_event` and a signature certifying its origination.
 See [the Oracle specifications](./Oracle.md#oracle-announcements) for more details.
 
-#### Version 0 `oracle_announcement`
+#### `oracle_announcement`
 
-1. type: 55332 (`oracle_announcement`)
-2. data:
+1. data:
+   * [`bigsize`: `version`]
```
nkohen: I'm not sure I understand, are these simple sub-types then? If so I don't get the point of a simple sub-type with a version number, wouldn't it make more sense to have a sibling sub-type with just one version?

Protocol.md
```diff
@@ -66,6 +66,7 @@ the funding transaction and CETs.
 
 1. type: 42778 (`offer_dlc_v0`)
 2. data:
+   * [`protocol_version`: `bigsize`]
    * [`byte`:`contract_flags`]
```
nkohen: seems a little redundant to have both of these

nkohen: I personally think the offer should signal the features it supports and requires which are two separate things, and then have a field in the accept with the feature set to be used, so for example you could have an offer that had a new feature enabled but not required where the accepter can choose whether or not to use it.

Protocol.md
```diff
@@ -66,6 +66,7 @@ the funding transaction and CETs.
 
 1. type: 42778 (`offer_dlc_v0`)
 2. data:
+   * [`protocol_version`: `bigsize`]
```
nkohen: nit: I think this is backwards, should be ```[`bigsize`:`protocol_version`]```

Protocol.md
```diff
@@ -81,6 +82,11 @@ the funding transaction and CETs.
    * [`u64`:`feerate_per_vb`]
    * [`u32`:`cet_locktime`]
    * [`u32`:`refund_locktime`]
+   * [`offer_tlvs`: `tlvs`]
```
nkohen, line 86: I take it none of `offer_tlvs`, `accept_tlvs` or `sign_tlvs` have any currently existing fields correct?

Messaging.md
```diff
@@ -53,13 +57,43 @@ Implementations MUST use a single connection per peer; contract messages (which
...
+#### Sibling sub-type
+
+Sibling sub-types are prefixed with a `bigsize` type identifier.
+The type identifiers are specific to a set of type variants, and can thus be reused across different sub-types.
+Type identifiers are defined within the type definitions, starting from `1` and increasing by `1`.
+
+#### Optional sub-type
+
+Optional sub-types are also prefixed with `bigsize` type identifier, but the identifiers are defined in the containing structures using the `(Optional: x)` notation where `x` is a number denoting the identifier, starting at `1` and increasing by `1`.
+Optional sub-types should be gathered as the last fields of a type.
```
nkohen: I'm not sure I understand the necessity for this requirement?

nkohen: Especially seeing as simple sub-types can presumably have optional fields?

Messaging.md
Tibo-lg: My thinking was that the odd/even requirements in LN are used for extensibility, and to determine whether a TLV or message is part of the base protocol or part of a higher layer (application specific) protocol. The way sibling types are defined, there is no case where they should be ignored, so I didn't see a good reason to use the odd/even requirements. If I am missing something and you have in mind a use case for it I'm happy to change.

Messaging.md
Tibo-lg: This follows somehow from the previous point. The `length` requirement is useful for parsing unknown types or things that should be ignored, but I couldn't think of how it would be useful here (again happy to change it if there is a good reason).

Messaging.md
Tibo-lg: My understanding and the way I think about things is that TLV are for extending the protocol, while optional sub-types are not. We could maybe use TLVs everywhere but I thought it would be better and simpler not to.

Messaging.md
Tibo-lg: Mmmm I feel `extends` is not really proper because there is no super type that is extended, rather there are multiple versions of the same field. I went with `implements` as it feels a bit more correct but let me know if you have a better idea.

Messaging.md
Tibo-lg: Sounds good updated

Messaging.md
Tibo-lg: You're right updated

Protocol.md
Tibo-lg: I personally see these two fields as providing different information.
`protocol_version` indicates which language I speak, and enables the receiver to determine whether they will be able to understand the content of the messages that are exchanged. Ideally in the future nodes should advertise which version of the protocol they support so that other peers will know how to communicate with them.
`contract_flags` are used to signal features as you mention.
Does that make sense?

Protocol.md
Tibo-lg: Thanks fixed

Protocol.md
Tibo-lg: correct

Messaging.md
Tibo-lg: You're right removed.

Messaging.md
Tibo-lg: Fixed thanks

Messaging.md
Tibo-lg: Fixed thanks

Messaging.md
nkohen: good by me

Messaging.md
```diff
@@ -11,55 +11,87 @@ All data fields are unsigned big-endian unless otherwise specified.
...
+#### Sibling sub-type
+
+Sibling sub-types are prefixed with a `bigsize` type identifier.
+The type identifiers are specific to a set of type variants, and can thus be reused across different sub-types.
+Type identifiers are defined within the type definitions, starting from `1` and increasing by `1`.
+
+#### Optional sub-type
+
+Optional sub-types are also prefixed with `bigsize` type identifier, but the identifiers are defined in the containing structures using the `Optional(type_name, x)` notation where `type_name` is the name of the type and `x` is a number denoting the identifier, starting at `1` and increasing by `1`.
+In the absence of any optional sub-type in a structure, a zero byte `0x00` should be written to enable parsing.
```
nkohen: Is this now a `0x00` for every absent optional field or for every empty group of consecutive fields or?

Tibo-lg: As discussed replaced the type identifier with a single byte indicating absence/presence of an optional field.

Messaging.md
```diff
@@ -11,55 +11,88 @@ All data fields are unsigned big-endian unless otherwise specified.
...
+We reuse the [Lightning Message Format](https://github.com/lightningnetwork/lightning-rfc/blob/master/01-messaging.md#lightning-message-format) for messages sent over the wire, the [Type-Length-Value Format](https://github.com/lightningnetwork/lightning-rfc/blob/master/01-messaging.md#type-length-value-format) (TLV) for message extensibility and extra features support, and a custom format for [sub-types](#sub-types).
+
+### Wire Messages
+
+Any encoded binary blob that can be sent over the wire will follow the Lightning Message Format.
+This means that wire messages are prefixed with a `u16` field (defined below) and their length is omitted from their encoding because the transport layer has the length in a separate un-encrypted field.
+
+### Type-Length-Value
+
+A TLV stream is defined for each wire message, and can be used to extend the protocol and for application specific features.
```
ariard, line 72: Browsing through the BOLTs again, `tlv_stream` are extensions to already defined messages, allowing to add optional fields in a backward compatible-way ?

```
Throughout the protocol, a TLV (Type-Length-Value) format is used to allow for the backwards-compatible addition of new fields to existing message types.
```

IIUC your proposal correctly it defines the content of any of our DLC messages as a TLV namespace itself ? Not sure if it's achieving it's purpose of bandwidth-saving as the type/length components of a `tlv_record` are redundant information for mandatory fields ?

Messaging.md
ariard: Instead of envisioning sibling type, we could define an even type for a field, servicing as reference and any future, alternative field as an odd one ?

Messaging.md
```diff
@@ -11,55 +11,88 @@ All data fields are unsigned big-endian unless otherwise specified.
...
+
+Wire messages (and sub-types themselves) contain embedded data structures.
+These sub-types can have three functions:
+* [Plain sub-types](#plain-sub-type) to factor out a number of fields to make specifications clearer,
+* [Sibling sub-types](#sibling-sub-type) to support multiple variants of a field,
+* [Optional sub-types](#optional-sub-type) to support optional fields.
+
+#### Plain sub-type
+
+Plain sub-types do not have any particular format, and their field can simply be replaced in place where they are used.
```
ariard, line 85: Can you give an example of "plain sub-type" ? I'm not sure I'm following on the replacement part...

Protocol.md
ariard, line 90: Maybe we should wait for the `channel_type` discussion to settle on the LN-side and piggy-back there, defining new DLC features as-yet-another-channel ? 

See https://github.com/lightningnetwork/lightning-rfc/pull/880 and this week LN meeting : https://gnusha.org/lightning-dev/2021-07-05.log

ariard (commented):
After reading this proposal, I'm leaning towards the LN-way for upgradeability/nesting.

If the message is flow is different, introduce a new message type. If the message flow is the same and we're extending message semantic, add a TLV field.

If we want nested data struct, I think we can achieve it the way onions (BOLT4) are done, versioning and new namespaces at each nesting entry ?

Messaging.md
ariard: But we could use TLVs everywhere and if you have option between A-or-B-but-at-least-one-them-must-be-present reject your message as a failure if none of them is present after deser ?

Protocol.md
ariard: In a Lightning/DLC world, I'm not sure if the `protocol_version` is making sense as new protocol subset can be introduced as new messages types and the question of whom to talk to is already solved with the feature bits space ?
I agree with the necessity of a `contract_flags`, see my other point mentioning the LN discussion on `channel_type`.

Messaging.md
Christewart: Aren't you assuming everything is a static size then? An example of something that has non deterministic size is ECDSA signatures. IIRC they can very anywhere in between 68 - 72 bytes in size. If you don't explicitly encode the size, you can't parse the message if i understand this correctly.

Protocol.md
```diff
@@ -81,6 +82,11 @@ the funding transaction and CETs.
...
+
+[//]: # (TODO: `protocol_version` should be advertised by DLC nodes in the future)
```
Christewart, line 88: is there any precedent for this in other deployed cryptocurrency protocols? Could you reference some resources you used for this if so? 

It seems like this isn't just a problem with our protocol, but every cryptocurrency protocol that has versioning.

Christewart: Oh sorry, this is a TODO nevermind.

Messaging.md
Tibo-lg: In general for fields that have variable length we have length prefixes in the type definition itself, like your example of ECDSA signature

Messaging.md
Tibo-lg: @ariard sorry I completely missed your comments. I don't really get what you mean here though, the point of sibling type is not for extensibility or upgradeability but just for cases where within a data structures, a field can have different types.

Protocol.md
Tibo-lg: For me `contract_flags` ad `protocol_version` are two separate things. The protocol version is for breaking changes in the protocol specifications and enabling nodes to see whether they are compatible with a version or not. Contract flags is for a given protocol version whether a node supports a feature or not.
I guess we could use contract version for both but it doesn't feel optimal to me.
Happy to discuss though, I don't have that strong of an opinion about how things should be, just that it's practical to have some sort of upgradeability mechanism.

Messaging.md
Tibo-lg: My point was to have `tlv_stream` for the exact same purpose as LN, so maybe the way I've written things is confusing?

Messaging.md
Tibo-lg: Added an example. I think the naming is probably not great. But basically the point is that before we had type prefixes for fields that were neither optional or sibling.

Protocol.md
Tibo-lg: Can you clarify what you would like to do here?

Protocol.md
Tibo-lg: I think having a protocol version is really common in any kind of protocol. It's for example used in the bitcoin p2p protocol: https://en.bitcoin.it/wiki/Protocol_documentation#version

Messaging.md
```diff
@@ -11,55 +11,90 @@ All data fields are unsigned big-endian unless otherwise specified.
...
+
+For example, the [funding input type](#the-funding_input-type) is a plain sub-type which does not require any particular prefix.
+
+#### Sibling sub-type
+
+Sibling sub-types are prefixed with a `bigsize` type identifier.
+The type identifiers are specific to a set of type variants, and can thus be reused across different sub-types.
+Type identifiers are defined within the type definitions, starting from `1` and increasing by `1`.
```
benthecarman: why not start at `0`?

Messaging.md
```diff
@@ -364,40 +386,42 @@ Note that `outcome_i` is the outcome value itself and not its hash that will be
...
-2. data:
+1. data:
    * [`u16`:`nb_nonces`]
    * [`nb_nonces*x_point`:`oracle_nonces`]
    * [`u32`:`event_maturity_epoch`]
    * [`event_descriptor`:`event_descriptor`]
    * [`string`:`event_id`]
 
-### The `oracle_announcement` Type
+### The `oracle_announcement` TLV
```
matthewjablack: I'm confused about the purpose of the `TLV` here

Should there be a separate `TLV` and message format? 

Currently `OracleAnnouncementV0` and `OracleAttestmentV0` are both TLVs (at least they're both implemented as TLVs in bitcoin-s and node-dlc)

(Also looks like `TLV` is not specified in #171 https://github.com/discreetlogcontracts/dlcspecs/pull/171/files#diff-b8a9c2888fe05c198758fbe393798cd0802b578f7eb07dbf82cc2c90616f97c3R402 )

benthecarman: +1 if nothing else is changing about the announcements I don't see the reasoning of changing the type

Tibo-lg: Sorry I confused myself. I think they should be `wire messages` as they can be broadcast directly (please see updated version). I guess we could define two formats but for the sake of simplicity I would personally prefer keeping a single one.

> +1 if nothing else is changing about the announcements I don't see the reasoning of changing the type

I was thinking since we are going to have breaking changes on these anyway we might as well make it so that the wire message types follow each others (I left an open space for the `close dlc`. Though now thinking about it I wonder if `close dlc` should actually have an odd type number 🤔

Messaging.md
matthewjablack: +1 Looks like #171 starts with 0

Messaging.md
matthewjablack: > Sorry I confused myself. I think they should be wire messages as they can be broadcast directly (please see updated version). I guess we could define two formats but for the sake of simplicity I would personally prefer keeping a single one.

+1 for having one format for sake of simplicity. Keeping it as a `wire message` makes sense to me

> Though now thinking about it I wonder if close dlc should actually have an odd type number 🤔

What would be the rationale for this? I'm guessing because it's optional?

Tibo-lg: > What would be the rationale for this? I'm guessing because it's optional?

Yes exactly, that's at least the semantics specified in BOLT 1: https://github.com/lightningnetwork/lightning-rfc/blob/master/01-messaging.md#lightning-message-format

Tibo-lg:
Fixed the sibling type numbers which where not consistent with the latest update.

Tibo-lg:
Noticed that we were sometimes using `BigSize` and sometimes `u16` for collection prefixes. Consolidated it to use `BigSize` everywhere as for the CetSignatures type it is probably safer than having a restriction on size.

Protocol.md
Christewart: it would be nice to have a list of known protocol_versions. My understanding is the first one is introduced in this PRs test vectors. It would be nice to have it listed somewhere in this doc

Protocol.md
```diff
@@ -66,6 +66,7 @@ the funding transaction and CETs.
 
 1. type: 42778 (`offer_dlc_v0`)
 2. data:
+   * [`bigsize`: `protocol_version`]
```
Christewart: I believe `u16`:`num_funding_inputs` is a `bigsize` in the test vectors (specifically `enum_single_oracle_test.json`)

Messaging.md
```diff
@@ -262,10 +291,9 @@ all of the `negotiation_fields` nested here must be version 0 or 1.
...
 This type contains information about a specific input to be used in a funding transaction, as well as its corresponding on-chain UTXO.
 
-#### Version 0 `funding_input`
+#### `funding_input`
 
-1. type: 42772 (`funding_input_v0`)
-2. data:
+1. data:
    * [`u64`:`input_serial_id`]
    * [`u16`:`prevtx_len`]
```
Christewart: I think this is a `bigsize` now, at least according to the `enum_single_oracle_test.json`

Protocol.md
```diff
@@ -185,7 +191,8 @@ and closing transactions.
    * [`u64`:`change_serial_id`]
    * [`cet_adaptor_signatures`:`cet_adaptor_signatures`]
    * [`signature`:`refund_signature`]
-   * [`negotiation_fields`:`negotiation_fields`]
+   * [`negotiation_fields`:`negotiation_fields`] (Optional: 1)
+   * [`accept_tlvs`: `tlvs`]
```
Christewart, line 205: `bigsize` here too i think https://github.com/discreetlogcontracts/dlcspecs/pull/163/files#diff-e0f5b925f91a1c09c6daf26f9a7d28816cb9ff9f08863faca719b7ee0a1cc065R188

Protocol.md
Tibo-lg: Thanks I had forgotten to update it, fixed.

Messaging.md
Tibo-lg: Yes thanks fixed.

Protocol.md
Tibo-lg: Sorry not sure which field you're referring to in this comment?

Christewart: `num_funding_inputs`

Tibo-lg: Thanks!

Tibo-lg:
* Switched protocol version from `BigSize` to `u32` as I find it cleaner.

* Fixed the missing protocol version in the test vectors.

Messaging.md
```diff
@@ -377,12 +402,13 @@ See [the Oracle specifications](./Oracle.md#oracle-event) for more details.
...
 This type contains an `oracle_event` and a signature certifying its origination.
+As oracle announcements can be broadcast directly, they are encoded as [wire messages](#wire-messages).
 See [the Oracle specifications](./Oracle.md#oracle-announcements) for more details.
 
-#### Version 0 `oracle_announcement`
+#### `oracle_announcement`
 
-1. type: 55332 (`oracle_announcement`)
-2. data:
+1. type: 42786
```
Christewart: why are these changed? This seems to break backwards compatability for no good reason?

Tibo-lg: Mainly just to make the range of types we use smaller. I don't mind reverting this, I assume you would prefer?

Christewart: Yes please. It's such a critical data structure in existing deployed applications such as our oracle explorer and Atomic Finance's app. I don't have any problems with restricting the range of types with future announcements we define in the spec.

Tibo-lg: Updated!

Tibo-lg:
Added a commit that reverts the oracle messages to the previous version, and updated the test vectors.

Messaging.md
```diff
@@ -364,11 +392,13 @@ Note that `outcome_i` is the outcome value itself and not its hash that will be
...
+**For backward compatibility reasons, this type is currently serialized as a TLV. It is expected to be changed to the new serialization format in a near future.**
+
+#### `oracle_event`
 
-1. type: 55330 (`oracle_event_v0`)
-2. data:
-   * [`u16`:`nb_nonces`]
+1. type: 55330
+1. data:
+   * [`bigsize`:`nb_nonces`]
```
Christewart: this could theoretically break compatability i believe (nested inside of oracle announcement). I'm not sure if it matters in practice though.

Messaging.md
```diff
@@ -336,11 +360,14 @@ Witnesses should be sorted by the `input_serial_id` sent in `funding_input` defi
...
+
+#### `enum_event_descriptor`
 
-1. type: 55302 (`enum_event_descriptor_v0`)
-2. data:
-   * [`u16`:`num_outcomes`]
+1. implements: `event_descriptor`
+1. type: 55302
+1. data:
+   * [`bigsize`:`num_outcomes`]
```
Christewart: This also can break compatability (nested inside of oracle announcement). This is an issue with the existing test vector `enum_single_oracle_test.json`. 

When trying to read your `serialized` string, and then write it back and compare the our serialized string vs your serialized string it seems that the length of the oracle announcement is slightly different than what we get. I believe this is because `BigSize` is not deterministic with it's sizing, while `uint16` is (always 2 bytes). 

Here is the branch to reproduce: 

https://github.com/Christewart/bitcoin-s-core/commit/9e2123dbbca17f34f382d72413b1440e4d5178bc

and you can run this command inside of `sbt` to see the results

>coreTestJVM/testOnly *ContractInfoV0Test* -- -z  "conform to the enum single test vector"

If you look carefully at the image below, you see it's the `length` of the TLV that seems to be off by a few bytes.

Screenshot from 2021-11-10 09-52-51

Tibo-lg: Thanks I had forgotten to revert that part, I updated the test vectors I hope this time it's ok!

Messaging.md
Tibo-lg: Reverted to u16 (I think it does matter)

Christewart:
All contract ids are broken for existing DLCs by this change. Database migrations are required when implementing this change, else you will have invalid contract ids in your database that are no longer correct for this version of the specification.

Tibo-lg:
> All contract ids are broken for existing DLCs by this change. Database migrations are required when implementing this change, else you will have invalid contract ids in your database that are no longer correct for this version of the specification.

Should we remove the requirement that contract ID must be a hash of offer message? I don't recall the reason for choosing to do that originally, but it doesn't feel like bringing much and we might have this issue again in the future.

Christewart:
> > All contract ids are broken for existing DLCs by this change. Database migrations are required when implementing this change, else you will have invalid contract ids in your database that are no longer correct for this version of the specification.
> 
> Should we remove the requirement that contract ID must be a hash of offer message? I don't recall the reason for choosing to do that originally, but it doesn't feel like bringing much and we might have this issue again in the future.

Perhaps. We needed to implement #143 in our codebase anyway which breaks contractIds as well (this is our fault for delaying implementation). Because of this i'll have to write the migration code regardless.

@matthewjablack what does this mean for your implementation? Do you guys use `contractId` quite a bit in your codebase?

EDIT: Here is our issue for this https://github.com/bitcoin-s/bitcoin-s/issues/3969#issue-1097199176

Tibo-lg:
Rebased on master (realized it was missing the hyperbola update)

Tibo-lg:
Updated to add the `temporary_contract_id` to the offer message as agreed during last spec meeting.

PayoutCurve.md
```diff
@@ -69,7 +69,7 @@ In this section we detail the TLV serialization for a general `payout_function`.
 
 1. type: 42790 (`payout_function_v0`)
```
Christewart: I think this needs to be updated as I don't believe its a TLV in the #163 test vectors?

Tibo-lg: Thanks! Fixed everything in that file, as well as the `RoundingIntervals` that was also not updated.

PayoutCurve.md
```diff
@@ -158,9 +157,8 @@ where a spline is made up of polynomial pieces so that the resulting interpolati
 
 #### Polynomial Serialization
 
-1. type: 42792 (`polynomial_payout_curve_piece`)
-2. data:
-   * [`u16`:`num_pts`]
+1. data:
```
Christewart, line 162: Shouldn't this be a sub type? And then you can easily parse the difference between `Polynomial`/`Hyperbola` curve pieces easily?

Tibo-lg: Yes you're right, fixed.

Messaging.md
matthewjablack: I noticed that in the latest version of #163 `OracleAnnouncement`, `OracleEvent`, `EnumEventDescriptor` and `DigitDecompositionEventDescriptor` are all `wire messages`

This makes sense for `OracleAnnouncement`, since it's likely announcements and attestments would be communicated over the transport layer. However, shouldn't the wire messages inside `OracleAnnouncement` actually be a TLVs since it's unlikely for them to be communicated over the transport layer.

Tibo-lg: The reason was that it was decided not to touch any oracle related message in this PR so that no update to oracle software is required for now, and then make all oracle related changes as part of #167 at once.

NumericOutcome.md
```diff
@@ -66,8 +66,7 @@ up in the case of a tie.
 
 #### Rounding Interval Serialization
 
-1. type: 42788 (`rounding_intervals_v0`)
-2. data:
+1. data:
    * [`u16`:`num_rounding_intervals`]
```
Christewart: @Tibo-lg  Is this consistent with your implementation?

Christewart: Addressed in 70032e32d69689547ed9a96faa8e366bbf039cd5

Christewart: I think there is another bug here, I believe this is prefixed by a `bigsize` in your implementation @Tibo-lg rather than a `u16`.

Messaging.md
```diff
@@ -367,10 +395,12 @@ Note that `outcome_i` is the outcome value itself and not its hash that will be
 This type contains information provided by an oracle on an event that it will attest to.
 See [the Oracle specifications](./Oracle.md#oracle-event) for more details.
 
-#### Version 0 `oracle_event`
+**For backward compatibility reasons, this type is currently serialized as a TLV and uses u16 for collection prefix. It is expected to be changed to the new serialization format in a near future.**
```
Christewart, line 398: I don't believe we need this for oracle event. We only need it for oracle announcements I believe. cc @matthewjablack 

Also I believe your test vectors don't keep the TLV around either. This is what I am parsing: 

```
00fdd824fd01ceba1b816b6563f9d5...
```

The `fdd824` is the TLV prefix on our current announcements [example](https://oracle.suredbits.com/announcement/b480fa0f0a3bad90a098284e4602dd14a85bc3d4df218e37bceebef3803fb161)

Christewart: Nevermind, this is incorrect. I thought the data structures was `OracleEvent(OracleAnnouncement)` rather than reality which is `OracleAnnouncement(OracleEvent)` :man_facepalming:

NumericOutcome.md
Tibo-lg: Yes in that case I should have updated the spec to be consistent with other collection prefixes. Updated now.

Protocol.md
```diff
@@ -181,17 +194,18 @@ and closing transactions.
    * [`point`:`funding_pubkey`]
    * [`spk`:`payout_spk`]
    * [`u64`:`payout_serial_id`]
-   * [`u16`:`num_funding_inputs`]
+   * [`bigsize`:`num_funding_inputs`]
```
Christewart, line 198: to make this change we need to either 

1. Add a version to the top of the `accept_dlc_tlv_v0`
2. revert this back to u16

The reason this is needed is to preserve backwards compatability with the previous version of the spec. This is how i've implemented the backwards compatible parse currently

```scala
    //parser for `offer_tlv`
    val fundingInputs = {
      if (protocolVersionOpt.isEmpty) {
        //means we are working with the old version of the spec
        iter.takeU16PrefixedList(() => iter.take(FundingInputV0TLV))
      } else {
        //means we are using version 1 of the spec
        iter.takeBigSizePrefixedList(() => iter.take(FundingInputV0TLV))
      }
    }
```

parser for the accept_tlv that is an attempt at backwards compatibility 

```scala
      //accept_tlv
      //
      //we don't have a version # here so we don't know what version we are trying to parse?
      //I suspect this code in the try block will not always cause an exception
      //so this is truely a guess.
      try {
        iter.takeU16PrefixedList(() => iter.take(FundingInputV0TLV))
      } catch {
        case scala.util.control.NonFatal(_) =>
          iter.takeBigSizePrefixedList(() => iter.take(FundingInputV0TLV))
      }
```

The crux of the issue is that if funding inputs are prefixed with a `u16` as required in the old version of the spec, a `bigsize` can be mistake as a `u16`. `bigsize` is required by the new version of the spec, which gives us ambiguity in the parser.

Tibo-lg: I don't mind adding a version at the top of the message, but can't you just keep the version from the offer message?

Christewart: That doesn't work for a few reasons

1. Its a layer violation to require a database read just to parse a tlv correctly.
2. You cannot guarantee you will parse the `accept_tlv` since you don't know which serialization format it uses to obtain information to _make_ the database read. I guess you might be able to parse the first 32 bytes (`tempContractId`) to parse just the 32 bytes, make the database read, and then determine which format the offer had, but this is very cumbersome. 

Lets keep the messages -- at the binary serialization level -- context free so that you can parse the `offer`, `accept`, and `sign` tlvs independent of one another.

Tibo-lg: Added a `protocol_version` field to the accept message and updated the test vectors. I wonder if it wouldn't be better to also add one to the sign message, what do you think?

Christewart: > I wonder if it wouldn't be better to also add one to the sign message, what do you think?

I think its probably a good idea. AFAIK, the `sign_tlv` does not have the same sort of ambiguity (`funding_inputs`) that `offer`,`accept` have so it isn't strictly necessary _now_, but it seems prudent to future proof a bit.

Tibo-lg: > it seems prudent to future proof a bit.

Yeah I feel like that too. Updated docs and test vectors accordingly.

Christewart:
So this PR on bitcoin-s passes all test vectors in this PR: 

I believe we have achieved the 2 impl requirement, which means we may want to merge this? 👀 👀 

https://github.com/bitcoin-s/bitcoin-s/pull/3859

Christewart:
As discussed at the June meeting, we have 2 implementations now so merging this.

PR #167, opened 2021-06-08
v0 Oracle Changes

nkohen:
As per our spec meeting discussions, this PR:

1. Splits the oracle key into separate announcement and attestation keys
2. Adds Schnorr signatures by the nonce keys to the announcement message
3. Allows for range timestamps in oracle events


Messaging.md
```diff
@@ -359,18 +360,33 @@ Note that `outcome_i` is the outcome value itself and not its hash that will be
...
+
+#### Fixed `oracle_event_timestamp`
+
+1. type: 55348 (`oracle_event_timestamp_fixed`)
+2. data:
+   * [`u32`:`expected_time_epoch`]
+
+#### Range `oracle_event_timestamp`
+
+1. type: 55350 (`oracle_event_timestamp_fixed`)
```
Tibo-lg: I think you forgot to update the name.

Messaging.md
```diff
@@ -379,15 +395,21 @@ See [the Oracle specifications](./Oracle.md#oracle-event) for more details.
...
 
-#### Version 0 `oracle_announcement`
+#### Version 1 `oracle_announcement`
 
-1. type: 55332 (`oracle_announcement`)
+1. type: 55354 (`oracle_announcement`)
 2. data:
-   * [`signature`:`annoucement_signature`]
-   * [`x_point`:`oracle_public_key`]
+   * [`signature`:`annoucement_pok_signature`]
```
Tibo-lg: I guess pok means proof of knowledge? Unless I missed it it'd be nice to write it down.

Messaging.md
```diff
@@ -379,15 +395,21 @@ See [the Oracle specifications](./Oracle.md#oracle-event) for more details.
...
+   * [`signature`:`attestation_pok_signature`]
+   * [`u16`:`nb_nonces`]
+   * [`nb_nonces*signature`:`nonce_pok_signatures`]
+   * [`x_point`:`oracle_announcement_public_key`]
+   * [`x_point`:`oracle_attestation_public_key`]
+   * [`nb_nonces*x_point`:`oracle_nonces`]
```
Tibo-lg: nit: might be good to have that before the `nonce_pok_signatures` in case someone would want to do de-serialization and validation at the same time (I don't think I would but it also seems like a tiny change)

Messaging.md
```diff
@@ -379,15 +395,21 @@ See [the Oracle specifications](./Oracle.md#oracle-event) for more details.
...
    * [`oracle_event`:`oracle_event`]
 
-where `signature` is a Schnorr signature over a sha256 hash of the serialized `oracle_event`, using the tag `announcement/v0`.
+where all `pok_signature`s are Schnorr signatures using the tag `announcement/v0` over a sha256 hash of the serialized `x_point`s concatenated in order and concatenated with `oracle_event`.
```
Tibo-lg: Would be good to add some rationals about why we need these pok.

Messaging.md
```diff
@@ -399,14 +421,10 @@ See [the Oracle specifications](./Oracle.md#oracle-attestations) for more detail
 1. type: 55400 (`oracle_attestation_v0`)
 2. data:
     * [`string`:`event_id`]
-    * [`x_point`:`oracle_public_key`]
+    * [`x_point`:`oracle_attestation_public_key`]
```
Tibo-lg: Similarly would be good to mention somewhere in text that oracle needs different keys.

Messaging.md
```diff
@@ -379,15 +395,21 @@ See [the Oracle specifications](./Oracle.md#oracle-event) for more details.
...
+   * [`u16`:`nb_nonces`]
+   * [`nb_nonces*signature`:`nonce_pok_signatures`]
+   * [`x_point`:`oracle_announcement_public_key`]
+   * [`x_point`:`oracle_attestation_public_key`]
+   * [`nb_nonces*x_point`:`oracle_nonces`]
    * [`oracle_event`:`oracle_event`]
 
-where `signature` is a Schnorr signature over a sha256 hash of the serialized `oracle_event`, using the tag `announcement/v0`.
+where all `pok_signature`s are Schnorr signatures using the tag `announcement/v0` over a sha256 hash of the serialized `x_point`s concatenated in order and concatenated with `oracle_event`.
+That is to say, the signature hash is of the message `oracle_announcement_public_key||oracle_attestation_public_key||oracle_nonces||oracle_event`.
```
LLFourn: My thoughts below :)

1. I think I've expressed this before -- sorry if I'm going over stuff that has already been discussed between you guys -- but I think the announcement should not have the oracle's public keys in it. They must be known beforehand. IIRC they are just being used as identifiers? I still think that is a dangerous choice but even in this case the `attestation_pok_signature` should not be in the announcement. The `attestation_pok_signature` should have been verified when the user originally saved the oracle's public key not at the time of any particular announcement.

2.  I think what was there before with `announcement_signature` made more sense because it's not a `pok` it's just a signature. Having all the oracle crypto data in `oracle_event` and having the signature over all of it in `oracle_announcement` made more sense to me.

3. Optional extra: IMO it's fine to do MuSig to aggregate all the `nonce_pok_signatures` into a single thing. Perhaps that is more specification work than you want to do now though. The reason I think it's fine is that the plan is to use MuSig to do cross-input aggregation which implies that it is secure for this purpose as well in a kind of *proof of possession* model which I think is reasonable here.

nkohen: Left the announcement key as an identifier, but removed attestation key and sig from announcement and moved it to a new `oracle_keys` message.

nkohen:
I've added a new `oracle_keys` message but I'm not sure I'm the best person to write in `Oracle.md` about how it is used, and likewise to write rationales for why we need two public keys and all the proofs of knowledge. @LLFourn would you be willing to write at least these last two things?

Messaging.md
```diff
@@ -359,35 +360,76 @@ Note that `outcome_i` is the outcome value itself and not its hash that will be
...
 2. data:
    * [`signature`:`annoucement_signature`]
-   * [`x_point`:`oracle_public_key`]
+   * [`u16`:`nb_nonces`]
+   * [`nb_nonces*signature`:`nonce_pok_signatures`]
+   * [`x_point`:`oracle_announcement_public_key`]
    * [`oracle_event`:`oracle_event`]
 
-where `signature` is a Schnorr signature over a sha256 hash of the serialized `oracle_event`, using the tag `announcement/v0`.
+where both the `announcement_signature` and the proof-of-knowledge `signature`s are Schnorr signatures over a sha256 hash of the serialized `oracle_event`, using the tag `announcement/v1`.
```
benthecarman: I don't see why we need to update the tag here

Messaging.md
```diff
@@ -359,35 +362,77 @@ Note that `outcome_i` is the outcome value itself and not its hash that will be
...
    * [`nb_nonces*x_point`:`oracle_nonces`]
-   * [`u32`:`event_maturity_epoch`]
+
+   * [`oracle_event_timestamp`:`timestamp`]
    * [`event_descriptor`:`event_descriptor`]
    * [`string`:`event_id`]
 
+### The `oracle_keys` Type
+
+This type contains static oracle information and can be used to import trusted oracles into wallets.
```
ariard, line 445: Do you mean static in the sense of fixed-for-this-outcome-announcement-lifecycle ? If so add requirement that keys should be saved in consequence by the DLC node.

Messaging.md
```diff
@@ -359,35 +362,77 @@ Note that `outcome_i` is the outcome value itself and not its hash that will be
...
+
+#### Version 0 `oracle_keys`
+
+1. type: 55356 (`oracle_keys`)
+2. data:
+   * [`x_point`:`announcement_public_key`]
+   * [`x_point`:`attestation_public_key`]
```
ariard: If `announcement_public_key` == `attestation_public_key`, MUST reject as invalid ?

Messaging.md
```diff
@@ -359,35 +362,77 @@ Note that `outcome_i` is the outcome value itself and not its hash that will be
...
+1. type: 55348 (`oracle_event_timestamp_fixed`)
+2. data:
+   * [`u32`:`expected_time_epoch`]
+
+#### Range `oracle_event_timestamp`
+
+1. type: 55350 (`oracle_event_timestamp_range`)
+2. data:
+   * [`u32`:`earliest_expected_time_epoch`]
+   * [`u32`:`latest_expected_time_epoch`]
```
ariard, line 429: Would be cool to add a rational, sum up of the ml discussions here :)

https://mailmanlists.org/pipermail/dlc-dev/2021-February/000015.html

Messaging.md
```diff
@@ -359,35 +362,106 @@ Note that `outcome_i` is the outcome value itself and not its hash that will be
...
+   * [`x_point`:`announcement_public_key`]
+   * [`x_point`:`attestation_public_key`]
+   * [`string`:`oracle_name`]
+   * [`string`:`oracle_description`]
+   * [`u32`:`timestamp`]
+   * [`signature`:`announcement_pok_signature`]
+   * [`signature`:`attestation_pok_signature`]
+
+where both proof-of-knowledge signatures are Schnorr signatures of over a sha256 hash of
+`announcement_public_key || attestation_public_key || oracle_name || oracle_description`
```
matthewjablack: What's the purpose of `timestamp`? Shouldn't it be included in msg used for sha256hash for pok signatures?

Messaging.md
```diff
@@ -359,35 +362,104 @@ Note that `outcome_i` is the outcome value itself and not its hash that will be
...
+
+1. type: 55356 (`oracle_keys`)
+2. data:
+   * [`x_point`:`announcement_public_key`]
+   * [`x_point`:`attestation_public_key`]
+   * [`string`:`oracle_name`]
+   * [`string`:`oracle_description`]
+   * [`u32`:`timestamp`]
+   * [`signature`:`announcement_pok_signature`]
+   * [`signature`:`attestation_pok_signature`]
```
LLFourn: I don't think `announcement_pok_signature` is the right name because it's purpose is not a proof of knowledge. It's purpose is to link the other details to this announcement key. Should just be `oracle_metadata_signature` or something.

`attestation_pok_signature` is not guaranteeing any security property -- rogue nonce attacks are easier than the rogue key attack here. We decided not to address this security flaw in this revision so there's not point in having it here. I think we should remove it.

In the future, I think the correct solution is to include a proof of possession via `MuSig(attestation_key, nonce_0, nonce_1 .. nonce_n )` in each announcement.

Messaging.md
```diff
@@ -359,35 +362,104 @@ Note that `outcome_i` is the outcome value itself and not its hash that will be
...
+using the tag `oraclekeys/v0`.
+
+#### Rationale
+
+Separate public keys are required for announcement and attestation because the attestation scheme
+used in the current DLC spec reduces the security of non-attestation Schnorr signatures issued by the
+attestation keys and hence a separate announcement key is required.
+
+Proof-of-knowledge signatures are used to force the two keys to authenticate each other as well as the
+rest of the data in the `oracle_keys` message.
```
LLFourn: It's not necessary for the two keys authenticate each other. The announcement key is the master key which is in charge of authenticating other keys.

Oracle.md
```diff
@@ -6,6 +6,9 @@ For the purpose of these specifications, an event is a digital representation of
 An [oracle](./Introduction.md#Oracle) is an entity that commits to publishing one or more signatures over a (number of) event outcome(s) ahead of time by releasing one or more [R-values](./Introduction.md#R-value) as well as necessary information for two parties to build a set of [CETs](./Introduction.md#Contract-Execution-Transaction-(CET)).
 This necessary information is committed to in a so-called [_event descriptor_](#Event-descriptor) that will be further detailed in this document.
 
+Before an oracle makes any announcements or attestations, it should first publish an [`oracle_keys`](Messaging.md#the-oracle_keys-type) message for
+clients to be able to import a new oracle (or new oracle attestation keys) into their trusted keychain.
+
```
LLFourn, line 11: Mention that the attestation key will always be included in the announcement anyway but that client SHOULD check this against the `oracle_keys` message and abort if they don't agree.

Messaging.md
```diff
@@ -359,35 +362,104 @@ Note that `outcome_i` is the outcome value itself and not its hash that will be
...
+   * [`x_point`:`announcement_public_key`]
+   * [`x_point`:`attestation_public_key`]
+   * [`string`:`oracle_name`]
+   * [`string`:`oracle_description`]
+   * [`u32`:`timestamp`]
+   * [`signature`:`announcement_pok_signature`]
+   * [`signature`:`attestation_pok_signature`]
+
+where both proof-of-knowledge signatures are Schnorr signatures of over a sha256 hash of
+`announcement_public_key || attestation_public_key || oracle_name || oracle_description || timestamp`
```
LLFourn: It'd be better to just lift the announcement_public_key and signature out of this structure and sign the standard serialization of the inner structure. It is in general bad practice to naively concatenate things when some of the things are variable length (strings) because you don't have a marker for when one starts and the other ends. The serialization naturally gives this to you so it's easier to just sign that.

LLFourn:
Meeting note:  attestation schemes should be declared separately from the event information. Taking this as an example: 

```
$ curl 'https://h00.ooo/x/BitMEX/BXBT/2022-01-12T01:00:00.price?n=20' | jq .announcement.oracle_event.data -r  |jq

 {
	"id": "/x/BitMEX/BXBT/2022-01-12T01:00:00.price?n=20",
	"expected-outcome-time": "2022-01-12T01:00:00",
	"descriptor": {
		"type": "digit-decomposition",
		"is_signed": false,
		"n_digits": 20,
		"unit": null
	},
	"schemes": {
		"olivia-v1": {
			"nonces": [
				"ea4c2c81a65e4df1c98515bf2af6a0e63ed086f803e4637130447056f2ff15c8",
				"e7a46ba921fab39fe9e953e430f72e7325c43aba4d181517b9078dfc09637054",
				"f9dce3712dfa25c3ab516bef99c5e828b349e53a134321938612f5b5af9dd4bf",
				"74ef83448b8fc533a1dc33f1f58d1c0bd6cfb086f9369d627d917884ccc21da7",
				"63c0e43218a9e31e0ff5b34ab492a0e9fa100eebfc888c7efc51791a55cd8283",
				"9e0c36a0607c16d6d6284c640a8ac00ab65a772c3fa0981be022520a9f44e47a",
				"e288c14d070f463a7d71bc330b4786c574aa8be02f7e7ad00e4fc9a6bdcbb32b",
				"e758b66c5df5ef161f00081fe87685230a0b2846863cf3da00828d112b71220f",
				"1ed08c216fddc19622553925ef9d2753715c7c01201a9a47ae4233a4878cc030",
				"8e5bc962f77346e721f25fe7871a03e95bf38a86e564c5729c86dc2fc5eb6871",
				"68843c83a7c9829f6ef5c4ae77067f9d7f72703ee9b8c6d5498a873180b15574",
				"aee0085e8c98a2afbc7d43d5774e145cfa0168d9e22506dc79fa1b74718e0043",
				"e4c038f95dac570ffe276c16a3518983abcc8216e9d318de39ca42c70d54ea3e",
				"cb4f5cc032ca1e343229256bb0d911b0e04ec2f409d0fa837cedd973ac45b742",
				"1e63f2116013988df505b26ddebec1adaeb41c7a2b55baf375f599431b5fa242",
				"1792e2fed1c346900702dcc208e286ea8a6865975cd961c834a275c505072247",
				"a5672e32a331cc360317a7329fc7f70d7843586ab4a24b961336183cc764bef9",
				"52272eca966fe72012abffcc481276dce404c27e4db783374a5a7c001f9477d2",
				"98649a7620abd662e2ff699659f58b7cf19d9ad4e5257180cc36ba79fff97c68",
				"92dddc74b7491067fab985962c60b8e0ce43e820e35405f6ac2a577e4b6eca6e"
			]
		},
		"ecdsa-v1": {}
	}
} 
```

so in the oracle_event you have the descriptor and other metadata like times on one level and then you have a list of attestation schemes and their parameters. Note that nonces and public keys are always parameters to a certain attestation scheme (my oracle is still missing the public key which should be in there with the nonces).

Christewart (commented):
@Tibo-lg @nkohen I believe this needs to be updated with the format in #163. We need to figure out which messages we want to keep as tlvs and what data structures we need to have keep the format in #163

Tibo-lg:
> @Tibo-lg @nkohen I believe this needs to be updated with the format in #163. We need to figure out which messages we want to keep as tlvs and what data structures we need to have keep the format in #163

Unless there is strong arguments, I think it'd be best to move everything to the same format as #163. I can update this PR if we agree on that (or probably make a new one as I'm not sure I can push to @nkohen branch).

Christewart:
> > @Tibo-lg @nkohen I believe this needs to be updated with the format in #163. We need to figure out which messages we want to keep as tlvs and what data structures we need to have keep the format in #163
> 
> Unless there is strong arguments, I think it'd be best to move everything to the same format as #163. I can update this PR if we agree on that (or probably make a new one as I'm not sure I can push to @nkohen branch).

I think it comes down to this: 

Do we expect announcements/attestations to be sent over the network directly to a peer, or do we always expected them to bet fetched from a 3rd party service like oracle.suredbits.com ? If we think the former, we should keep TLVs, if we think the latter, I think it might be ok to remove them (although I do think this makes parsing harder for the 3rd party service, there is no easy identifier for what the thing is being submitted to the service. A TLV is unique, whereas a subtype is not).


EDIT: 

Thinking about this more, my own standard I laid out above is sort of violated already. We host `contract_info` on oracle.suredbits.com, for instance here is a list of numeric contract infos: https://oracle.suredbits.com/contract/numeric

Since contract_info no longer has a TLV prefix, they cannot be easily identified for parsing. In the case of contract info, we only support building them on the site (and not directly submitting them like we do oracle announcements/attesattions), so this is slightly different. Removing TLVs seems to really break down modularity and ease of parsing of independent pieces of the protocol.

Tibo-lg:
> Do we expect announcements/attestations to be sent over the network directly to a peer, or do we always expected them to bet fetched from a 3rd party service like oracle.suredbits.com ? If we think the former, we should keep TLVs, if we think the latter, I think it might be ok to remove them (although I do think this makes parsing harder for the 3rd party service, there is no easy identifier for what the thing is being submitted to the service. A TLV is unique, whereas a subtype is not).

I would think that we should at least plan for enabling peers to exchange announcements/attestations directly. But in that case I think they'd need to be in wire message format, not TLV?

Christewart:
> 

[I'm not sure how you have implemented the wire message format, but in our code base the wire message format uses the nested TLV.](https://github.com/bitcoin-s/bitcoin-s/blob/8c5288d75833d56d388247cdc928cf6694c46f46/core/src/main/scala/org/bitcoins/core/protocol/tlv/LnMessage.scala#L19) 

```scala
case class LnMessage[+T <: TLV](tlv: T) extends NetworkElement {
  require(tlv.tpe.toLong <= 65535L, s"LN Message format requires UInt16 types")
  val tpe: UInt16 = UInt16(tlv.tpe.toInt)
  val payload: ByteVector = tlv.value
  override lazy val bytes: ByteVector = tpe.bytes ++ payload
  val typeName: String = tlv.typeName
}
```

The key line is

```scala
val tpe: UIn16 = UInt16(tlv.tpe.toInt)
```

Since the announcement is no longer a TLV, it cannot have a `tpe` which is _only_ defined for TLVs. How are you proposing we send non TLVs with the wire message format?

Tibo-lg:
I'm not sure what you mean by "nested TLV"? The wire message doesn't have any length prefix and uses a `u16` instead of a `bigsize` as written in [the specs](https://github.com/discreetlogcontracts/dlcspecs/blob/master/Messaging.md#message-format).

Christewart:
> [the specs](https://github.com/discreetlogcontracts/dlcspecs/blob/master/Messaging.md#message-format).

>**Meanwhile all typed sub-messages (which follow TLV format) will** 

^ What I am talking about. I can detail our implementation in Scala if you'd like but I think we have a misunderstanding of what the specification says. You adhere to this with all of the test vectors on #163 (which is why we are compatible), but want to break this convention -- IIUC -- on this PR. 

If we are intending to send announcements/attestations over the wire independently they will need to be TLVs

Tibo-lg:
Sorry I should have linked to the spec after #163 : https://github.com/Tibo-lg/dlcspecs-1/blob/serialization-update-proposal/Messaging.md#message-format

> You adhere to this with all of the test vectors on https://github.com/discreetlogcontracts/dlcspecs/pull/163 (which is why we are compatible), but want to break this convention -- IIUC -- on this PR.

This is because you wanted to keep the oracle message format intact so as not to have to update your oracle infrastructure before this PR. 

> If we are intending to send announcements/attestations over the wire independently they will need to be TLVs

I don't see why. But maybe it will be easier to discuss it during the spec meeting? It feels indeed like we are not understanding each others.

Christewart:
> Sorry I should have linked to the spec after #163 : https://github.com/Tibo-lg/dlcspecs-1/blob/serialization-update-proposal/Messaging.md#message-format
> 
> > You adhere to this with all of the test vectors on #163 (which is why we are compatible), but want to break this convention -- IIUC -- on this PR.
> 
> This is because you wanted to keep the oracle message format intact so as not to have to update your oracle infrastructure before this PR.
> 
> > If we are intending to send announcements/attestations over the wire independently they will need to be TLVs
> 
> I don't see why. But maybe it will be easier to discuss it during the spec meeting? It feels indeed like we are not understanding each others.

>**A TLV stream is defined for each wire message**

How are you suggesting to parse an announcement if it isn't a TLV? IIUC, we will have the length of the announcement, but no unique identifier? What is the unique identifier for an announcement so that I can identify it as an announcement before parsing the payload?

Perhaps it would be useful to make this more concrete, can you give me an example announcement you are envisioning in wire message format?

Tibo-lg:
> IIUC, we will have the length of the announcement, but no unique identifier? What is the unique identifier for an announcement so that I can identify it as an announcement before parsing the payload?
 
Wire messages have a u16 type prefix, isn't that enough?

> Perhaps it would be useful to make this more concrete, can you give me an example announcement you are envisioning in wire message format?

There might be some things that need to be updated, but roughly the spec should look like it was before the commit that reverted the oracle related messages to the old format: https://github.com/discreetlogcontracts/dlcspecs/pull/163/commits/225e3b0693766c2d74442424249f83d044ef2316#diff-b8a9c2888fe05c198758fbe393798cd0802b578f7eb07dbf82cc2c90616f97c3R407

Concretely an announcement would be serialized similarly to an `offer_dlc` message, with first a `u16` type prefix, and then the other fields (which wouldn't be TLVs as they are now in #163). When embedding an announcement inside an `offer_dlc`, I think we agreed that we would just keep the wire message format as is to make things simple, and so you'd have also the type prefix (which you wouldn't actually need in the context of the offer message), and again the other fields of announcement following.

Christewart:
>with first a u16 type prefix

IMO, you are using a TLV but just not calling it a TLV :-). Or perhaps my mental model of TLVs is incorrect. But functionally speaking, I think we are in agreement.

An announcement needs a type identifier similar to `offer_dlc` in 163 right? 

Screenshot from 2022-06-06 06-37-11


https://github.com/discreetlogcontracts/dlcspecs/pull/163/files#diff-e0f5b925f91a1c09c6daf26f9a7d28816cb9ff9f08863faca719b7ee0a1cc065R67

If that is the case, I don't really care about definitions, we are referring to the same thing with different definitions. I'll do some research later to correct my definitions -- if they are wrong -- so we can use a common language to describe the protocol.

Tibo-lg:
> An announcement needs a type identifier similar to offer_dlc in 163 right?

Yes! The main difference between TLV and wire message format is that wire message don't have the length part (because you get that from the network layer already) and use `u16` instead of `bigsize` for the type prefix.

Also note that wire messages include TLV extensions which might make things confusing (I'm not sure we want oracle messages to have that though, but it's a detail). If you think something could be made clearer in the specs let me know!

Christewart:
I think the next question is how do announcements get serialized inside of other TLVs? I.e. `offer_dlc` has announcements inside of it, do you think it should be serialized in wire message format _inside_ of the offer?

Christewart:
> Yes! The main difference between TLV and wire message format is that wire message don't have the length part (because you get that from the network layer already) and use `u16` instead of `bigsize` for the type prefix.

What do you mean by "network layer"? Our length is serialized as part of the TLV, i.e. type length value.

Tibo-lg:
I mean that the buffer you receive from the network socket should have a certain length, so you don't need a field to tell you the same thing. You should already be doing that for other messages like offer I think? (Also if the message is wellformed you don't even need that information, you'll just be parsing the content and be able to retrieve all the information you expect)

Christewart:
Doesn't this fall apart when your messages are larger than a tcp frame? My understanding is we need things like #192 because our messages sizes may be too large, specifically with adaptor signatures. With TLVs, you _know_ what the payload length should be as its given to you in the very first tcp frame.

Christewart:
> You should already be doing that for other messages like offer I think?

This might be the source of some misunderstanding on my part, we do infer the length as you say here:

https://github.com/bitcoin-s/bitcoin-s/blob/8c5288d75833d56d388247cdc928cf6694c46f46/core/src/main/scala/org/bitcoins/core/protocol/tlv/LnMessage.scala#L32

My comment above stands, but this does make more sense now

Tibo-lg:
> Doesn't this fall apart when your messages are larger than a tcp frame? My understanding is we need things like #192 because our messages sizes may be too large, specifically with adaptor signatures. With TLVs, you _know_ what the payload length should be as its given to you in the very first tcp frame.

AFAIU the reason TLVs have a length field is so that you can ignore fields that you don't care or know about in extensions (and aren't mandatory). Having a length field in every single message might be one possible solution to message segmentation, though it wouldn't be my favorite because:
* it would be useless in many cases 
* It would be more difficult to handle disconnects/reconnects (if I disconnect while sending a segmented message, and restart sending the same message upon reconnecting, the receiving peer will not notice)

Roughly what #192 does is to have a length field but only for messages that need them. There might be better ways to do it (and actually just re-looking at it now I feel it can be even simpler) but I think that's a discussion that we should have in #192 (and I'd be really happy to have it :) !).

Christewart:
>     * it would be useless in many cases

This really isn't a compelling argument in my opinion, the point is that it is _needed_ in some cases to have a coherent idea of the message that is being sent to you across the wire. Perhaps you have a deeper understanding of networking than I do, but I don't understand how the networking layer can have knowledge about our protocol? 

>     * It would be more difficult to handle disconnects/reconnects (if I disconnect while sending a segmented message, and restart sending the same message upon reconnecting, the receiving peer will not notice)

I don't understand this. Why would I keep my peers bytes in my buffer if they disconnected from me? It could be days before they reconnect?

> Roughly what #192 does is to have a length field but only for messages that need them. There might be better ways to do it (and actually just re-looking at it now I feel it can be even simpler) but I think that's a discussion that we should have in #192 (and I'd be really happy to have it :) !).

Ok, at least we are in agreement that a length field is needed for some messages.

Tibo-lg:
Rebased on #163 and updated to match the new serialization @Christewart @nkohen please check.

@LLFourn I liked your idea of being able to support multiple attestation schemes, can you check if what I've done matches with what you had in mind?

LLFourn:
@Tibo-lg I think that looks ok. The one thing I'd consider doing is putting the attestation key with the attestation scheme data itself. Probably "oracle_keys" should just have the announcement key and each attestation scheme inside each announcement should declare its attestation key. Each scheme should use a different attestation key. Clients could then by convention require that the same attestation key is used for every announcement (of the same scheme).

Tibo-lg:
> @Tibo-lg I think that looks ok. The one thing I'd consider doing is putting the attestation key with the attestation scheme data itself. Probably "oracle_keys" should just have the announcement key and each attestation scheme inside each announcement should declare its attestation key. Each scheme should use a different attestation key. Clients could then by convention require that the same attestation key is used for every announcement (of the same scheme).

Makes sense, updated.

Messaging.md
```diff
@@ -339,80 +371,182 @@ Witnesses should be sorted by the `input_serial_id` sent in `funding_input` defi
...
+This type contains static oracle information and can be used to import trusted oracles into wallets.
+This information should be saved by DLC nodes.
+
+#### `oracle_keys`
+
+1. data:
+   * [`x_point`:`announcement_public_key`]
+   * [`string`:`oracle_name`]
+   * [`string`:`oracle_description`]
+   * [`u32`:`timestamp`]
```
Tibo-lg, line 454: @nkohen what is the use for this field? I think it'd be nice to have it explained somewhere.

Messaging.md
```diff
@@ -339,80 +371,182 @@ Witnesses should be sorted by the `input_serial_id` sent in `funding_input` defi
...
-2. data:
-   * [`u16`:`nb_nonces`]
-   * [`nb_nonces*x_point`:`oracle_nonces`]
-   * [`u32`:`event_maturity_epoch`]
+1. data:
+   * [`oracle_event_timestamp`:`timestamp`]
    * [`event_descriptor`:`event_descriptor`]
    * [`string`:`event_id`]
 
+### The `oracle_keys` Type
```
Tibo-lg: Would it make sense to rename this to `oracle_metadata` or something else? Since it contains quite more than just keys.

Messaging.md
```diff
@@ -363,92 +371,182 @@ Witnesses should be sorted by the `input_serial_id` sent in `funding_input` defi
...
+
+#### `oracle_keys`
+
+1. data:
+   * [`x_point`:`announcement_public_key`]
+   * [`string`:`oracle_name`]
+   * [`string`:`oracle_description`]
+   * [`u32`:`timestamp`]
+   * [`oracle_schemes`:`oracle_schemes`]
+   * [`signature`:`oracle_metadata_signature`]
```
Tibo-lg, line 456: @nkohen why is there a signature here and then one on the announcement? Isn't it redundant?

Messaging.md
nkohen: I figured this would just be nice to have in case some oracle going as `oracle_name` publishes a new public key they use now and now when the user imports the new message, they can see that this is a newer one by looking at the timestamp. Open to changing this, but the idea was to have this be the time at which the oracle first announced this pubkey.

Messaging.md
nkohen: So as it stands this message is static and can be imported by a wallet independently of an oracle announcement right? In this case it is important that the oracle metadata is what the oracle put down. If the keys are what you expect but someone takes the message and re-writes something about the scheme or description or such, this could be bad for wallet impls

Messaging.md
nkohen: agree, makes sense to call it that

Messaging.md
Tibo-lg: Mmmm ok thanks, I'll at least add the explanation and maybe we can discuss it next time.

Messaging.md
Tibo-lg: If it is supposed to be a standalone message, does it make sense to include it in every single announcement? I'll go ahead with that for now but I feel it's also worth discussing.

Messaging.md
Tibo-lg: I was wondering, what prevents anybody to just releasing a new struct with the name of an existing oracle but changing the public key (and using a newer timestamp)? How do you ensure that it is indeed the correct oracle?

Tibo-lg:
I pushed updated test vectors including these oracle changes.

The questions that I have left:
* Do we really need a timestamp for the metadata? (https://github.com/discreetlogcontracts/dlcspecs/pull/167#discussion_r896482864)
* Do we need the oracle metadata in each announcement? (https://github.com/discreetlogcontracts/dlcspecs/pull/167#discussion_r896286653)
*  Are we sure we don't want to give proof of knowledge for attestation key and nonces?
* (more cosmetic) Should we remove all the `Oracle` prefixes from the field names?  It feels redundant to me (and I think we don't add them consistently).

Tibo-lg:
Closes #183

Tibo-lg:
Added proof of knowledge in the `Schnorr` scheme (@LLFourn when you have time to check)

Messaging.md
```diff
@@ -363,92 +371,196 @@ Witnesses should be sorted by the `input_serial_id` sent in `funding_input` defi
...
+
+1. type: 0
+1. data:
+   * [`x_point`:`attestation_public_key`]
+   * [`bigsize`:`nb_nonces`]
+   * [`nb_nonces*x_point`:`oracle_nonces`]
+   * [`signature:`proof_of_knowledge`]
+
+where `proof_of_knowledge` ensures that the oracle knows the pre-image of the `attestation_public_key` and all `oracle_nonce`.
+It is generated following bip-musig2 [signing](https://github.com/jonasnick/bips/blob/musig2/bip-musig2.mediawiki#signing) and [signature aggregation](https://github.com/jonasnick/bips/blob/musig2/bip-musig2.mediawiki#partial-signature-aggregation) algorithms, using the attestation key as the first key and the nonces as the following keys, in the same order as provided in the data structure, and the message `dlcoraclepok`.
```
Christewart: So does this block implementation on this until all libraries support musig2?

Tibo-lg: I guess it's up for us to decide. The alternatives to musig is to do separate proofs of knowledge, which is fine but we'll end up with rather big announcements for numerical events, with a signature per nonce.

Note it's mainly about binding to the musig2 implementation, once you have the bindings it's quite little work (you can see my implementation [here](https://github.com/p2pderivatives/rust-dlc/commit/3bf2d7c6904ba7309154c4c30104ab55f525a063)

Christewart: Do you have a musig2 implementation available to you?

Also does @matthewjablack have a musig2 implementation available to him? 

We are _just_ implementing it: https://github.com/bitcoin-s/bitcoin-s/pull/4418

Christewart: Also this brings up stability questions of the current musig2 spec. For instance, I know Schnorr was deployed on the bitcoin cash blockchain years before schnorr got deployed into Bitcoin. 

I think things like `XOnlyPubKey`s weren't included in the schnorr spec, thus leading to incompatibilities between the two implementations. 

We could end up in a similar situation if we don't wait for the specification to mature. I guess lnd is deploying musig2 (although not incorporating it in the LN spec, so its easier to change for now!)

https://docs.bitcoincashnode.org/doc/bips/

Screenshot from 2022-06-27 07-37-37

TLDR, how stable is musig2?

matthewjablack: Taproot hasn't been added to bitcoinjs-lib let alone musig2

If we wanted to get access to musig2 we'd likely need to get it from secp256k1-zkp and update cfd/cfd-core

This is a little annoying since it would mean validation for Oracle Announcements would require cfd-dlc-js and couldn't be done in native JS as it's done currently https://github.com/AtomicFinance/node-dlc/blob/master/packages/messaging/lib/messages/OracleAnnouncementV0.ts#L62

Tibo-lg: @Christewart
> Do you have a musig2 implementation available to you?

Yes, WIP but it's there: https://github.com/ElementsProject/rust-secp256k1-zkp/pull/48

> TLDR, how stable is musig2?

I don't know, maybe @LLFourn does?

@matthewjablack 

> This is a little annoying since it would mean validation for Oracle Announcements would require cfd-dlc-js and couldn't be done in native JS as it's done currently https://github.com/AtomicFinance/node-dlc/blob/master/packages/messaging/lib/messages/OracleAnnouncementV0.ts#L62

Two possible solutions would be use a wasm module generated from either cfd-core or rust-secp256k1-zkp (I think it could work, but honestly haven't tried yet), or if you have access to EC operations in JS implement the key aggregation algorithm which is all you need on client side (not sure it'd be a good idea tbh, and it might take some work).

We can probably discuss at the next meeting whether we want to do that or not. I think it's a bit nicer as it gives more compact and constant size announcements, but if people prefer to have separate signatures we can also go that way.

LLFourn: whatever happens with DLCs going forward I expect MuSig2 will be part of it so I think relying on the MuSig key aggregation function is not too much to ask to simplify things a bit. It should be stable soon.

One thing I will bring up is whether the authors of the spec think that MuSig is appropriate for this kind of thing.

Messaging.md
nkohen: I think the idea is that one way or another people need to get the oracle public key in a good way (e.g. authenticated channel)

Messaging.md
nkohen: My personal opinion upon reading this is that perhaps we want to utilize sub types here and have the default (for now) be having all of them but in the future (specifically after musig2 gets a bip number, after which there will be breaking changes in the hash tags to include the bip number ... I think) I think it would be totally reasonable to introduce a new subtype for bip-??? musig aggregated proof of knowledge, which would likely become the default.

Tibo-lg:
Updated specs and test vectors to use regular Schnorr signatures for proof of knowledge as agreed during spec meeting (using a type prefix to enable using something different in the future).

Messaging.md
```diff
@@ -363,92 +371,210 @@ Witnesses should be sorted by the `input_serial_id` sent in `funding_input` defi
...
+   * [`u32`:`timestamp`]
+   * [`oracle_schemes`:`oracle_schemes`]
+   * [`signature`:`oracle_metadata_signature`]
+
+where the `oracle_metadata_signature` is a Schnorr signature by the `announcement_public_key` using the tag `oraclemetadata/v0` of the sha256 hash of the serialized `oracle_metadata` message omitting the `announcement_public_key` and of course the the `oracle_metadata_signature`.
+
+#### Rationale
+
+Separate public keys are required for announcement and attestation because the attestation scheme
+used in the current DLC spec reduces the security of non-attestation Schnorr signatures issued by the
```
Christewart, line 463: Any citation/reason why it degrades security? It seems we state this as truth without some sort of reason why.

Messaging.md
```diff
@@ -363,92 +371,210 @@ Witnesses should be sorted by the `input_serial_id` sent in `funding_input` defi
...
+
+In order to enable oracles to support multiple attestation schemes, and clients to chose their preferred one, the `oracle_schemes` type can contain multiple attestations schemes.
+Each scheme is defined as a TLV so that clients can easily ignore schemes that they don't know about.
+Note that at the moment the only defined scheme is the `schnorr_scheme` that uses Schnorr signatures.
+
+#### `oracle_schemes`
+
+1. data:
+   * [`bigsize`:`nb_schemes`]
+   * [`nb_schemes*scheme`:`attestation_schemes`]
```
Christewart: Is `attestations_schemes` defined anywhere or is this a typo and meant to be `schnorr_scheme`?

Messaging.md
```diff
@@ -363,92 +371,210 @@ Witnesses should be sorted by the `input_serial_id` sent in `funding_input` defi
...
-   * [`signature`:`annoucement_signature`]
-   * [`x_point`:`oracle_public_key`]
+1. type: 55354 (`oracle_announcement`)
+2. data:
+   * [`signature`:`announcement_signature`]
+   * [`oracle_metadata`:`oracle_metadata`]
    * [`oracle_event`:`oracle_event`]
 
-where `signature` is a Schnorr signature over a sha256 hash of the serialized `oracle_event`, using the tag `announcement/v0`.
+where both the `announcement_signature` is a Schnorr signature over a sha256 hash of the serialized `oracle_event`, using the tag `announcement/v1`.
```
Christewart: what is the public key used to check against the announcement signature?

Tibo-lg: Clarified (it's the `oracle_annoucement_public_key`

Messaging.md
Tibo-lg: Added a `attestation_scheme` type.

Messaging.md
```diff
@@ -363,90 +374,170 @@ Witnesses should be sorted by the `input_serial_id` sent in `funding_input` defi
...
-    * [`signature`:`signature_n`]
-    * [`string`:`outcome_1`]
-    * ...
-    * [`string`:`outcome_n`]
-
+   * [`string`:`event_id`]
+   * [`x_point`:`oracle_attestation_public_key`]
+   * [`bigsize`: `nb_outcomes`]
+   * [`nb_signatures*attested_outcome`:`attested_outcomes`]
+1. subtype: `attested_outcome`
```
Christewart: So I don't understand why this is necessary. Do we use this pattern any where else? Why can't we just serialize the signatures / outcomes with bigsize prefixes?

Issue #161, opened 2021-04-01
Protocol for closing DLC by mutual consent


matthewjablack:
The [introduction](https://github.com/discreetlogcontracts/dlcspecs/blob/master/Introduction.md) specifies that in a DLC, contract collateral is locked in an output that can only be spent by mutual consent.

However, there is currently no protocol for requesting to close a DLC before cet_locktime with mutual consent.

Additionally, there are some concerns with requesting to close a DLC early. Say Alice wishes to exit her DLC early. Alice could simply provide a mutual close signature to Bob. However, if Bob does not respond this could result in a [free option problem](https://github.com/discreetlogcontracts/dlcspecs/issues/155) where Bob can simply wait for favorable market conditions and sign and broadcast when it is optimal for him. 

After chatting with @nkohen it seems the simplest choices to solve this issue are invalidation or revocation (without going into the realm of verifiable encryption).

Invalidation could be solved by having Alice provide her own Input in addition to the DLC Input. If Bob doesn't respond after a certain amount of time, Alice could spend this input, invalidating this close tx.

A `close_dlc` Message could be formatted as following:

1. type ??? (`close_dlc_v0`)
2. data:
    - [`32*byte`:`contract_id`]
    - [`signature`:`close_signature`]
    - [`u64`:`offerPayoutSatoshis`]
    - [`u64`:`acceptPayoutSatoshis`]
    - [`u16`:`num_funding_inputs`]
    - [`num_funding_inputs*funding_input`:`funding_inputs`]

`payout_spk` and `payout_serial_id` from `offer_dlc` as well as `payout_spk` and `payout_serial_id` from `accept_dlc` could be used for constructing the close transaction

`funding_inputs` would be the input(s) provided by Alice that could be invalidated after a timeout. 

Curious to hear if folks think this should be added to the spec. I personally think there are quite a few use cases for closing a DLC (such as requesting to close a position where a market maker is the counterparty).

This might also make more sense to add as an optional [second layer interaction](https://github.com/discreetlogcontracts/dlcspecs/issues/155#issuecomment-810782403) rather than to the core protocol.

benthecarman:
I think a PSBT would serve better than a new tlv

matthewjablack:
PSBT could make more sense, but would still need to be communicated during closing negotiation, meaning a tlv (idk if psbt in a tlv makes sense)

[...]

Issue #132, opened 2021-01-27
How can a DLC implementation bug lead to loss of user funds


Tibo-lg:
I think it would be a good idea to setup a document describing all the possible ways in which a bug in a DLC implementation can lead to loss of user funds, maybe with a grade on criticality of the bug or feasibility of exploiting it. It would then make it easier to implementation maintainers to setup (integration) tests to validate (to the extent possible) against them (the document could even have suggestions on how to do this validation).

As a start we could include (thanks @nkohen): #97 #109

nkohen:
Another thing to include might be #98 as if someone is not parsing the full `prev_tx` themselves then I believe they can be exploited

ariard:
A good start is referring this : https://en.bitcoin.it/wiki/Common_Vulnerabilities_and_Exposures

PR #170, opened 2021-07-20
DLC Mutual Close

matthewjablack:
This PR adds new `close_dlc` as per the discussion regarding https://github.com/discreetlogcontracts/dlcspecs/issues/161 which enables a way to request to close a DLC before cet_locktime with mutual consent

### DIFF
--- a/Protocol.md
+++ b/Protocol.md
@@ -279,6 +279,50 @@ The recipient:
 
   These script pub key forms include only standard forms accepted by the wider set of deployed Bitcoin clients in the network, which increase the chances of successful propagation to miners.
 
+## Second Layer Interaction
+
+Second Layer Interaction consist of second layer "features" which parties can opt-in to using. These messages are not needed to facilitate construction of a DLCs.
+
+### The `close_dlc` Message
+
+This message contains information about mutual close, which allows the counterparty
+to broadcast a mutual closing transaction.
+
+1. type: ? (`close_dlc_v0`)
+2. data:
+   * [`32*byte`:`contract_id`]
+   * [`signature`:`close_signature`]
+   * [`u64`:`offerPayoutSatoshis`]
+   * [`u64`:`acceptPayoutSatoshis`]
+   * [`u64`:`fund_input_serial_id`]
+   * [`u16`:`num_funding_inputs`]
+   * [`num_funding_inputs*funding_input`:`funding_inputs`]
+
+
+`payout_spk` and `payout_serial_id` from `offer_dlc` as well as `payout_spk` and `payout_serial_id` from `accept_dlc` should be used for constructing the close transaction
+
+`fund_input_serial_id` is a randomly chosen number which uniquely identifies the funding input of the initiating party.
+Inputs in the closing transaction will be sorted by `fund_input_serial_id` and `input_serial_id` in `funding_inputs`.
+
+`funding_inputs` are extra inputs to mutual close to avoid free option, and in the future revocation mechanism will be introduced
+
+#### Requirements
+
+The sender MUST:
+
+  - set `contract_id` by exclusive-OR of the `funding_txid`, the `funding_output_index` and the `temporary_contract_id` from the `offer_dlc` and `accept_dlc` messages.
+  - set `close_signature` to the valid signature, using its `funding_pubkey` for the close transaction, as defined in the [transaction specification](Transactions.md#close-transaction).
+
+The recipient:
+
+  - if any input in `funding_inputs` is not a BIP141 (Segregated Witness) input.
+    - MUST reject the contract.
+  - if the `close_signature` is incorrect:
+    - MUST reject the contract.
+  - MUST NOT broadcast the closing transaction before receipt of a valid `close_dlc`.
+  - on receipt of a valid `close_dlc`:
+    - SHOULD broadcast the closing transaction.
+
 # Authors
 
 Nadav Kohen <nadavk25@gmail.com>
--- a/Protocol.md
+++ b/Protocol.md
@@ -279,9 +279,9 @@ The recipient:
 
   These script pub key forms include only standard forms accepted by the wider set of deployed Bitcoin clients in the network, which increase the chances of successful propagation to miners.
 
-## Second Layer Interaction
+## Optional features
 
-Second Layer Interaction consist of second layer "features" which parties can opt-in to using. These messages are not needed to facilitate construction of a DLCs.
+Optional features consist of "features" which parties can opt-in to using. These messages are not needed to facilitate construction of a DLCs.
 
 ### The `close_dlc` Message
 
@@ -301,24 +301,24 @@ to broadcast a mutual closing transaction.
 
 `payout_spk` and `payout_serial_id` from `offer_dlc` as well as `payout_spk` and `payout_serial_id` from `accept_dlc` should be used for constructing the close transaction
 
-`fund_input_serial_id` is a randomly chosen number which uniquely identifies the funding input of the initiating party.
+`fund_input_serial_id` is a randomly chosen number which uniquely identifies the funding output to be spent.
 Inputs in the closing transaction will be sorted by `fund_input_serial_id` and `input_serial_id` in `funding_inputs`.
 
-`funding_inputs` are extra inputs to mutual close to avoid free option, and in the future revocation mechanism will be introduced
+`funding_inputs` are extra inputs to mutual close to enable the sending party to nullify the transaction by double spending the inputs.
 
 #### Requirements
 
 The sender MUST:
 
-  - set `contract_id` by exclusive-OR of the `funding_txid`, the `funding_output_index` and the `temporary_contract_id` from the `offer_dlc` and `accept_dlc` messages.
-  - set `close_signature` to the valid signature, using its `funding_pubkey` for the close transaction, as defined in the [transaction specification](Transactions.md#close-transaction).
+  - set `contract_id` from the `sign_dlc` message.
+  - set `close_signature` to a valid signature, using its `funding_pubkey` for the close transaction, as defined in the [transaction specification](Transactions.md#close-transaction).
 
 The recipient:
 
   - if any input in `funding_inputs` is not a BIP141 (Segregated Witness) input.
-    - MUST reject the contract.
+    - MUST ignore the message.
   - if the `close_signature` is incorrect:
-    - MUST reject the contract.
+    - MUST ignore the message.
   - MUST NOT broadcast the closing transaction before receipt of a valid `close_dlc`.
   - on receipt of a valid `close_dlc`:
     - SHOULD broadcast the closing transaction.
--- a/Protocol.md
+++ b/Protocol.md
@@ -298,6 +298,7 @@ to broadcast a mutual closing transaction.
    * [`u16`:`num_extra_inputs`]
    * [`num_extra_inputs*extra_input`:`extra_inputs`]
    * [`extra_signatures`:`extra_signatures`]
+   * [`u32`:`close_locktime`]
 
 
 `payout_spk` and `payout_serial_id` from `offer_dlc` as well as `payout_spk` and `payout_serial_id` from `accept_dlc` should be used for constructing the close transaction
@@ -307,6 +308,8 @@ Inputs in the closing transaction will be sorted by `fund_input_serial_id` and `
 
 `extra_inputs` are extra inputs to mutual close to enable the sending party to nullify the transaction by double spending the inputs.
 
+`close_locktime` is the nLockTime to be put in close messages.
+
 #### Requirements
 
 The sender MUST:
@@ -314,6 +317,8 @@ The sender MUST:
   - set `contract_id` from the `sign_dlc` message.
   - set `close_signature` to a valid signature, using the private key associated with its `funding_pubkey` for the close transaction, as defined in the [transaction specification](Transactions.md#close-transaction).
   - set `extra_signatures` to contain valid witnesses for every funding input specified by `extra_inputs` and in the same order.
+  - set `close_locktime` to be a UNIX timestamp, or a block height as distinguished [here](https://en.bitcoin.it/wiki/NLockTime).
+  - set `close_locktime` to be less than or equal to current UNIX timestamp or block.
 
 The recipient:
 
--- a/Protocol.md
+++ b/Protocol.md
@@ -316,10 +316,9 @@ The sender MUST:
 The recipient:
 
   - if any input in `funding_inputs` is not a BIP141 (Segregated Witness) input.
-    - MUST ignore the message.
+    - MAY ignore the message.
   - if the `close_signature` is incorrect:
     - MUST ignore the message.
-  - MUST NOT broadcast the closing transaction before receipt of a valid `close_dlc`.
   - on receipt of a valid `close_dlc`:
     - SHOULD broadcast the closing transaction.
 
--- a/Protocol.md
+++ b/Protocol.md
@@ -312,7 +312,7 @@ Inputs in the closing transaction will be sorted by `fund_input_serial_id` and `
 The sender MUST:
 
   - set `contract_id` from the `sign_dlc` message.
-  - set `close_signature` to a valid signature, using its `funding_pubkey` for the close transaction, as defined in the [transaction specification](Transactions.md#close-transaction).
+  - set `close_signature` to a valid signature, using the private key associated with its `funding_pubkey` for the close transaction, as defined in the [transaction specification](Transactions.md#close-transaction).
   - set `funding_signatures` to contain valid witnesses for every funding input specified by `funding_inputs` and in the same order.
 
 The recipient:
--- a/Protocol.md
+++ b/Protocol.md
@@ -9,6 +9,8 @@
           * [The `accept_dlc` Message](#the-accept_dlc-message)
           * [The `sign_dlc` Message](#the-sign_dlc-message)
           * [Script Pubkey Standardness Definition](#script-pubkey-standardness-definition)
+      * [Optional Features](#optional-features)
+          * [The `close_dlc` Message](#the-close_dlc-message)
 * [Authors](#authors)
 
 # Contract
--- a/Protocol.md
+++ b/Protocol.md
@@ -295,17 +295,17 @@ to broadcast a mutual closing transaction.
    * [`u64`:`offerPayoutSatoshis`]
    * [`u64`:`acceptPayoutSatoshis`]
    * [`u64`:`fund_input_serial_id`]
-   * [`u16`:`num_funding_inputs`]
-   * [`num_funding_inputs*funding_input`:`funding_inputs`]
-   * [`funding_signatures`:`funding_signatures`]
+   * [`u16`:`num_extra_inputs`]
+   * [`num_extra_inputs*extra_input`:`extra_inputs`]
+   * [`extra_signatures`:`extra_signatures`]
 
 
 `payout_spk` and `payout_serial_id` from `offer_dlc` as well as `payout_spk` and `payout_serial_id` from `accept_dlc` should be used for constructing the close transaction
 
 `fund_input_serial_id` is a randomly chosen number which uniquely identifies the funding output to be spent.
-Inputs in the closing transaction will be sorted by `fund_input_serial_id` and `input_serial_id` in `funding_inputs`.
+Inputs in the closing transaction will be sorted by `fund_input_serial_id` and `input_serial_id` in `extra_inputs`.
 
-`funding_inputs` are extra inputs to mutual close to enable the sending party to nullify the transaction by double spending the inputs.
+`extra_inputs` are extra inputs to mutual close to enable the sending party to nullify the transaction by double spending the inputs.
 
 #### Requirements
 
@@ -313,13 +313,13 @@ The sender MUST:
 
   - set `contract_id` from the `sign_dlc` message.
   - set `close_signature` to a valid signature, using the private key associated with its `funding_pubkey` for the close transaction, as defined in the [transaction specification](Transactions.md#close-transaction).
-  - set `funding_signatures` to contain valid witnesses for every funding input specified by `funding_inputs` and in the same order.
+  - set `extra_signatures` to contain valid witnesses for every funding input specified by `extra_inputs` and in the same order.
 
 The recipient:
 
-  - if any input in `funding_inputs` is not a BIP141 (Segregated Witness) input.
+  - if any input in `extra_inputs` is not a BIP141 (Segregated Witness) input.
     - MAY ignore the message.
-  - if any `signature` or `witness` in `funding_signatures` is incorrect:
+  - if any `signature` or `witness` in `extra_signatures` is incorrect:
     - MUST ignore the message.
   - if the `close_signature` is incorrect:
     - MUST ignore the message.
--- a/Protocol.md
+++ b/Protocol.md
@@ -326,6 +326,10 @@ The recipient:
 
   - if any input in `extra_inputs` is not a BIP141 (Segregated Witness) input.
     - MAY ignore the message.
+  - if `offer_payout_satoshis` is greater than zero and below the dust limit
+    - MUST ignore the message
+  - if `accept_payout_satoshis` is greater than zero and below the dust limit
+    - MUST ignore the message
   - if any `signature` or `witness` in `extra_signatures` is incorrect:
     - MUST ignore the message.
   - if the `close_signature` is incorrect:
--- a/Protocol.md
+++ b/Protocol.md
@@ -75,8 +75,8 @@ the funding transaction and CETs.
    * [`spk`:`payout_spk`]
    * [`u64`:`payout_serial_id`]
    * [`u64`:`offer_collateral_satoshis`]
-   * [`u16`:`num_funding_inputs`]
-   * [`num_funding_inputs*funding_input`:`funding_inputs`]
+   * [`u16`:`num_inputs`]
+   * [`num_inputs*input`:`inputs`]
    * [`spk`:`change_spk`]
    * [`u64`:`change_serial_id`]
    * [`u64`:`fund_output_serial_id`]
@@ -99,8 +99,8 @@ the funding transaction output. `payout_spk` specifies the script
 pubkey that CETs and the refund transaction should use in the sender's output.
 
 `offer_collateral_satoshis` is the amount the sender is putting into the
-contract. `num_funding_inputs` is the number of funding inputs contributed by
-the sender and `funding_inputs` contains outputs, outpoints, and expected weights
+contract. `num_inputs` is the number of funding inputs contributed by
+the sender and `inputs` contains outputs, outpoints, and expected weights
 of the sender's funding inputs. `change_spk` specifies the script pubkey that funding
 change should be sent to.
 
@@ -162,10 +162,10 @@ The receiving node MUST reject the contract if:
   - `payout_spk` or `change_spk` are not a [standard script pubkey](#script-pubkey-standardness-definition).
   - it considers `feerate_per_vb` too small for timely processing or unreasonably large.
   - `funding_pubkey` is not a valid secp256k1 pubkey in compressed format.
-  - `funding_inputs` do not contribute at least `total_collateral_satoshis` plus full [fee payment](Transactions.md#fee-payment).
+  - `inputs` do not contribute at least `total_collateral_satoshis` plus full [fee payment](Transactions.md#fee-payment).
   - Any `input_serial_id` is duplicated
   - The `fund_output_serial_id` and `change_serial_id` are not set to different value
-  - Any input in `funding_inputs` is not a BIP141 (Segregated Witness) input.
+  - Any input in `inputs` is not a BIP141 (Segregated Witness) input.
 
 ### The `accept_dlc` Message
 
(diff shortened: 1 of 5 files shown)


Protocol.md
```diff
@@ -279,6 +279,50 @@ The recipient:
 
   These script pub key forms include only standard forms accepted by the wider set of deployed Bitcoin clients in the network, which increase the chances of successful propagation to miners.
 
+## Second Layer Interaction
```
Tibo-lg: NIT: I don't think second layer is a good choice as it will become confusing once we have an off-chain version. What about just "optional features"?

Protocol.md
```diff
@@ -279,6 +279,50 @@ The recipient:
...
+   * [`u64`:`offerPayoutSatoshis`]
+   * [`u64`:`acceptPayoutSatoshis`]
+   * [`u64`:`fund_input_serial_id`]
+   * [`u16`:`num_funding_inputs`]
+   * [`num_funding_inputs*funding_input`:`funding_inputs`]
+
+
+`payout_spk` and `payout_serial_id` from `offer_dlc` as well as `payout_spk` and `payout_serial_id` from `accept_dlc` should be used for constructing the close transaction
+
+`fund_input_serial_id` is a randomly chosen number which uniquely identifies the funding input of the initiating party.
```
Tibo-lg: I'm not sure I understand, isn't that for the input coming from the fund transaction?

Protocol.md
```diff
@@ -279,6 +279,50 @@ The recipient:
...
+Inputs in the closing transaction will be sorted by `fund_input_serial_id` and `input_serial_id` in `funding_inputs`.
+
+`funding_inputs` are extra inputs to mutual close to avoid free option, and in the future revocation mechanism will be introduced
+
+#### Requirements
+
+The sender MUST:
+
+  - set `contract_id` by exclusive-OR of the `funding_txid`, the `funding_output_index` and the `temporary_contract_id` from the `offer_dlc` and `accept_dlc` messages.
```
Tibo-lg: I guess you could just say that it should be the same as the one in the sign message?

Protocol.md
```diff
@@ -279,6 +279,50 @@ The recipient:
...
+   * [`u16`:`num_funding_inputs`]
+   * [`num_funding_inputs*funding_input`:`funding_inputs`]
+
+
+`payout_spk` and `payout_serial_id` from `offer_dlc` as well as `payout_spk` and `payout_serial_id` from `accept_dlc` should be used for constructing the close transaction
+
+`fund_input_serial_id` is a randomly chosen number which uniquely identifies the funding input of the initiating party.
+Inputs in the closing transaction will be sorted by `fund_input_serial_id` and `input_serial_id` in `funding_inputs`.
+
+`funding_inputs` are extra inputs to mutual close to avoid free option, and in the future revocation mechanism will be introduced
```
Tibo-lg: nit: "future revocation" => "future a revocation"

Protocol.md
```diff
@@ -279,6 +279,50 @@ The recipient:
...
+
+#### Requirements
+
+The sender MUST:
+
+  - set `contract_id` by exclusive-OR of the `funding_txid`, the `funding_output_index` and the `temporary_contract_id` from the `offer_dlc` and `accept_dlc` messages.
+  - set `close_signature` to the valid signature, using its `funding_pubkey` for the close transaction, as defined in the [transaction specification](Transactions.md#close-transaction).
```
Tibo-lg: nit: "the valid" => "a valid"

Protocol.md
```diff
@@ -279,6 +279,50 @@ The recipient:
...
+
+The recipient:
+
+  - if any input in `funding_inputs` is not a BIP141 (Segregated Witness) input.
```
Tibo-lg: I think it's better anyway but I'm not sure it's a strong requirement in this case.

Protocol.md
```diff
@@ -279,6 +279,50 @@ The recipient:
...
+    - MUST reject the contract.
```
Tibo-lg: nit: maybe "MUST ignore the message"? The contract cannot really be rejected at that point I'd say.

Protocol.md
```diff
@@ -279,6 +279,50 @@ The recipient:
...
+  - if the `close_signature` is incorrect:
+    - MUST reject the contract.
```
Tibo-lg: ditto

Protocol.md
```diff
@@ -279,6 +279,50 @@ The recipient:
...
+  - MUST NOT broadcast the closing transaction before receipt of a valid `close_dlc`.
```
Tibo-lg: Not sure I understand this, how could they broadcast the transaction before?

Protocol.md
```diff
@@ -279,6 +279,50 @@ The recipient:
...
+  - on receipt of a valid `close_dlc`:
+    - SHOULD broadcast the closing transaction.
```
Tibo-lg, line 338: Is it though? What if the counter party doesn't want to close? Or are you considering a setting where there is implicit trust between the parties that they will close when requested? In that case it would be better to specify it, but maybe better to be general and not required that the transaction be broadcast.
That actually make me think that in practice it might be better to have first a "propose close" message to check if the counter party is willing to close or not. But I guess that's maybe to complex for now.

Protocol.md
Tibo-lg: also maybe better to explain what the free option is, i.e. that it enables the sending party to nullify the transaction by double spending the inputs.

Protocol.md
matthewjablack: Good point. I think optional features works better here as well.

Protocol.md
matthewjablack: Yes that was a mistake. It should be `funding output to be spent.` (essentially there should be a serial id for initiator inputs, and dlc to be spent).

Protocol.md
matthewjablack: Good point. Updated

Protocol.md
matthewjablack: Good point, because malleability attacks aren't a concern since there's no transaction to spend the outputs of dlc close. I'll change to `MAY ignore the message`

Protocol.md
matthewjablack: Agreed. Updated

Protocol.md
matthewjablack: ditto

Protocol.md
matthewjablack: Perhaps I misunderstood the steps in `sign_dlc` the conditions for which the acceptor should broadcast the fund transactions

- if any witness exceeds its corresponding max_witness_len from the offer_dlc message:
  - MAY reject the contract.
  - MUST NOT broadcast the funding transaction before receipt of a valid sign_dlc.
  - on receipt of a valid sign_dlc:
    - SHOULD broadcast the funding transaction.
    
I guess this is specifically referring to the case where `witness exceeds its corresponding max_witness_len from the offer_dlc message` in which case they should hold off on broadcasting

I'll remove this line

Protocol.md
matthewjablack: In the particular setting I had in mind, there would probably be a level of trust between the parties i.e. market maker and user. However you're right, this probably doesn't belong in the protocol. Additionally, even if there is trust, there's no way for the initiator to know if `offerPayoutSatoshis` and `acceptPayoutSatoshis` are set to an amount that the counterparty will accept since the market can change quickly. 

I agree that a "propose close" message would be better to check if the other party is open to closing, but maybe out of scope for this PR. On a separate note, there should probably be a "propose dlc offer" message so that a party can check if a counter party is willing to enter into a dlc with them without revealing inputs.

Protocol.md
matthewjablack: Good point. Updated

benthecarman:
imo this would be better if it was more generalized. It would be great if this could be used to inside of any transaction. That way we could close our DLC inside of a regular transaction or in a batch with many other people, or even inside of a coinjoin.

Protocol.md
```diff
@@ -279,6 +279,49 @@ The recipient:
...
+## Optional features
+
+Optional features consist of "features" which parties can opt-in to using. These messages are not needed to facilitate construction of a DLCs.
+
+### The `close_dlc` Message
+
+This message contains information about mutual close, which allows the counterparty
+to broadcast a mutual closing transaction.
+
+1. type: ? (`close_dlc_v0`)
```
ariard, line 293: You can use something in the LN experimental range : 32768 - 65535 and mark it as reserved in the waiting room : https://github.com/lightningnetwork/lightning-rfc/issues/716

Protocol.md
```diff
@@ -279,6 +279,49 @@ The recipient:
...
+   * [`u16`:`num_funding_inputs`]
+   * [`num_funding_inputs*funding_input`:`funding_inputs`]
+
+
+`payout_spk` and `payout_serial_id` from `offer_dlc` as well as `payout_spk` and `payout_serial_id` from `accept_dlc` should be used for constructing the close transaction
+
+`fund_input_serial_id` is a randomly chosen number which uniquely identifies the funding output to be spent.
+Inputs in the closing transaction will be sorted by `fund_input_serial_id` and `input_serial_id` in `funding_inputs`.
+
+`funding_inputs` are extra inputs to mutual close to enable the sending party to nullify the transaction by double spending the inputs.
```
ariard: I'm not sure you can really guarantee the double-spend without further checks on the payouts outputs `witnessScripts`. I think your counterparty might start to spend its payouts, even honestly, as a start of a new DLC or LN, thus falsifying your fee-estimation to successfully double-spend.

One mitigation could be to require to reveal `witnssScripts` and enforce there is a 1 OP_CSV to ensure children can be accepted as final in the network mempools.

ariard (commented):
I agree with Ben, it could be better to uplift this scheme as an application on top of LN splice-in/splice-out. See BOLT specification draft : https://github.com/lightningnetwork/lightning-rfc/pull/863

Protocol.md
Tibo-lg: I guess ideally the best would be to use `42784` to follow after the signing message.

Protocol.md
Tibo-lg: I'm not sure what is your concern, but the point of being able to double spend is in case your counter party doesn't broadcast the closing TX, so I feel if he would be spending the closing TX output it means that it's already broadcast and thus you don't need to worry about double spending?

Protocol.md
Tibo-lg: Yeah it's tricky... I don't have a really good solution for now.

Protocol.md
Christewart: Isn't this implied that all inputs are BIP141 compliant as the we require dlc funding outputs to be BIP141?

Protocol.md
```diff
@@ -279,6 +279,49 @@ The recipient:
...
+   * [`32*byte`:`contract_id`]
+   * [`signature`:`close_signature`]
+   * [`u64`:`offerPayoutSatoshis`]
+   * [`u64`:`acceptPayoutSatoshis`]
+   * [`u64`:`fund_input_serial_id`]
+   * [`u16`:`num_funding_inputs`]
+   * [`num_funding_inputs*funding_input`:`funding_inputs`]
+
+
+`payout_spk` and `payout_serial_id` from `offer_dlc` as well as `payout_spk` and `payout_serial_id` from `accept_dlc` should be used for constructing the close transaction
```
Christewart, line 306: not saying this is a blocker, but is there a scenario where you want to mutually close _because_ your `payout_spk` is compromised. Perhaps out of scope, but I wanted to raise the concern

Protocol.md
Tibo-lg: That's about the inputs added to the closing transactions, not the funding outputs, so unless I'm mistaken I don't think we need specific constraints on these (even though it's probably a good idea).

Protocol.md
Tibo-lg: That's a good point I think. Though on the other hand it feels like it could allow an attacker who doesn't control the `payout_spk` to steal your funds (the attacker would need to have compromised your node I guess?).

Protocol.md
matthewjablack: Agree with @Tibo-lg. Curious to know if there's a case that's not accounted for here. 

Generally the initiator should double-spend the `funding_inputs` in the case that the closing tx is not created in a timely manner (very dependent on the type of contract of course, and market conditions). This should be done shortly after sending the dlc close message, in the case that the counterparty doesn't mutually close the dlc. 

In the case that the counterparty does mutually close the dlc in a timely manner, then nullifying the transaction would no longer be a concern.

Protocol.md
Christewart: So your closing transaction can have unrelated inputs to the in progress DLC that you are trying to close?

Tibo-lg: Yes in order to be able to double spend them to make the closing transaction invalid if your counter party doesn't broadcast it in a timely manner.

matthewjablack: That's correct. The purpose of these unrelated inputs is to enable the initiator of the mutual close to have the ability to nullify the close transaction by spending those inputs in the event that counterparty doesn't immediately act on the mutual close.

Protocol.md
dillionverma: `42784` seems to be already used by `contract_descriptor_v1`

Tibo-lg: Yes, so until #171 or similar is merged it is taken, but after all the inner structures will use context specific type numbers.

matthewjablack:
I realized `funding_signatures` for `funding_inputs` was missing, which are required sigs to broadcast and finalize close dlc. This has been added to `close_dlc` specification in `Protocol.md`

Protocol.md
ariard: Just to be clear, what do you qualify by "extra inputs", the funding outputs of a DLC funding transaction or inputs beyond the scope of the protocol ?

matthewjablack: These extra inputs are beyond the scope of the protocol. Just utxos that exist in a users wallet

ariard: Thanks I see, though I think my point still hold.

Let's say you have a mutual closing transaction between Alice-Bob composed of 2-inputs and 2-outputs.

First input is the outpoint of the funding transaction output (the 2-of-2 P2WSH)
Second input is an input from Alice wallet.
First output is Alice's payout.
Second output is Bob's payout.

The mutual close is broadcasted by Alice with an absolute fee of 1000 sats.
Bob spends his own payout output with a child transaction paying an absolute fee of 1000 sats.
Alice attempts to double-spend the mutual close by spending her wallet input with a transaction paying an absolute fee of 1500 sats.

According BIP125, rule 3, a replacement transactions must pay an absolute fee higher than the replaced transactions. Alice attempts should fail as the fee is lower than the mutual close + Bob's child one ?

I think if we want to guarantee the replacement of the mutual close with an unilateral CET, I think we should ensure that
1) the payout outputs are encumbered by a 1 CSV (to expel any child from the network mempools)
2) the absolute fee, the feerate and replacement fee penalty of any CET transaction is higher than the mutual close one

However, I think fulfilling the second requirement is in contradiction with the goal of this proposal, namely fast-close in period of fees higher than the feerate with which CET have been counter-signed.

Let me know if you think the description is correct.

So I think a better behavior, instead of double-spending the mutual close with wallet inputs could be to CPFP the mutual close ? And if it doesn't work, assumes the CET has been broadcast across network mempools and CPFP the CET ?

Or we can swallow the bullet for now, ship it as it is, mark it as an issue in #132 and wait for package-relay simplifying that kind of issue :) ?

matthewjablack: I think I see the confusion. The initiator has a wallet input, but the reciprocator (the one broadcasting) does not. 

> So I think a better behavior, instead of double-spending the mutual close with wallet inputs could be to CPFP the mutual close ? And if it doesn't work, assumes the CET has been broadcast across network mempools and CPFP the CET ?

Additionally, the purpose of the wallet input is to nullify the mutual close in the case that the mutual close has not been broadcast at all. So it would not be possible to `CPFP the mutual close` if it has not been broadcast in the first place. 

> The mutual close is broadcasted by Alice with an absolute fee of 1000 sats.
> Bob spends his own payout output with a child transaction paying an absolute fee of 1000 sats.
> Alice attempts to double-spend the mutual close by spending her wallet input with a transaction paying an absolute fee of 1500 sats.

In the DLC Close spec proposed, only one party has a wallet input, and that would be the initiator. So if Alice is broadcasting the closing tx, she would be the reciprocator, not the initiator, and she would not have her own wallet input as part of the closing tx. Bob however would have a wallet input. 

So a feasible scenario similar to what you've proposed might look like this (assuming Alice is the initiator and Bob is the reciprocator and Alice is spending her wallet input because Bob did not broadcast mutual close in a timely manner):

First input is the outpoint of the funding transaction output (the 2-of-2 P2WSH)
Second input is an input from Alice wallet.
First output is Alice's payout.
Second output is Bob's payout.

Alice spends her wallet input with a transaction paying an absolute fee of 1500 sats.
Bob attempts to double-spend Alice's wallet input by broadcasting the mutual close with an absolute fee of 1000 sats.
Bob continues the attempt to double spend Alice's wallet input by spending his own payout output with a child transaction paying an absolute fee of 1000 sats.

In this instance, Alice would be spending her wallet input since Bob did not broadcast in a timely manner. 

What's not clear to me is whether BIP 125 rule 3 would make Bob's double-spend using the mutual close tx invalid since the absolute fees would be lower than Alice's original spend of her wallet input.

But regardless of whether Bob's broadcast of the mutual close is successful, Alice has either successfully spent her wallet input, nullifying the mutual close, or has forced Bob's hand to actual broadcast the mutual close in a timely manner, which is why Alice was spending her wallet input in the first place. 

In a real-world scenario, one could imagine a futures contract that Alice and Bob have entered into for a DLC. After some time inside the contract, Alice wishes to exit the position with Bob, and waits only 1 minute after passing a DlcClose message to him. After 1 minute, Alice does not detect the mutual close tx in the mempool, so Alice spends her wallet input and either she succeeds in nullifying the mutual close tx, or Bob successfully double-spends her wallet input with the mutual close tx which achieves Alice's goal in the first place.

ariard: Thanks, now I think we have the same model in mind.

> What's not clear to me is whether BIP 125 rule 3 would make Bob's double-spend using the mutual close tx invalid since the absolute fees would be lower than Alice's original spend of her wallet input.

Currently yes, BIP 125 rule 3 would prevent Bob's double spend Alice input. Though if package relay is deployed Bob would be able to broadcast a mutual_close + CPFP and if mempool fee spikes at the same time, blocks Alice wallet input spending. Of course, Alice could override assuming she is able to observe Bob mutual_close + CPFP broadcast (an attacker could prevent this by broadcasting low-fees M1 inside Alice's mempool and pinning M2 across the rest of the network).

I think that case is good to know and to monitor when package relay is deployed.

> But regardless of whether Bob's broadcast of the mutual close is successful, Alice has either successfully spent her wallet input, nullifying the mutual close, or has forced Bob's hand to actual broadcast the mutual close in a timely manner, which is why Alice was spending her wallet input in the first place.

I think that's correct and it would be a good rational to add to the PR. Further, I think it could be good to specify an implementation-defined `nullifying_point` for Alice to enforce, similar in spirit to the `security_point` from `Non-Interactive-Protocol.md`. Eg, value could be dynamic in function of mempool fees changes nullifying the economic equilibrium of the DLC.

> In a real-world scenario, one could imagine a futures contract that Alice and Bob have entered into for a DLC. After some time inside the contract, Alice wishes to exit the position with Bob, and waits only 1 minute after passing a DlcClose message to him. After 1 minute, Alice does not detect the mutual close tx in the mempool, so Alice spends her wallet input and either she succeeds in nullifying the mutual close tx, or Bob successfully double-spends her wallet input with the mutual close tx which achieves Alice's goal in the first place.

Really agree on the long-term design that we should have a bunch of options, even negotiated with fee premiums to update or exit the DLC in function of other changes, such as mempool fee spikes or lightning routing fees ones :)

matthewjablack: Glad we're on the same page :) 

> Currently yes, BIP 125 rule 3 would prevent Bob's double spend Alice input. Though if package relay is deployed Bob would be able to broadcast a mutual_close + CPFP and if mempool fee spikes at the same time, blocks Alice wallet input spending. Of course, Alice could override assuming she is able to observe Bob mutual_close + CPFP broadcast (an attacker could prevent this by broadcasting low-fees M1 inside Alice's mempool and pinning M2 across the rest of the network).

Got it, so essentially as it stands, BIP 125 would favor Alice, and package relay would favor Bob. Either way, she achieves her goal.

> I think that's correct and it would be a good rational to add to the PR. Further, I think it could be good to specify an implementation-defined nullifying_point for Alice to enforce, similar in spirit to the security_point from Non-Interactive-Protocol.md. Eg, value could be dynamic in function of mempool fees changes nullifying the economic equilibrium of the DLC.

Great point, I've added `nullifying_point` and rationale to [`Non-Interactive-Protocol.md` ](https://github.com/discreetlogcontracts/dlcspecs/pull/170/files#diff-fb2caf21cc01d287e559f7cbfa10bae18416bc18449507e4849720f584e2655dR260) explaining the reasoning for spending funding-inputs. 

> Really agree on the long-term design that we should have a bunch of options, even negotiated with fee premiums to update or exit the DLC in function of other changes, such as mempool fee spikes or lightning routing fees ones :)

💯  I'm sure there's lots more that can be done and improved with future iterations of closing messages!

matthewjablack:
Based on feedback from @ariard I've added `nullifying_point` and rationale to `Non-Interactive-Protocol.md` to explain why funding-inputs are necessary, and introduce a point where the initiator should nullify the close signature by spending funding-inputs as the outcome of the close transaction become increasingly unfavorable.

Protocol.md
```diff
@@ -279,6 +279,53 @@ The recipient:
...
+   * [`u64`:`acceptPayoutSatoshis`]
+   * [`u64`:`fund_input_serial_id`]
+   * [`u16`:`num_funding_inputs`]
+   * [`num_funding_inputs*funding_input`:`funding_inputs`]
+   * [`funding_signatures`:`funding_signatures`]
+
+
+`payout_spk` and `payout_serial_id` from `offer_dlc` as well as `payout_spk` and `payout_serial_id` from `accept_dlc` should be used for constructing the close transaction
+
+`fund_input_serial_id` is a randomly chosen number which uniquely identifies the funding output to be spent.
```
benthecarman, line 308: calling this `fund_input_serial_id` is a bit confusing when we call the other inputs `funding_inputs` however I can't think of a better name

Protocol.md
```diff
@@ -279,6 +279,53 @@ The recipient:
...
+Inputs in the closing transaction will be sorted by `fund_input_serial_id` and `input_serial_id` in `funding_inputs`.
+
+`funding_inputs` are extra inputs to mutual close to enable the sending party to nullify the transaction by double spending the inputs.
+
+#### Requirements
+
+The sender MUST:
+
+  - set `contract_id` from the `sign_dlc` message.
+  - set `close_signature` to a valid signature, using its `funding_pubkey` for the close transaction, as defined in the [transaction specification](Transactions.md#close-transaction).
```
benthecarman: > using its `funding_pubkey`

maybe better to do "using the private associated with its `funding_pubkey`

Protocol.md
```diff
@@ -279,6 +279,53 @@ The recipient:
...
+The recipient:
+
+  - if any input in `funding_inputs` is not a BIP141 (Segregated Witness) input.
+    - MAY ignore the message.
+  - if any `signature` or `witness` in `funding_signatures` is incorrect:
+    - MUST ignore the message.
+  - if the `close_signature` is incorrect:
+    - MUST ignore the message.
+  - on receipt of a valid `close_dlc`:
+    - SHOULD broadcast the closing transaction.
```
benthecarman, line 338: they need to at least sign the tx too

Transactions.md
```diff
@@ -99,6 +99,10 @@ This output sends funds won by the accepter corresponding to this CET's outcome
 
 The refund transaction is exactly the same as a [Contract Execution Transaction](#contract-execution-transaction) except that its locktime is `refund_locktime` (as negotiated in the offer message) instead of `cet_locktime` and the output values for the offerer and the accepter are their respective total collateral values from their offer/accept messages.
 
+# Close Transaction
+
+The close transaction is exactly the same as a [Contract Execution Transaction](#contract-execution-transaction) except that its locktime is `0` and it has additional funding input(s) that are used to avoid the free option problem.
```
benthecarman: > The close transaction is exactly the same as a Contract Execution Transaction

I don't think this applies, there is additional inputs along with their serial ids. This should probably be explained for clarity.

Protocol.md
matthewjablack: in [Transactions.md](https://github.com/discreetlogcontracts/dlcspecs/blob/master/Transactions.md#funding-output) we call the DLC output the **Funding Output** but not sure if `funding_output_serial_id` would be appropriate, since it's an input in this case

Protocol.md
Tibo-lg: Gonna resolve this. My thinking now is that a "propose close" of "propose dlc offer" are things that should happen outside of the core protocol (at least for now).

Protocol.md
matthewjablack: in [Protocol.md](https://github.com/discreetlogcontracts/dlcspecs/blob/master/Protocol.md#the-sign_dlc-message) it only mentions `SHOULD broadcast the funding transaction` not actually signing it. Is signing mentioned somewhere else for this one?

Protocol.md
benthecarman: yeah :/ I'm not sure what a better name is

Protocol.md
matthewjablack: good idea, added

Transactions.md
matthewjablack: changed the language to specify that it's similar but not exactly the same

Protocol.md
Tibo-lg: I think what should be changed is the `funding_inputs` naming. It conflict with the inputs of the funding transactions and is IMHO quite confusing. What about something like `extra_inputs` or `cancellable_inputs`?

Tibo-lg (commented):
It might be nice to consider defining a feature flag for this as well.

matthewjablack:
> It might be nice to consider defining a feature flag for this as well.

I agree that a contract flag could make sense to ensure one party does not waste sats sending a DlcClose message to a peer that is not going to accept it. Perhaps adding contract flag definition would make sense in a subsequent PR. 

> I think what should be changed is the funding_inputs naming. It conflict with the inputs of the funding transactions and is IMHO quite confusing. What about something like extra_inputs or cancellable_inputs?

I changed `funding_inputs` => `extra_inputs` and `funding_signatures` => `extra_signatures` to reduce confusion. I considered `cancellable_inputs` and `cancellable_signatures` however the concept of `cancellable_signatures` seemed very strange and confusing at face value.

matthewjablack:
I also added `close_locktime` to help prevent fee sniping. It should be set to less than or equal to current timestamp/block.

Protocol.md
```diff
@@ -279,6 +281,58 @@ The recipient:
...
+1. type: ? (`close_dlc_v0`)
+2. data:
+   * [`32*byte`:`contract_id`]
+   * [`signature`:`close_signature`]
+   * [`u64`:`offerPayoutSatoshis`]
+   * [`u64`:`acceptPayoutSatoshis`]
+   * [`u64`:`fund_input_serial_id`]
+   * [`u16`:`num_extra_inputs`]
+   * [`num_extra_inputs*extra_input`:`extra_inputs`]
+   * [`extra_signatures`:`extra_signatures`]
```
Tibo-lg: This is not a defined type. I guess you wanted to do the same as for `funding_signatures`, but note that `funding_signatures` is actually defined in `Messaging.md`. Maybe the best would be to change the type name of `funding_signatures` so that it's more abstract and reuse that (if the goal is to have the exact same structure). I guess originally there was only a single use for it which is why it was named like that.

Tibo-lg (commented):
I checked on how this will be affected by #163, and it seems the only things to do is change the field `num_extra_inputs` from u16 to bigsize.

Protocol.md
matthewjablack: This is a great point. I agree that `funding_signatures` should be abstracted. I think `funding_inputs` should also be abstracted since it's used with `DlcClose` as well. 

Renamed `funding_input` to `input` and `funding_signatures` to `input_signatures` to abstract them from funding transaction and allow them to be re-used for `DlcClose` messages.

matthewjablack:
> I checked on how this will be affected by #163, and it seems the only things to do is change the field num_extra_inputs from u16 to bigsize.

Fantastic 👌 

TODO: 
- change `num_extra_inputs` from `u16` to `bigsize` after #163 is merged into master
- set the type number for close message

Protocol.md
Tibo-lg: Thanks! I might have gone with `tx_input` instead of just `input` but we can change it later if the meaning is not clear enough.

Protocol.md
benthecarman: what number was ended up being chosen here?

Protocol.md
```diff
@@ -279,6 +281,58 @@ The recipient:
...
+  - set `contract_id` from the `sign_dlc` message.
+  - set `close_signature` to a valid signature, using the private key associated with its `funding_pubkey` for the close transaction, as defined in the [transaction specification](Transactions.md#close-transaction).
+  - set `extra_signatures` to contain valid witnesses for every funding input specified by `extra_inputs` and in the same order.
+  - set `close_locktime` to be a UNIX timestamp, or a block height as distinguished [here](https://en.bitcoin.it/wiki/NLockTime).
+  - set `close_locktime` to be less than or equal to current UNIX timestamp or block.
+
+The recipient:
+
+  - if any input in `extra_inputs` is not a BIP141 (Segregated Witness) input.
+    - MAY ignore the message.
```
benthecarman, line 328: Should require the payouts to be above the dust limit

Protocol.md
Tibo-lg: At the end I think we decided not to touch message type numbers in #163 so I guess whatever is ok?

matthewjablack: good point, added to `TODO` section at end of PR

Protocol.md
matthewjablack: Good catch, added in [a9ce72b](https://github.com/discreetlogcontracts/dlcspecs/pull/170/commits/a9ce72b82514ef34c6333bcd93627f07cddfba23)

PR #173, opened 2021-09-09
Wednesday September 15 2021 Meeting

nkohen:

### DIFF
--- a/meeting/meeting_september_15_2021.md
+++ b/meeting/meeting_september_15_2021.md
@@ -1,0 +1,70 @@
+# September 15th (7 PM CST)/16th (9 AM JST) Meeting 2021
+
+## Individual Updates (Sync)
+
+* **#Query(Nadav)**
+* **#Query(Tibo)**
+* **#Query(Lloyd)**
+* **#Query(Chris)**
+* **#Query(Jesse)**
+* **#Query(Matt)**
+* **#Query(Ben)**
+* **#Query(Antoine)**
+* **#Query(all)**
+
+## Implementation Updates (Sync)
+
+* bitcoin-s
+* rust-dlc
+* rust-lightning
+* atomic finance
+* Binary outcome DLC implementation from Lloyd
+* others?
+
+## Specification Writing
+
+**#Status_Update_Interrupt**
+
+* If anyone is interested it would be nice to document loss-of-fund vectors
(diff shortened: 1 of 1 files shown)


meeting/meeting_september_15_2021.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Implementation Updates (Sync)
+
+* bitcoin-s
+* rust-dlc
+* atomic finance
+* Binary outcome DLC implementation from Lloyd
+* others?
+
+## Specification Writing
```
Tibo-lg, line 44: Would be nice to discuss the lexicographical ordering thing

meeting/meeting_september_15_2021.md
```diff
@@ -0,0 +1,64 @@
...
+**#Status_Update_Interrupt**
+
+* If anyone is interested it would be nice to document loss-of-fund vectors
+  * https://github.com/discreetlogcontracts/dlcspecs/issues/132
+  * Being robust against random failures?
+    * Nothing in specification about what should be backed up when before sending messages
+    * In LN they also have the channel-reestablish message
+* Draft Updating P2P message serialization to include sub-types, optional fields, tlv streams
+  * https://github.com/discreetlogcontracts/dlcspecs/pull/171
+  * Closer to TLVs than the previous draft
```
Tibo-lg, line 55: Actually I'd like to discuss which one of the two drafts we should go for. After implementing something a bit close to the TLV version one, I'm not really sure it's the best approach.

meeting/meeting_september_15_2021.md
```diff
@@ -0,0 +1,64 @@
...
+* **#Query(Antoine)**
+* **#Query(all)**
+
+## Implementation Updates (Sync)
+
+* bitcoin-s
+* rust-dlc
+* atomic finance
+* Binary outcome DLC implementation from Lloyd
+* others?
```
ariard, line 42: dlc support in rust-lightning? Tibo has done the first minimal step :)

LLFourn:
One thing I thing I'd discuss is the general format of the oracle public key list kind of data structure and perhaps signing it with the main "announcement" key. Or whether we even need such a thing. Perhaps the attestation keys could be in the announcement for example and there is only the announcement key up front.

PR #182, opened 2021-12-23
Simpler(?) version of numerical decomposition

Tibo-lg:
This PR proposes an update to the numerical decomposition document, to try to make things a bit easier to understand. I think "easy to understand" is a subjective notion though and so if people don't think this version is simpler I'll happily close this PR.

One thing to note is that unless I'm mistaken, the stated `O(B*log_B(L))` seems incorrect to me as the worst case is above it.

Also if people think rust is not easy to read I could change it to typescript. I hope the code comments are enough to understand the code though.

### DIFF
--- a/NumericOutcomeCompression.md
+++ b/NumericOutcomeCompression.md
@@ -101,261 +101,265 @@ these integers (interpreted in base `B`) where all digits not used may be any va
 Before generalizing or specifying the algorithm, let us run through a concrete example.
 
 We will consider the interval `[135677, 138621]`, in base `10`.
-Note that the `start` and `end` both begin with the prefix `13` which must be included in every digit prefix, for this purpose we will omit these digits for
-the remainder of this example as we can simply examine the interval `[5677, 8621]` and prepend a `13` to all results to get a result for our original interval.
-
-To cover all cases while looking at as few digits as possible in this interval we need only consider
-`5677`, `8621` individually in addition to the following cases:
+The expected output of the algorithm is:
 
 ```
-5678, 5679,
-568_, 569_,
-57__, 58__, 59__,
-
-6_, 7_,
-
-80__, 81__, 82__, 83__, 84__, 85__,
-860_, 861_,
-8620
+135677, 135678, 135679,
+13568_, 13569_,
+1357__, 1358__, 1359__,
+136_, 137_,
+1380__, 1381__, 1382__, 1383__, 1384__, 1385__,
+13860_, 13861_,
+138620, 138621
 ```
 
 where `_` refers to an ignored digit (an omission from the array of integers representing the digit prefix).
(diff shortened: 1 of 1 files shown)


NumericOutcomeCompression.md
```diff
@@ -101,261 +101,265 @@ these integers (interpreted in base `B`) where all digits not used may be any va
...
-But this method misses out on a good amount of compression as re-introducing the endpoints allows us to replace the first two rows with
-a single `22__` and the last 3 rows with just `4___`.
+```rust
+fn decompose_value(mut value: usize, base: usize, nb_digits: usize) -> Vec<usize> {
+    let mut res = Vec::new();
 
-This optimization is called the **endpoint optimization**.
+    while value > 0 {
+        res.push(value % base);
+        value = ((value as f64) / (base as f64)).floor() as usize;
```
benthecarman, line 163: maybe should just make `value` and `base` `f64`s insted of casting them since it is just an example

NumericOutcomeCompression.md
```diff
@@ -101,261 +101,265 @@ these integers (interpreted in base `B`) where all digits not used may be any va
...
+                list.push(stack.clone());
+                stack[i] += 1;
+            }
+            assert!(stack.iter().all(|x| x < &base));
         }
-    } else if (prefixDigits.length == numDigits - 1) { // Special Case: Front Grouping = Back Grouping
-        startDigits.last.to(endDigits.last).toVector.map { lastDigit =>
-            prefixDigits :+ lastDigit
+
+        // All the single digits in ]start[0]; end[0][ are sufficient to cover their respective intervals.
```
benthecarman, line 326: > `]start[0]; end[0][`

I think this is meant to be `[start[0]; end[0]]`

benthecarman (commented):
Just some nits

but I think this is easier to understand and I think rust is more approachable for most ppl and it seems a lot of the bitcoin space is moving towards rust

NumericOutcomeCompression.md
Tibo-lg: I think `] start[0]; end[0] [` is correct. If it were  `[start[0]; end[0]]` then in the example it would include `1` and `4` which are covered in the loops above and under. Let me know if maybe something is unclear (or if maybe I'm just missing something).

NumericOutcomeCompression.md
Tibo-lg: Mmmmm I'm a bit hesitant because I feel it's quite important to return integers and not floats and I'd be worried that someone just copy pasting that thinks that they can just manipulate floats all over and be fine (tbh I'm not sure if it's actually fine or not, but I think it's just better to avoid it).

Tibo-lg:
> but I think this is easier to understand

Happy to hear :)

PR #187, opened 2022-01-20
Add validity requirement to contract descriptor

Tibo-lg:
Specify that a contract descriptor needs to cover all possible outcomes that can be attested by the oracle(s).

### DIFF
--- a/Messaging.md
+++ b/Messaging.md
@@ -139,6 +139,9 @@ This type contains information about a contract's outcomes and their correspondi
 To save space, only the offerer's payouts are included in this message as the accepter's can be derived using
 `accept_payout = total_collateral - offer_payout`.
 
+**Validity requirement**
+For a contract descriptor to be valid, it is necessary that *a single* payout is defined for any possible outcome that can be attested by the oracle(s).
+
 #### Version 0 `contract_descriptor`
 
 1. type: 42768 (`contract_descriptor_v0`)
--- a/Protocol.md
+++ b/Protocol.md
@@ -130,6 +130,7 @@ The sending node MUST:
   - set `cet_locktime` to be less than `refund_locktime`.
   - use a unique `input_serial_id` for each input
   - set `change_serial_id` and `fund_output_serial_id` to different values
+  - use valid [contract descriptor(s)](./Messaging.md#The-contract_descriptor-Type) within `contract_info`.
 
 The sending node SHOULD:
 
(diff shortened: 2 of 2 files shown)


Tibo-lg:
Rephrased to emphasize that there should be a single payout per possible outcome, as having multiple payout per outcome should I think be prohibited.

benthecarman (commented):
Probably worth adding a note under the `offer_dlc` requirements

https://github.com/discreetlogcontracts/dlcspecs/blob/master/Protocol.md#requirements

benthecarman:
> Rephrased to emphasize that there should be a single payout per possible outcome, as having multiple payout per outcome should I think be prohibited.

Disjoint union contracts can have multiple payout per outcome, we may need an exception for that. Or describe it as per descriptor

Tibo-lg:
> Probably worth adding a note under the `offer_dlc` requirements
> 
> https://github.com/discreetlogcontracts/dlcspecs/blob/master/Protocol.md#requirements

Good point! updated

Tibo-lg:
> > Rephrased to emphasize that there should be a single payout per possible outcome, as having multiple payout per outcome should I think be prohibited.
> 
> Disjoint union contracts can have multiple payout per outcome, we may need an exception for that. Or describe it as per descriptor

IIRC disjoint union conctracts use multiple contract descriptors, each with their own set of payouts and outcomes. So since the validity is defined on the `contract_descriptor` type I think it works (since a contract descriptor should still have only a single payout per outcome even with disjoint union).
It might change if we start supporting "and" contracts, but I'd say we should then update this when we do?

Christewart (approved):
LGTM

benthecarman (approved):

matthewjablack (approved):
LGTM 🔥 

Thanks for finding this @Tibo-lg. We're in the middle of adding a bunch of new validation for node-dlc and currently missing this

Christewart:
Merging as 3 people have approved

Christewart:
How we've implemented it: https://github.com/bitcoin-s/bitcoin-s/blob/ac3bae403bb34152c1372211847026bfad67da1c/core/src/main/scala/org/bitcoins/core/protocol/dlc/models/ContractOraclePair.scala#L14

Test cases: 
https://github.com/bitcoin-s/bitcoin-s/blob/8857af2b66e8d493814d0bb7b817d8d46579709d/core-test/src/test/scala/org/bitcoins/core/protocol/dlc/ContractOraclePairTest.scala#L14

Tibo-lg:
> How we've implemented it: https://github.com/bitcoin-s/bitcoin-s/blob/ac3bae403bb34152c1372211847026bfad67da1c/core/src/main/scala/org/bitcoins/core/protocol/dlc/models/ContractOraclePair.scala#L14
> 
> Test cases: https://github.com/bitcoin-s/bitcoin-s/blob/8857af2b66e8d493814d0bb7b817d8d46579709d/core-test/src/test/scala/org/bitcoins/core/protocol/dlc/ContractOraclePairTest.scala#L14

My implementation is here for reference: https://github.com/p2pderivatives/rust-dlc/pull/40/files

PR #191, opened 2022-04-12
Tuesday April 12 2022 Meeting

nkohen:

### DIFF
--- a/meeting/meeting_april_12_2022.md
+++ b/meeting/meeting_april_12_2022.md
@@ -1,0 +1,44 @@
+# April 12th (6 PM CST)/9th (9 AM JST) Meeting 2022
+
+## Individual Updates (Sync)
+
+* **#Query(Nadav)**
+* **#Query(Chris)**
+* **#Query(Matt)**
+* **#Query(Tibo)**
+* **#Query(Ben)**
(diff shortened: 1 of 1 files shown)


meeting/meeting_april_12_2022.md
```diff
@@ -0,0 +1,44 @@
...
+  * Anything new?
+* Conjunctive Contract Interest
+  * https://mailmanlists.org/pipermail/dlc-dev/2022-January/000098.html
+  * If anyone wanted to attempt writing a spec, feel free
+* P2P discussion (Transport Layer Specs)
+  * https://mailmanlists.org/pipermail/dlc-dev/2022-March/000135.html
+  * **#Discussion**
+
+## DLC Research
+
```
ariard, line 83: Channel jamming research : if DLC start to be routed across the LN, they're going to constitute spontaneous channel jamming, as the liquidity along the path should be locked until the contract maturity. Further, the settlement might be conditional instead of binary like HTLC, i,e not guaranteeing the payment of routing fees to node operators. Likely, routed DLCs would have to pay routing fees based on the length of their `cltv_expiry_delta`s. Happy to collect though on that subject :)

PR #192, opened 2022-04-21
Message segmentation

Tibo-lg:
The noise protocol has a hard limit on message length, which means that messages that exceed that limit need to be split before being sent over it. This PR specifies a message segmentation protocol to achieve this.

Implementation: https://github.com/p2pderivatives/rust-dlc/pull/44


Protocol.md
```diff
@@ -296,6 +296,76 @@ The recipient:
...
+
+#### Rationale
+
+A peer should not use the segmentation protocol if the message can fit withing a single segment.
+
+Given the size constraint of 65535 bytes per segment, a single segment can fit at maximum 404 ecdsa adaptor signatures (a bit less considering the serialization overhead).
+Allowing at maximum 1000 segments for a single message limits the maximum number of ECDSA adaptor signatures that can be exchanged to roughly 404000, which is already a very large number considering the significant cost of signing and verifying them. 
+
+The first segment can fit 65526 (65535 - 2 for wire message type - 2 for `nb_segments` - 5 for `data_len`) bytes for the data field, and all of these should be used.
+While it might be possible to get read of the `data_len` field in this case the increased code complexity is not worth the 0.007% saving.
```
cdecker:
```suggestion
While it might be possible to get rid of the `data_len` field in this case the increased code complexity is not worth the 0.007% saving. 
```

cdecker: An alternative would be to use the `segment_start` message solely for signalling the start, without any data itself. That however means we now have the message overhead without the payload which may be worse. The upside is that you'd only extract payload from a single message type instead of having two paths.

Protocol.md
```diff
@@ -296,6 +296,76 @@ The recipient:
...
+1. data:
+  * [`u16`:`nb_segments`]
+  * [`bigsize`:`data_len`]
+  * [`data_len*u8`:`data`]
+
+### Requirements
+
+The sender MUST:
+  - set `nb_segments` to the number of segments into which the message will be split into
+  - set `nb_segments` to a number greater than 1 and smaller than 1000
```
cdecker, line 323: Since you have a fixed payload size (`data_len`) you could also send the overall payload size (reassembled buffer length from all segments) and the number of segments could be inferred from that. Knowing the size ahead of time precisely could allow the receiver to allocate a buffer that exactly fits, instead of having to either overallocate or expand for each segment.

cdecker: Then again, having the `data_len` in there I expect you might want to allow for other `data_len` sizes in future, making the number of segments unknown, but then we can use a greedy strategy of just consuming segments while the buffer hasn't been filled.

Protocol.md
```diff
@@ -296,6 +296,76 @@ The recipient:
...
+To circumvent this issue, a simple message segmentation protocol is defined, that splits large messages into a `segment_start` message followed by 
+
+## The `segment_start` message
+
+The segment start message indicates the start of a message that is split in several segments.
+
+1. type: 42900 (`segment_start`)
+1. data:
+  * [`u16`:`nb_segments`]
+  * [`bigsize`:`data_len`]
```
cdecker: `data_len` is guaranteed to be a `u16`, so don't use `bigsize`, even if we had <253 bytes which would fit in 1 u8, it's a corner case and can only happen for the last segment anyway. For all but the last segment we're guaranteed to require 3 bytes for bigsize.

Protocol.md
```diff
@@ -296,6 +296,76 @@ The recipient:
...
+  * [`data_len*u8`:`data`]
+
+### Requirements
+
+The sender MUST:
+  - set `data_len` to 65528 if the segment is not the last one.
+
+The receiver:
+  - if the segment is not the last expected one and `data_len` is not set to 65528:
+    - MUST reject the contract
```
cdecker, line 358: I wonder if rejecting the message wouldn't make this more flexible, and reusable for other protocols, though handing this error up to the app layer (contract in this case) might be required too.

Protocol.md
```diff
@@ -296,6 +296,76 @@ The recipient:
...
+  - if a `segment_start` was not received prior to receiving the `segment_chunk` message:
+    - MUST reject the contract
+  - if receiving a message different than `segment_chunk` while expecting one:
+    - MAY reject the contract
```
cdecker, line 362: What happens if I send `>nb_segments` segments?

cdecker: Uhm, sounds like we got a head-of-line blocking here: queue a large message, notice that something urgent needs to be transmitted (keep-alive ping?) and we're unable to send that urgent thing because we're trickling out the large message? Remember that the max size for a segmented message is `1000 * 65528 B = 65'528'000 B =~ 65 MB`, so that can keep slow pipes unusable for some time.

cdecker: Receiving a new `segment_start` while still receiving a `segment_chunk` on the other hand definitely is indicative of an issue at the sender or recipient when using reliable transport.

cdecker (commented):
Just a couple of nits and alternative ideas that could be considered.

Protocol.md
Tibo-lg: I see the upside, but I'm not sure it's worth the loss of space, I'll think about it.

Protocol.md
Tibo-lg: You made two very good points, so if IIUC you're saying either:
* have a fixed number of segments, using a single `data_len` and `nb_segments` in the start message
* have a variable number of segments, using a `data_len` per message (I guess it still requires a `start` though)

I can't think of a case where we would need actual streaming of data, but leaving the possibility open is tempting, tough choice...

Protocol.md
Tibo-lg: True, updated to u16

Protocol.md
Tibo-lg: Could you expand on what you mean here? To me this is feels like a case where you're communicating with either a buggy or malicious peer in which case you'd probably want to stop communication.

Protocol.md
Tibo-lg: The difficulty here is how much the segmentation protocol should be abstracted away from the underlying layer. At first I had specified that the protocol should be reset when the connection is dropped, but it puts quite some constraints for the implementation (at least it was not easy to do) which is why I ended up with this. Not sure what's the best tbh.  

> What happens if I send > nb_segments segments?

Currently that'd be an error, but not a critical one (my thinking was it's possible that a node crashes, reboots and resends a message because it didn't save its state)

 > Uhm, sounds like we got a head-of-line blocking here...

Good point! I guess it depends where we think this segmentation protocol lives. When I wrote this document my thinking was that it was at the DLC application level, so things like pings and such wouldn't reach the protocol logic, but if we want it to be more generic there would probably need to be some hierarchy between the messages (I think receiving any other application level message when expecting a `segment_chunk` should be an error).

Tibo-lg:
@cdecker thanks a lot for the comments!

Roasbeef:
FWIW, `lnd` implemented such segmenting from the very start (this ended up being removed from the spec to have a simpler initial version). AFAICT, based on my initial implementation, you don't actually need to make the segments explicit. Instead the writer just needs to chunk on the wire, w/ the reader continuing to read out logical transport packets until they've reconstructed the entire message. In other words, the application layer knows how many bytes it wants to write/read and can just chunk them on the transport layer then reconstruct on the other side. 

As an example, lets say we're writing out an entire block. The sender serialized in normal Bitcoin wire format, and then sends them out into 65KB chunks. The receiver reads the initial message header (the type), sees that it's a block, then uses a normal stream decoding that'll parse all the Bitcoin p2p varints to continue reading the entire block out. 

Here's a commit w/ some code that demonstrates how this would work in practice: https://github.com/lightningnetwork/lnd/commit/767c550d65ef97a765eabe09c97941d91e05f054. See the `TestWriteMessageChunking` test at the bottom that writes out 195 KB in a chunked message, with the receiver reading out the entire thing. The main idea is that the application layer operates on a stream level (read out N bytes), while the transport operates on a message level (read the next message). 

A bit of the interaction on the reader side in that commit is hidden behind these lines:
```go
	// Attempt to read the entirety of the message generated above.
	buf := make([]byte, len(largeMessage))
	if _, err := io.ReadFull(remoteConn, buf); err != nil {
		t.Fatalf("unable to read message")
	}
```

Which ends up calling into this routine (read in a loop until you read out enough bytes): https://cs.opensource.google/go/go/+/refs/tags/go1.18.3:src/io/io.go;drc=17083a2fdf4475c3f11a3e6a0ef8cb595a5fc4d6;l=326. An internal buffer is then always written to each time we read a transport message off the wire. The reader then gets a stream-like API into that buffer: https://github.com/lightningnetwork/lnd/commit/767c550d65ef97a765eabe09c97941d91e05f054#diff-68fea4a34829ad17d09f775b8196da5e09285ad9fcd8ee9ce74704dd2a1c0d61R98.

Tibo-lg:
@Roasbeef Thanks a lot for the comment and pointers to your code!

> AFAICT, based on my initial implementation, you don't actually need to make the segments explicit. Instead the writer just needs to chunk on the wire, w/ the reader continuing to read out logical transport packets until they've reconstructed the entire message. In other words, the application layer knows how many bytes it wants to write/read and can just chunk them on the transport layer then reconstruct on the other side.

One of the motivation I had for having the segment explicit was to be able to handle disconnects/reconnects without requiring access to the low level network layer. Granted it might not be the best reason, but another reason I can think of was [mentioned by @cdecker](https://github.com/discreetlogcontracts/dlcspecs/pull/192#discussion_r894383344), the ability to send an urgent message in the middle of sending a large one (I cannot think how this could be achieved without headers, but I might just lack imagination).

That being said, I don't have a strong opinion on what is the best, and if there is a need for something similar in LN it'd be really nice that we use the same thing (less thing for our small community to maintain, and we already rely on quite some of the LN messages anyway so I think it'd make sense).

Christewart:
> FWIW, `lnd` implemented such segmenting from the very start (this ended up being removed from the spec to have a simpler initial version). AFAICT, based on my initial implementation, you don't actually need to make the segments explicit. Instead the writer just needs to chunk on the wire, w/ the reader continuing to read out logical transport packets until they've reconstructed the entire message. In other words, the application layer knows how many bytes it wants to write/read and can just chunk them on the transport layer then reconstruct on the other side.
> 
> As an example, lets say we're writing out an entire block. The sender serialized in normal Bitcoin wire format, and then sends them out into 65KB chunks. The receiver reads the initial message header (the type), sees that it's a block, then uses a normal stream decoding that'll parse all the Bitcoin p2p varints to continue reading the entire block out.
> 
> Here's a commit w/ some code that demonstrates how this would work in practice: [lightningnetwork/lnd@767c550](https://github.com/lightningnetwork/lnd/commit/767c550d65ef97a765eabe09c97941d91e05f054). See the `TestWriteMessageChunking` test at the bottom that writes out 195 KB in a chunked message, with the receiver reading out the entire thing. The main idea is that the application layer operates on a stream level (read out N bytes), while the transport operates on a message level (read the next message).
> 
> A bit of the interaction on the reader side in that commit is hidden behind these lines:
> 
> ```go
> 	// Attempt to read the entirety of the message generated above.
> 	buf := make([]byte, len(largeMessage))
> 	if _, err := io.ReadFull(remoteConn, buf); err != nil {
> 		t.Fatalf("unable to read message")
> 	}
> ```
> 
> Which ends up calling into this routine (read in a loop until you read out enough bytes): https://cs.opensource.google/go/go/+/refs/tags/go1.18.3:src/io/io.go;drc=17083a2fdf4475c3f11a3e6a0ef8cb595a5fc4d6;l=326. An internal buffer is then always written to each time we read a transport message off the wire. The reader then gets a stream-like API into that buffer: [lightningnetwork/lnd@767c550#diff-68fea4a34829ad17d09f775b8196da5e09285ad9fcd8ee9ce74704dd2a1c0d61R98](https://github.com/lightningnetwork/lnd/commit/767c550d65ef97a765eabe09c97941d91e05f054#diff-68fea4a34829ad17d09f775b8196da5e09285ad9fcd8ee9ce74704dd2a1c0d61R98).

This is what we do in our Scala implementation currently, if we can't correctly parse a known TLV, we continue reading bytes from the buffer. 

https://github.com/bitcoin-s/bitcoin-s/blob/f3c443804b4202e838213f501b3d0de286fca2ec/dlc-node/src/main/scala/org/bitcoins/dlc/node/DLCConnectionHandler.scala#L53

We do this for our bitcoin p2p network messages too, I think this is the easiest path forward with the least complexity

Tibo-lg:
During today's specification meeting it was agreed to not use any specific message at the moment so closing that for now.

PR #193, opened 2022-04-22
Add some validity checks for offer messages and oracle announcements

Tibo-lg:
Title says it all

### DIFF
--- a/Protocol.md
+++ b/Protocol.md
@@ -115,7 +115,9 @@ Outputs in the funding transaction will be sorted by `change_serial_id` and `fun
 sides will use to compute fees in the funding transaction, as described in the
 [transaction specification](Transactions.md).
 
-`cet_locktime` is the nLockTime to be put on CETs. `refund_locktime` is the nLockTime to be put on the refund transaction.
+`cet_locktime` is the nLockTime to be put on CETs.
+
+`refund_locktime` is the nLockTime to be put on the refund transaction.
 
 #### Requirements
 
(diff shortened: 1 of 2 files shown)


Protocol.md
```diff
@@ -130,7 +132,9 @@ The sending node MUST:
   - set `cet_locktime` to be less than `refund_locktime`.
   - use a unique `input_serial_id` for each input
   - set `change_serial_id` and `fund_output_serial_id` to different values
-  - use valid [contract descriptor(s)](./Messaging.md#The-contract_descriptor-Type) within `contract_info`.
+  - use valid [contract descriptor(s)](./Messaging.md#The-contract_descriptor-Type) within `contract_info`
+  - set `cet_locktime` to the earliest `event_maturity_epoch` amongst all `oracle_event` used included in the `contract_info`
```
nkohen: Is this strictly necessary? I feel like any time before the earliest oracle announcement is fine (including current block height) so as to avoid any UX issues? (I agree that refund_locktime should be about a week after latest contract execution option)

Protocol.md
```diff
@@ -166,6 +172,9 @@ The receiving node MUST reject the contract if:
   - The `fund_output_serial_id` and `change_serial_id` are not set to different value
   - Any input in `funding_inputs` is not a BIP141 (Segregated Witness) input.
   - invalid [contract descriptor(s)](./Messaging.md#The-contract_descriptor-Type) are used within `contract_info`.
+  - `cet_locktime` is not set to the earliest maturity time of all included oracle events.
```
nkohen: as I said above, I feel like `cet_locktime` doesn't need to be so strict (and refund locktime should be far enough away from the latest possible execution)

Protocol.md
Tibo-lg: Good point thanks! Updated to "less than or equal"

nkohen (approved):

Protocol.md
```diff
@@ -139,6 +143,7 @@ The sending node SHOULD:
   - set `refund_locktime` sufficiently long after the latest possible release of oracle signatures added to all other delays to closing the contract.
   - set `payout_spk` to a previously unused script public key.
   - set `change_spk` to a previously unused script public key.
+  - set `refund_locktime` to a value not too big that both party can be expected to be refunded in a reasonable time-frame (recommended not more than latest contract maturity + 86400 * 14 meaning 2 weeks after latest contract maturity).
```
matthewjablack, line 146: Perhaps it'd be good to have a recommendation for minimum as well (i.e. recommended no less than 2 days after the latest contract maturity) to give the oracle time to attest in case there are any infrastructure issues

Tibo-lg: Isn't that cover [here](https://github.com/discreetlogcontracts/dlcspecs/pull/193/files#diff-e0f5b925f91a1c09c6daf26f9a7d28816cb9ff9f08863faca719b7ee0a1cc065R137)?

matthewjablack: Yep you're right, looks good 👍

PR #196, opened 2022-05-16
DLC channel specifications

Tibo-lg:


Protocol.md
```diff
@@ -300,6 +317,173 @@ The recipient:
...
+
+The lifetime of a channel is composed of its establishment, and the possibility for each party to offer the settlement or renewal of the contract within the channel, as well as to close the channel.
+
+The establishment protocol is depicted in the following [state machine](#state-machine-representation).
+
+<img src="./images/dlc_channel_establish_ta.svg">
+
+The protocol for renewal, settlement and closing of the channel is shown in the following [state machine](#state-machine-representation).
+
+<img src="./images/dlc_channel_settle_renew.svg">
```
bonomat, line 332: Am I reading it correctly that if `RenewOffer` was not accepted (`RenewOffered` state), e.g. it timed out, it will be unilateral closed?

Why did you design it that way? Is it not safe to go back to `Established` so that `RenewOffer` can be sent again?

bonomat: Same question in regards to `SettleOffer`

Protocol.md
```diff
@@ -300,6 +317,173 @@ The recipient:
...
+### State machine representation
+ The state machine is represented as a timed automaton augmented with:
+* channel synchronization representing message passing between the processes (`?` indicating the receiving of a message while `!` indicates its sending),
+* edge guards only enabling state transition to be taken when the condition is satisfied (e.g. `Invalid` being true),
+* location invariants requiring the associated condition to hold for the automaton to stay in the location,
+* updates, setting the value of free variables and clocks. 
+
+Note that transitions setting the variable `Invalid` to true are omitted for clarity.
+
+## The `offer_channel` message
```
bonomat, line 347: The `CloseOffer` message seems to be missing.

Protocol.md
```diff
@@ -300,6 +317,173 @@ The recipient:
...
+   * [`point`:`first_per_update_point`]
+   * [`spk`:`payout_spk`]
+   * [`u64`:`payout_serial_id`]
+   * [`u64`:`offer_collateral_satoshis`]
+   * [`bigsize`:`num_funding_inputs`]
+   * [`num_funding_inputs*funding_input`:`funding_inputs`]
+   * [`spk`:`change_spk`]
+   * [`u64`:`change_serial_id`]
+   * [`u64`:`fund_output_serial_id`]
+   * [`u64`:`feerate_per_vb`]
```
bonomat, line 370: Do you intend to take the same `feerate_per_vb` for all transactions (`FundTx`, `RefundTx` and `CloseTx`)?

Protocol.md
```diff
@@ -300,6 +317,173 @@ The recipient:
...
+   * [`32*byte`:`prev_per_update_secret`]
+
+
+## The `renew_offer` message
+
+1. type: 43014
+2. data:
+   * [`channel_id`:`channel_id`]
+   * [`contract_id`:`temporary_contract_id`]
+   * [`u64`:`counter_payout`]
```
bonomat, line 447: In our usecase of CFD-DLCs we have the concept of _funding fees_, i.e. a fee which needs to be paid to keep a position open. 
Do you intend to implicitly put this into `counter_payout`?

Protocol.md
bonomat: I think it makes sense to include a `feerate_per_vb` in the `CloseOffer` message.

Transactions.md
```diff
@@ -312,9 +312,57 @@ accept_cet_vbytes = Ceil(accept_cet_weight / 4.0) vbytes
...
+* txin count: 1
+  * `txin[0]` outpoint: `txid` of funding transaction and `output_index` 0
+  * `txin[0]` sequence: 0xFFFFFFFE
+  * `txin[0]` script bytes: 0
+  * `txin[0]` witness: `0 <signature_for_pubkey1> <signature_for_pubkey2>`
+
+### Outputs
+
+* The settle transaction contains two outputs, paying each party their respective payout agreed during the setup of the settle state (see [here](./Protocol.md#the-settleoffer-message)).
+Note that if the payout of a party is lower than dust threshold, the corresponding output is dropped.
```
bonomat, line 363: If it is dropped it would mean it goes to the miners. Wouldn't it make sense to add this amount to the other party instead?

Transactions.md
```diff
@@ -312,9 +312,57 @@ accept_cet_vbytes = Ceil(accept_cet_weight / 4.0) vbytes
 
 TODO
 
-# References
+# Channel
 
-* [Bitcoin-S implementation](https://github.com/bitcoin-s/bitcoin-s/blob/adaptor-dlc/dlc/src/main/scala/org/bitcoins/dlc/builder)
+## Channel Contract Execution Transaction
```
bonomat, line 317: This is pretty much the same as we have in [maia](https://github.com/comit-network/maia/blob/main/maia/src/protocol/transactions.rs]. 

We call the `Buffer Transaction` `CommitTransaction` because you _commit_ to the state :)

Transactions.md
```diff
@@ -312,9 +312,57 @@ accept_cet_vbytes = Ceil(accept_cet_weight / 4.0) vbytes
...
+  * `txin[0]` sequence: 0xFFFFFFFE
+  * `txin[0]` script bytes: 0
+  * `txin[0]` witness: `0 <signature_for_pubkey1> <signature_for_pubkey2>`
+
+### Output
+
+* The buffer transaction output script is a P2WSH defined by the following miniscript policy:
+
+```
+or(and(pk(offer_own_pk),pk(accept_own_pk)),or(and(pk(offer_own_pk),and(pk(accept_publish_pk), pk(accept_rev_pk))),and(pk(offer_own_pk),and(pk(offer_publish_pk),pk(offer_rev_pk)))))
```
bonomat, line 343: In [maia](https://github.com/comit-network/maia/blob/fc6b78b98407b10b55f8cfd152062ad77f98cd9f/maia/src/protocol/transactions.rs#L277) we have an additional `RefundTransaction` which spends from the `Buffer Transaction` (`CommitTransaction` in our case). This allows you to get your funds back if the other party disappears and the oracle never attests.

bonomat (commented):
Sorry for the long wait for feedback. The transactions are very similar to what we have implemented in https://github.com/comit-network/maia/. The hard part to make it interoperable will be the different message types/styles.

Protocol.md
Tibo-lg: I see your point. My thinking was that if your peer is unresponsive, you probably want to close the channel. It is true that these two states are a bit special in that the application of the counter party would need to get approval (or refusal) from the user before responding. But it is also not necessary to have a single timeout for all states. You could have a long one for these two, and only close if you don't get a response after it. Other solutions I can think of:
* Being able to send `RenewOffer` and `SettleOffer` while already in these states. Things become a bit tricky though, I guess we would need to add a counter to them and have the responder reference it to know which one is being accepted/rejected.
* Being able to cancel an offer. That feels a bit easier.

 > Is it not safe to go back to Established so that RenewOffer can be sent again?

I might not be understanding exactly, but if you go back to `Established` without notifying the other peer you sort of become out of sync don't you?

Protocol.md
Tibo-lg: Thanks, added the message as well as the `feerate_per_vb`. In LN they actually use a range, which can be nice but also mean a more complex protocol (can't send the signature right away).

Protocol.md
Tibo-lg: That's what I do at the moment yes. Do you have a different suggestion? Also my thinking is to maybe have something to update the fee [like in LN](https://github.com/lightning/bolts/blob/master/02-peer-protocol.md#updating-fees-update_fee).

Protocol.md
Tibo-lg: We used to have something similar in our [p2pderivatives app](https://github.com/p2pderivatives/p2pderivatives-client), users could set a premium to pay to the other party which would be included in the funding transaction. Though in that case if it doesn't require a different output I think indeed it might be better to have it implicit in the `counter_payout`. You could then have some extra TLVs for application specific things like that (I haven't specified the extra TLV field yet but you can assume any message can have them).

Transactions.md
Tibo-lg: I don't have a strong opinion but that's how it's specified for the on-chain version. It would be a bit annoying to have two different rules so maybe open an issue and we can discuss if things should be changed for both?

Transactions.md
Tibo-lg: Yeah I remember that's the naming you use. Again I don't mind that much about the naming but what I worry is that it can be confused with the LN `CommitTransaction` especially if we later specified DLC channels working with LN ones.

Transactions.md
Tibo-lg: Yes I have that as well, just didn't mention it. Added a small note l320 let me know if it's clear enough or not.

Tibo-lg:
> Sorry for the long wait for feedback. The transactions are very similar to what we have implemented in https://github.com/comit-network/maia/. The hard part to make it interoperable will be the different message types/styles.

Thanks for the review @bonomat. I'm open for changes, though I'd like to keep things consistent with the on-chain specs where it make sense as multiple implementations already support it.

Just a side note, lately I have been considering if it would be better to share the `next_per_update_point` as part of the previous update. IIUC that's what LN is doing and I think it would save a message per update (since if a party already has the point they can already build and sign transactions).

Protocol.md
bonomat: We are doing the same thing and additionally we send a new `feerate_per_vb` whenever we extend/rollover a contract.

For our usecase in particular an `update_fee` message would not be needed as either you contact the other party to _rollover/extend_ your contract and provide a new `feerate_per_vb` or you closed the position with `CloseOffer` which also sending over a `feerate_per_vb`.

Transactions.md
bonomat: No worries, I just wanted to point it out :) I have no hard feelings about namings in general.

Protocol.md
Tibo-lg: Very good point. Though I wonder if we should have something to indicate that we want to accept a renewal but if a different fee rate?

PR #202, opened 2022-08-30
Hyperbolic curve piece usage clarification

rorp:
Undefined points on a hyperbola should be defined somehow.


Tibo-lg (approved):
I think this is somehow already covered by the [validity requirement of the contract info](https://github.com/discreetlogcontracts/dlcspecs/blob/master/Messaging.md#the-contract_descriptor-type) but I guess it doesn't hurt to have it there as well.

PR #206, opened 2023-02-09
Correct Typo in Signature Point calculation

SachinMeier:
in BIP340,  the signature Point is calculated like so: `s⋅G = R + hash(R || P || m)⋅P`. Unless the order of inputs to the hash function is changed here from BIP 340, this is a typo.

see: https://github.com/bitcoin/bips/blob/master/bip-0340.mediawiki#design


Tibo-lg (approved):
Thanks!

PR #209, opened 2025-01-25
Use full tags for signing

raphjaph:
I think the tagged hash should be created from the full path name `DLC/oracle/.....`.


Christewart:
ACK

PR #211, opened 2025-12-18
Changes to the standard scripts

raulcano:
The description in the specs about the acceptable standard scripts is confusing (or wrong, depending on how it's read).
For example, the typical P2PKH script has this form:
OP_DUP OP_HASH160 <20-bytes> OP_EQUALVERIFY OP_CHECKSIG 

not this one (as written in the current specs - notice the number '20' in the script)
OP_DUP OP_HASH160 **20** <20-bytes> OP_EQUALVERIFY OP_CHECKSIG 

I might be wrong, but I couldn't find a reference where such '20' or '32' is pushed before the actual hashes. Maybe it was intended for clarity? 
In any case, I believe my proposal here is slightly more clear. 
Reference: https://en.bitcoin.it/wiki/Script#Standard_Transaction_to_Bitcoin_address_(pay-to-pubkey-hash)

### DIFF
--- a/Protocol.md
+++ b/Protocol.md
@@ -303,7 +303,7 @@ OP_HASH160 <20-bytes> OP_EQUAL
 # Pay-to-witness-pubkey-hash (P2WPKH), version 0
 OP_0 <20-bytes>
 
-# Pay-to-witness-pubkey-hash (P2WSH), version 0
+# Pay-to-witness-script-hash (P2WSH), version 0
 OP_0 <32-bytes>
 
 # Witness program versions 1 through 16, followed by a single push of 2 to 40 bytes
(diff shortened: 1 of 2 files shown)


Christewart (approved):
Seems reasonable to me as this appears to be the convention throughout the rest of the Script ecosystem

raulcano:
Some further clarification to my original post:

According to https://en.bitcoin.it/wiki/Script:
"When notating scripts, data to be pushed to the stack is generally enclosed in angle brackets and **data push commands are omitted**. Non-bracketed words are opcodes. These examples include the “OP_” prefix, but it is permissible to omit it. Thus “<pubkey1> <pubkey2> OP_2 OP_CHECKMULTISIG" may be abbreviated to "<pubkey1> <pubkey2> 2 CHECKMULTISIG".

So it is true that the data push opcode is part of the actual script, but not displayed in common notation. I suppose my PR still makes sense if we are to keep the scripts aligned with this notation.

Issue #214, opened 2026-01-25
Conflictive definition of ECDSA Adaptor Signature


raulcano:
An adaptor signature is described in this page as the byte concatenation of `R || R_a || s_a || proof`
https://github.com/discreetlogcontracts/dlcspecs/blob/master/ECDSA-adaptor.md#encrypted-signing

See this extract:
> and is defined as:
> 
>     Set k to sample_nonce(Y || message_hash || x)
>     Set m to scalar(message_hash)
>     Set R_a to k * G
>     Set R to k * Y
>     Set proof to DLEQ_prove(k, (R_a, Y, R))
>     Set r to the x-coordinate of R modulo n (i.e. convert it to a scalar)
>     Set s_a to k⁻¹ (m + r * x)
>     Set a to R || R_a || s_a || proof
>     Return a
> 
> The return value is an adaptor signature.

Likewise, we see in [Fundamental Types](https://github.com/discreetlogcontracts/dlcspecs/blob/master/Messaging.md#fundamental-types) 
- the definition of `ecdsa_adaptor_signature` as a convenience type of 65-byte length.
- the definition of `dleq_proof` as a convenience type of 97-byte length.

I believe these definitions are incompatible. The `a` from the first description includes already a `proof`. So, if the proof is already 97-byte, then `a` cannot be 65-bytes.

From my (limited) understanding, it seems the adaptor signature should be the `s_a` value and not the concatenation of all `R || R_a || s_a || proof`. While this may be useful, a bit more clarity would be welcome. Thanks and have a good evening.

Tibo-lg:
These two sections describe different things. The first describe the algorithms for generating and verifying adaptor signatures. The second the messages being exchanged in the protocol. In the second `ecdsa_adaptor_signature` is I think `R || s_a`, which is encoded similarly as a regular ECDSA signature in compact format. Then notice that it is followed by the `dleq_proof`, which is the `proof` described in the algorithm. Finally, the `R_a` is not necessary there because in the protocol we use the anticipation points for the events as encryption keys which are computed from information obtained from the oracle(s).

So TLDR; is that these two sections describe different things.

raulcano:
Thanks for the reply, Tibo-lg.
[...]

PR #215, opened 2026-02-08
Description  to ECDSA adaptor signature and DLEQ proof types

raulcano:
Following the discussion I initiated here https://github.com/discreetlogcontracts/dlcspecs/issues/214
I believe it will be positive to add a detailed description to what exactly are the types `ecdsa_adaptor_signature` and `dleq_proof`. For this, I also add a new convenience type `scalar`.

### DIFF
--- a/ECDSA-adaptor.md
+++ b/ECDSA-adaptor.md
@@ -67,9 +67,9 @@ The challenge hash `H` below is defined as `H(x) = scalar(SHA256(tag || x))`.
 
 ### Proving
 
-The `DLEQ_prove(x, (X, Y, Z))` algorithm takes the following inputs:
+The `DLEQ_prove(k, (X, Y, Z))` algorithm takes the following inputs:
 
-- `x`: A **non-zero** scalar representing the witness for the proof.
+- `k`: A **non-zero** scalar representing the witness for the proof.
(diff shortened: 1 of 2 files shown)


raulcano:
Just a kind reminder to review this update. I believe it's important to have the types `ecdsa_adaptor_signature` and `dleq_proof` properly documented.

One caveat though, the description I have added to the Messaging.md follows the original spec of "65 bytes for an adaptor sig" and "97 bytes for a dleq proof". 
However, the [ECDSA-Adaptor.md](https://github.com/discreetlogcontracts/dlcspecs/blob/master/ECDSA-adaptor.md) file still describes the adaptor signature as `R || R_a || s_a || proof`, which is not consistent with the description in Messaging.md. 

I would propose to slightly modify that bit to something like:
```
Set a to R || s_a || R_a || proof
Return a
The return value is an adaptor signature along with the proof.
```
Note that the new return value has:
- R || s_a = 65 bytes, which is what I have described as the `ecdsa_adaptor_signature` type in this PR.
- R_a || proof = 97 bytes, which is what I have described as the `dleq_proof` type in this PR.

Probably we should work out a formulation to ensure consistency but given this [discussion](https://github.com/discreetlogcontracts/dlcspecs/pull/114) I'm not sure if we would open a can of worms.

PR #216, opened 2026-03-01
minor correction to u64 (from BigSize) of types in the polynomial curve

raulcano:
The text said each point has “two bigsize integers,” but the serialization (lines 164–166) and the general payout curve endpoints (lines 72–74) both use u64 for the two numeric fields. So the prose was wrong: it should say u64, not bigsize.

### DIFF
--- a/PayoutCurve.md
+++ b/PayoutCurve.md
@@ -170,7 +170,7 @@ where a spline is made up of polynomial pieces so that the resulting interpolati
    * [`u16`:`extra_precision_num_pts`]
 
 `num_pts` is the number of midpoints specified in this curve piece which will be used along with the surrounding `endpoint`s to perform interpolation.
-Each point consists of a two `bigsize` integers and a `u16` which are interpreted as `x` and `y` coordinates in exactly the same manner as is done
+Each point consists of two `u64` integers and a `u16` which are interpreted as `x` and `y` coordinates (and extra precision) in exactly the same manner as is done
 for `endpoint`s in [general payout curves](#version-0-payout_function).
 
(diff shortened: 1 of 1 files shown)


Christewart:
FWIW, bitcoin-s follows the spec as written: https://github.com/bitcoin-s/bitcoin-s/blob/80be2f59892b2049d66a506b1e8ae49e30c5ebe8/core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala#L1096

It appears that the implementation you are citing doesn't. Links would be helpful rather than just raw line numbers as it is unclear to me what files those are referring to

raulcano:
Ok, fair enough, sorry about that. I thought the actual changed file in the commit tab was enough , but it's true it's very unconvenient to have to go and check.

In any case, the inconsistency persists. I have no strong opinion on what type to keep, but either it's BigSize or it's u16.
If there is agreement that we need to align the text to the bitcoin-s implementation, fine for me. I just didn't write this text originally and don't know the reasons of why one type was chosen over the other.

For clarity, what I was citing is not another implementation but this very specification, in the file **PayoutCurve.md**
Lines 72-74:
https://github.com/discreetlogcontracts/dlcspecs/blob/9cd9148938c616690c79d99ec6f330e213c246c5/PayoutCurve.md?plain=1#L72-L74

Lines 164-166:
https://github.com/discreetlogcontracts/dlcspecs/blob/9cd9148938c616690c79d99ec6f330e213c246c5/PayoutCurve.md?plain=1#L164-L166

Line 173:
https://github.com/discreetlogcontracts/dlcspecs/blob/9cd9148938c616690c79d99ec6f330e213c246c5/PayoutCurve.md?plain=1#L173

